PR #5, opened 2017-10-25
More advanced auth

addaleax:
- try/catch reading + parsing the config file
- If config data is missing/invalid, prompt the user for data

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -1,10 +1,64 @@
 'use strict';
 
+const assert = require('assert');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const util = require('util');
+const readline = require('readline');
+const readFile = util.promisify(fs.readFile);
+const writeFile = util.promisify(fs.writeFile);
 const authFile = path.join(os.homedir(), '.ncurc');
-// TODO: try-catch, validate properties
-const { username, token } = JSON.parse(fs.readFileSync(authFile, 'utf8'));
-const auth = Buffer.from(`${username}:${token}`).toString('base64');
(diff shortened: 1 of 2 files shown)


lib/auth.js
```diff
@@ -1,10 +1,62 @@
...
+  // Ask the user for input and write to ~/.ncurc, then try again
+  process.stdout.write('Please enter your Github user information:\n' +
+      '[Github tokens can be created as described in ' +
+      'https://help.github.com/articles/' +
+      'creating-a-personal-access-token-for-the-command-line/]\n');
+  username = await prompt('Github user name');
+  token = await prompt('Github token');
+  check(username, token);
+  const json = JSON.stringify({ username, token }, null, '  ');
+  await writeFile(authFile, json);
```
jasnell: should set some reasonable default permissions on this file perhaps? owner read+write only perhaps?

addaleax: yes. :) done!

lib/request.js
```diff
@@ -8,7 +8,7 @@ async function request(query, variables) {
     uri: 'https://api.github.com/graphql',
     method: 'POST',
     headers: {
-      'Authorization': `Basic ${auth}`,
+      'Authorization': `Basic ${await auth()}`,
```
joyeecheung: I think we should do this during start up instead of reading/prompting for crendentials everytime a request is sent?

addaleax: @joyeecheung Ah, right, I *meant* to cache the value here :smile:

joyeecheung:
Awesome work! Thanks!

PR #6, opened 2017-10-25
allow passing full URLs on the command line

addaleax:
i.e. make `get-metadata https://github.com/nodejs/node/pull/16489` work

### DIFF
--- a/bin/metadata.js
+++ b/bin/metadata.js
@@ -25,7 +25,7 @@ const MetadataGenerator = require('../lib/metadata_gen');
 const OWNER = 'nodejs';
 const REPO = 'node';
 
-const PR_ID = parseInt(process.argv[2]);  // example
+const PR_ID = parsePRId(process.argv[2]);
 
 async function main(prid, owner, repo) {
   logger.trace(`Getting collaborator contacts from README of ${owner}/${repo}`);
@@ -68,3 +68,13 @@ main(PR_ID, OWNER, REPO).catch((err) => {
   logger.error(err);
   process.exit(-1);
 });
+
+function parsePRId(id) {
+  // Fast path: numeric string
+  if (`${+id}` === id)
+    return +id;
+  const match = id.match(/^https:.*\/pull\/([0-9]+)(?:\/(?:files)?)?$/);
+  if (match !== null)
+    return +match[1];
+  throw new Error(`Could not understand PR id format: ${id}`);
+}


bin/metadata.js
```diff
@@ -68,3 +68,13 @@ main(PR_ID, OWNER, REPO).catch((err) => {
   logger.error(err);
   process.exit(-1);
 });
+
+function parsePRId(id) {
+  // Fast path: numeric string
+  if (`${+id}` === id)
+    return +id;
+  const match = id.match(/^https:.*\/([0-9]+)$/);
```
jasnell: perhaps try `new URL(id)`, and fallback if it errors?

addaleax: yeah, question: fall back to what? :smile:

jasnell: heh... good point ;-) ... try `new url.URL(id)`, check the protocol, and error if it's not a valid https URL

bin/metadata.js
```diff
@@ -25,7 +25,7 @@ const MetadataGenerator = require('../lib/metadata_gen');
 const OWNER = 'nodejs';
 const REPO = 'node';
 
-const PR_ID = parseInt(process.argv[2]);  // example
+const PR_ID = parsePRId(process.argv[2]);  // example
```
joyeecheung: I think the `// example` comment can be removed now XD I added it when I chose 14782 as the default PR to test...I can do that later.

bin/metadata.js
joyeecheung: Maybe a `\/?` at the end because sometimes people do paste the slash, like `https://github.com/nodejs/node/pull/14782/`...also maybe we can just make the regexp out of `repo` and `owner`? That way we can validate it's a pull instead of an issue as well (right now if it's an issue we just throw the errors around).

joyeecheung: Ah wait, we don't provide command line options for `repo`/`owner` right now :/ OK maybe just validate `pull/([0-9])+\/?` part..

addaleax: Yup – I’ve added checks for `pull/` and `/` or `/files` at the end :)

bin/metadata.js
addaleax: done! :)

joyeecheung:
I will just use the squash and merge button here...thanks!

PR #11, opened 2017-10-26
chore: introduce eslint

abouthiroppy:
```
> eslint . "--fix"
/node-core-utils/lib/auth.js
  47:10  error  Redundant use of `await` on a return value      no-return-await
  70:5   error  Return statement should not contain assignment  no-return-assign

/node-core-utils/lib/request.js
  38:7  error  'totalCount' is assigned a value but never used  no-unused-vars
```

We have 3 errors. What do you think?(accept rules, etc...)
But `totalCount` seems to be used 🤔

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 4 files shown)


addaleax:
I’m merging this since I have fixes for the first ones because I’m trying to do the same thing :)

PR #12, opened 2017-10-26
add first tests :)

addaleax:
Really no idea what the current cool™ testing framework is these days, I always use mocha but it has its downsides – this just adds a single file that can run as a node script.

It’s also one of the things that’s slightly harder to test, so I feel responsible for writing the test for it ;)

@joyeecheung Could you enable Travis for this repo?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,5 +5,14 @@
     "space-before-function-paren": ["error", "never"],
     "no-multi-spaces": ["error", { "ignoreEOLComments": true }],
     "camelcase": "off"
-  }
+  },
+  "env": {
+    "node": true
+  },
+  "parserOptions": {
+    "ecmaFeatures": {
+      "globalReturn": true
+    },
+    "sourceType": "script"
+  },
 }
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 npm-debug.log
 .nyc_output
 coverage
+tmp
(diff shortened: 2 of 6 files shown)


joyeecheung:
No idea about the current cool™ testing framework either, these days I use mocha + [power-assert](https://github.com/power-assert-js/power-assert) when I don't want to remember all the assertion APIs out there, or jest when I am feeling lazy.

I have enabled travis, I will send in another PR to get the test infra ready first(I am thinking about mocha & power-assert & nyc)

addaleax:
rebased :)

joyeecheung:
About that syntax error complained by the loader: I just realized the object spread syntax is still in stage 3 and not supported by acorn (used by power-assert underneath) even turning the experimental support of ES2018 on. We are probably too used to use new syntaxes in core, lol.

```
'use strict';
const acorn = require('acorn');
const code = '"use strict"; var a = { ...{a: 2} };';
try {
   acorn.parse(code, { locations: true, ecmaVersion: 2018,  })
} catch(e) {
  console.log(e);
}
```

addaleax:
@joyeecheung Just curious, how does parsing (transpilation?) actually play a role here? I mean, it shouldn’t affect the way the test is running, right?

joyeecheung:
@addaleax My guess is the power-assert module needs to parse the code to understand the asserts...so that it can display those fancy object diffs (you can try deepStrictEqual two different objects to see it). Anyway we don't really have to use it tho, it's just eye candy for the assert core module

PR #13, opened 2017-10-27
Add test infrasturcture

joyeecheung:


joyeecheung:
Based on https://github.com/joyeecheung/node-core-utils/pull/12 , @addaleax PTAL

joyeecheung:
Also: probably need a way to get Travis run the intergration tests..or not?

PR #14, opened 2017-10-27
use `coverage-all` as the CI test script

addaleax:
I think this should be okay?


joyeecheung:
Yes, I was thinking about adding coveralls support later, but we can change the test script now.

PR #16, opened 2017-10-28
fix link tests description

joyeecheung:


joyeecheung:
Looks like eslint has updated and can find the trailing errors in comments now, so the build has failed since the master fails anyway.

Master build: https://travis-ci.org/joyeecheung/node-core-utils/builds/294157076

PR #17, opened 2017-10-28
Test metadata generator

joyeecheung:


lib/metadata_gen.js
```diff
@@ -1,14 +1,24 @@
 'use strict';
 
 const LinkParser = require('./links');
-
+/**
+ * @typedef {{reviewer: Collaborator}} Reviewer
+ */
```
addaleax: Is this from some specific typechecking/type annotation language? It might be nice to know for when I write my own PRs and I don’t recognize this as something I know :)

joyeecheung: It's just [JSDoc comments](http://usejsdoc.org/tags-typedef.html). I use vscode, it's supposed to understand this, but apparently it does not take `/// <reference path="./collaborators.js"/>` for non-typescript projects at the moment, so the intellisense stops at per-file scopes..but it's good enough for human to read as well, I normally annotate the code even not for the intellisense, this helps me remember what to put in fixtures and what to expect.

screen shot 2017-10-29 at 1 57 48 am

PR #18, opened 2017-10-28
test: add codecov support

joyeecheung:


addaleax:
nice!

joyeecheung:
No base because master does not have this yet.

PR #19, opened 2017-10-29
metadata: put fixes and refs before reviewers

joyeecheung:
This is the convention of node-review.


joyeecheung:
Hmm, I turned off branch update build in Travis so it does not trigger when I push to a branch but have not submitted an update, but looks like this prevents Travis from triggering builds for master merge events as well..I'll turn it on for now and try not to push before a branch is ready.

PR #21, opened 2017-10-29
check if commit author is PR author

joyeecheung:


joyeecheung:
@addaleax 

> We might want to check .mailmap for email management as well?

Great idea, totally forgot that we have this XD

> Checking whether the author’s github name matches the name in the commit Author: field would also be really nice as a thing in the future :)

Hmm...you mean their name as in, "Joyee Cheung" or "Anna Henningsen"? I believe that should be `author.name`(Github) `committer.name`(`git config user.name`)...let me give it a try to see if I guess right.

joyeecheung:
Tests added, the coverage decreases a bit because I have not tested every checks implemented by `PRChecker`, don't want to make this PR super huge so I'll open another PR that tests the rest of that class..

PR #23, opened 2017-10-30
test: add simple tests for comparisons

maclover7:


joyeecheung:
Also would be great if this tests the comparison is actually good for sorting...so instead of checking the return values, just check if it sorts an array when passed to `arr.sort()`

maclover7:
updated @joyeecheung, let me know if this looks good :)

joyeecheung:
Thanks!

PR #28, opened 2017-10-31
Add linter CI checks and test CI checks

joyeecheung:


joyeecheung:
Went to the Travis CI issue tracker and looks like there is an outage. Wait until they fix it.

PR #31, opened 2017-10-31
travis: cache node_modules so faster build

cPhost:
Right now the time it takes to run a test is not large but in future if the npm packages increases the build will take longer doing so will make build quicker.


joyeecheung:
Thanks!

PR #33, opened 2017-11-01
Make auth tests work with async/await

joyeecheung:
Fixes: https://github.com/joyeecheung/node-core-utils/issues/32

### DIFF
--- a/test/unit/auth.test.js
+++ b/test/unit/auth.test.js
@@ -8,9 +8,10 @@ const fs = require('fs');
 const assert = require('assert');
 let testCounter = 0; // for tmp directories
 
-describe('auth', () => {
-  it('asks for auth data if no ncurc is found', () => {
-    runAuthScript(undefined, [
+describe('auth', async function() {
+  it('asks for auth data if no ncurc is found', async function() {
+    this.timeout(1000);
+    await runAuthScript(undefined, [
       'Reading configuration for node-core-utils failed:',
       /ENOENT: no such file or directory, open/,
       'Please enter your Github user information:',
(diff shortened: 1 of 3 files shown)


joyeecheung:
cc @cPhost @addaleax @refack

test/unit/auth.test.js
```diff
@@ -33,80 +36,98 @@ describe('auth', () => {
...
-    { env: Object.assign({}, process.env, { HOME }) });
-  let stderr = '';
-  proc.stderr.setEncoding('utf8');
-  proc.stderr.on('data', (chunk) => { stderr += chunk; });
-  proc.on('close', () => {
-    if (stderr) { throw new Error(`unexpected stderr:\n${stderr}`); }
-  });
+    const proc = spawn(process.execPath,
+      [ require.resolve('../fixtures/run-auth') ],
+      // XXX this could just be env: { ...process.env, HOME } but the test loader
```
joyeecheung: Ah we already know why this happens in https://github.com/joyeecheung/node-core-utils/pull/12#issuecomment-339955478 ...Can remove this now.

addaleax: Yup… it’s a shame though :smile:

cPhost:
Great! and that closes https://github.com/issues/32

test/unit/auth.test.js
joyeecheung: @addaleax We'll just have to wait until 2018 (March?), I guess

addaleax: Is that the acorn release schedule or why March? ^^

addaleax: like, this works in Node 8 & 9 without a flag, I’m surprised userland parsers haven’t picked it up before V8…

joyeecheung: @addaleax IIRC ECMA has an assembly that approve the spec in March so by Feb they would need to cut releases before submitting them.

joyeecheung: @addaleax Acorn has a few plugins that support the object spread syntax already, they are just not included by default, I guess they are waiting for the spec to drop.

cPhost:
That should be fixed now.

refack:
Wow you quick 😄

PR #34, opened 2017-11-01
Update README

joyeecheung:


README.md
```diff
@@ -44,8 +47,7 @@ This one is inspired by Evan Lucas's [node-review](https://github.com/evanlucas/
 - [x] Generate `Fixes`
 - [x] Generate `Refs`
 - [x] Check for CI runs
-- [ ] Check if commiters match authors
-  - Only when `"authorAssociation": "FIRST_TIME_CONTRIBUTOR"`
+- [x] Check if commiters match authors
 - [x] Check 48-hour wait
 - [x] Check two TSC approval for semver-major
 - [ ] Warn new commits after reviews
```
joyeecheung: Hmm..also I am not quite sure how this one should be handled because in practice, we still count reviews that give approval before new commits, as long as they don't object afterwards.

refack: So, there was a conversation at some point that people who comment "LGTM" want that to be dismissed if a new commit comes in.
https://github.com/nodejs/node/pull/13413#discussion_r119986852

joyeecheung: Hmm..I think I can not get a conclusion from that thread, or did I miss something? Should we open an issue in the core repo to get this settled?

PR #38, opened 2017-11-01
add tests to `lib/logger`

alopezsanchez:
Adds a fixture to capture the stdout output
in order to make tests as real as possible.

Please, review it. Thanks!

### DIFF
--- a/test/unit/logger.test.js
+++ b/test/unit/logger.test.js
@@ -1,0 +1,94 @@
+'use strict';
+
+const assert = require('assert');
+const chalk = require('chalk');
+
+const logger = require('../../lib/logger');
+const captureStdout = require('../fixtures/capture_stdout');
+
+// Complete ISO date format -> YYYY-MM-DDThh:mm:ss.sTZD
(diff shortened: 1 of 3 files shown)


test/unit/logger.test.js
```diff
@@ -0,0 +1,94 @@
...
+
+// Complete ISO date format -> YYYY-MM-DDThh:mm:ss.sTZD
+const dateRegex = new RegExp(/\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/);
+
+describe('Logger', () => {
+  let stdout = null;
+
+  // Capture the stdout output before each test
+  beforeEach(() => {
+    stdout = captureStdout(process.stdout);
```
joyeecheung: Test bit LGTM, though we should probably avoid monkey patching `process.stdout` in any way. We can get pino to write to a writable stream that implemented as a fixture, just need to figure out how to get it to tell if it is being tested. This is not blocking though, we can fix it later, just having seen another breakage caused by monkey-patching Node internals makes me kinda sensitive...

joyeecheung:
CI is green, I will merge this with a TODO there reminding people to fix the monkey-patching later.

PR #39, opened 2017-11-01
add .editorconfig, fix .eslintrc and correct misspelled test/integration folder

alopezsanchez:
I think it's a good practice to include a `.editoconfig` in the project. What do you think?


joyeecheung:
CI is happy so I will just go ahead and merge this, thanks!

PR #40, opened 2017-11-02
add -a flag to apply PR patch & in some cases metadata

apapirovski:
*So this is basically just a proof of concept before I clean it up and expand on this work. Wanted to check if there's any interest in this before I spend more than 30 minutes...*

This PR adds `-a` flag that, if there are no warnings, applies a PR patch onto the current branch and if the PR only has a single commit (which has no previous metadata), also applies the generated metadata to it.

I personally already have a bunch of bash scripts that do this for me but I would rather just use one tool for everything, if I can. This would also make landing doc changes and similar much faster.

### DIFF
--- a/bin/metadata.js
+++ b/bin/metadata.js
@@ -1,8 +1,12 @@
 #!/usr/bin/env node
 'use strict';
 
+const { exec } = require('child_process');
 const fs = require('fs');
 const path = require('path');
+const { promisify } = require('util');
+
+const execAsync = promisify(exec);
 
 function loadQuery(file) {
   const filePath = path.resolve(__dirname, '..', 'queries', `${file}.gql`);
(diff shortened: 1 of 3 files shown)


joyeecheung:
Great to see you here!

I have given the "amend message" bit some thought lately and I believe we should leave it to other targets, because I would not really expect a command called `get-metadata` to start to do anything to my git history..as for https://github.com/joyeecheung/node-core-utils/issues/35, I think we don't really need a flag because we can detect if stdout is tty and redirect the metadata here. We can defer the git-manipulation part to another target that does not imply idempotence.

apapirovski:
Yeah, I had similar thoughts re: `get-metadata`. Any support for adding something like `apply-pr` or similar?

joyeecheung:
@apapirovski See https://github.com/joyeecheung/node-core-utils/issues/10, I am thinking about a `land-pr` that behaves kinda like `git cherry-pick` that can let you do stuff step-by-step

apapirovski:
Ha! I even looked around but clearly I'm not very observant... Thanks for linking it! I'm going to spend a few days and think about it. At first glance, I feel like the current proposal in that thread is a tiny bit complicated but maybe that's what it takes to cover all the possibilities.

PR #41, opened 2017-11-02
refactor to put data in a central place

joyeecheung:
Also parallelized the requests.

Before:

```
get-metadata 16548  0.85s user 0.12s system 15% cpu 6.487 total
```

After:

```
get-metadata 16548  0.64s user 0.08s system 27% cpu 2.620 total
```

Needs tests for pr_data to land


joyeecheung:
This should be good to go now

PR #42, opened 2017-11-02
Update package.json

joyeecheung:
Appvoyer is complaining about the missing repository field..


joyeecheung:
All right, it's now complaining the actual errors reported in https://github.com/joyeecheung/node-core-utils/issues/36

PR #45, opened 2017-11-02
Fix metadata generation test on Windows

tniessen:
On Windows, the strings mismatched due to different EOLs. (Template string literals always use LF while the metadata generator emits CRLF.)

### DIFF
--- a/test/unit/metadata_gen.test.js
+++ b/test/unit/metadata_gen.test.js
@@ -6,6 +6,7 @@ const {
   allGreenReviewers
 } = require('../fixtures/data');
 
+const { EOL } = require('os');
 const assert = require('assert');
 const data = {
   repo: 'node',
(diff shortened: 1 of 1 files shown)


cPhost:
@refack 
Fixes https://github.com/joyeecheung/node-core-utils/issues/36?

joyeecheung:
Waiting for appveyor now..

joyeecheung:
@cPhost This does not fix that because the auth tests are still failing..

cPhost:
@joyeecheung appveyor is stalling or something there are lot PR Queued for test and each test are running for 1 hour per quota.

Tiriel:
Appveyor seems to keep every PR hanging, indeed

cPhost:
There seems to be no config commited to repo `appveyor.yml`

joyeecheung:
@cPhost I have manually configured the appveyor setting when I wanted to try it out, should probably put that down in the yml later.

joyeecheung:
I've cancelled all the appveyor builds, let's see if this one finishes in half an hour. If not, I will cancel this and see if https://github.com/joyeecheung/node-core-utils/pull/47 fixes that.

PR #46, opened 2017-11-02
logger: add red color to [STACK] and [DATA]

alopezsanchez:
In order to be consistent with [ERROR] label.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -4,19 +4,19 @@ const pino = require('pino');
 const chalk = require('chalk');
 const { EOL } = require('os');
 
-function paint(level) {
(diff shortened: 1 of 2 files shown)


lib/logger.js
```diff
@@ -7,6 +7,8 @@ const { EOL } = require('os');
 function paint(level) {
```
joyeecheung: This should probably be renamed to something like "text" or "label" now because stack or data are not logging levels

alopezsanchez: :+1: You are right. I'll correct it.

alopezsanchez: Done.

PR #47, opened 2017-11-02
Fix logger test failures on Windows

tniessen:
This PR contains a fix for JSON data not using EOL and temporary workarounds for missing upstream support within pino, see https://github.com/pinojs/pino/issues/326.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -32,10 +32,14 @@ const prettyOptions = {
       timestamp = `[${new Date(obj.time).toISOString()}] `;
     }
     if (level === 'ERROR') {
+      let data;
+      if (obj.data !== undefined) {
+        data = JSON.stringify(obj.data, null, 2).replace(/\n/g, EOL);
+      }
       return `${paint(level)} ${timestamp}${obj.type ? obj.type + ' ' : ''}` +
         `${obj.msg}${EOL}` +
         `${paint('STACK')} ${obj.stack || ''}${EOL}` +
-        `${paint('DATA')} ${JSON.stringify(obj.data, null, 2) || ''}${EOL}`;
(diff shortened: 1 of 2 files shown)


lib/logger.js
```diff
@@ -32,10 +32,14 @@ const prettyOptions = {
...
     if (level === 'ERROR') {
+      let data;
+      if (obj.data !== undefined) {
+        data = JSON.stringify(obj.data, null, 2).replace(/\n/g, EOL);
+      }
       return `${paint(level)} ${timestamp}${obj.type ? obj.type + ' ' : ''}` +
         `${obj.msg}${EOL}` +
         `${paint('STACK')} ${obj.stack || ''}${EOL}` +
-        `${paint('DATA')} ${JSON.stringify(obj.data, null, 2) || ''}${EOL}`;
+        `${paint('DATA')} ${data || ''}${EOL}`;
```
tniessen: This fix is unrelated to pino.

test/unit/logger.test.js
```diff
@@ -22,25 +22,25 @@ describe('Logger', () => {
     it('should have blue color when the level is INFO', () => {
       logger.info('test');
       assert.strictEqual(stream.toString(),
-        `${chalk.blue('[INFO]')} test${EOL}`);
+        `${chalk.blue('[INFO]')} test\n`);
```
tniessen: This and the following changes are necessary due to pino using `\n` on all platforms, see https://github.com/pinojs/pino/issues/326.

joyeecheung:
The logger tests passed in Appvoyer, somehow the job fails to finish after the test is done. Maybe a proper appvoyer.yml will fix that. I will merge this now, thanks!

PR #48, opened 2017-11-02
add appveyor.yml

joyeecheung:


joyeecheung:
I will delete the appveyor project and re add it to see if we can start fresh..

refack:
the test process doesn't seem to end (I believe @cPhost observed that)

joyeecheung:
@refack Yes, it does not continue to run the coverage part after the tests are finished. I will open a PR to trace down which part is causing the hang later...

PR #52, opened 2017-11-02
Properly parse "Fixes:"

tniessen:
Fixes: https://github.com/joyeecheung/node-core-utils/issues/44


test/unit/links.test.js
```diff
@@ -6,18 +6,21 @@ const assert = require('assert');
 const htmls = fixtures.readJSON('op_html.json');
 
 const expected = [{
-  fixes: ['https://github.com/node/issues/16437'],
+  fixes: ['https://github.com/nodejs/node/issues/16437'],
```
tniessen: I believe this test was incorrect before. Not sure whether I am supposed to pass `nodejs/node` as the repo name.

apapirovski:
This seems great! Wondering two things: a) should this be case-insensitive matching and b) should we enforce the colon? "fixes #XYZ" seems valid to me...

tniessen:
@apapirovski Good points, I think both should be changed!

test/unit/links.test.js
joyeecheung: I think we can just pass the repo + the owner name to `LinkParser`

joyeecheung: ^to be clear, I meant change the signature of  the constructor of `LinkParser` to `(owner, repo, html)`

tniessen:
@apapirovski I made the checks case-insensitive. Making the colon optional requires much stricter parsing as the keywords can appear in regular sentences as well.

addaleax:
@tniessen If I’m reading this correctly the regexes still expect the colon to always be there, right?

apapirovski:
I feel like it's probably fine if we parse "fixes #XYZ" anywhere in the copy, no? It's possible I'm missing some downside to that approach though.

tniessen:
> If I’m reading this correctly the regexes still expect the colon to always be there, right?

@addaleax Yes, making the colon optional requires stricter filtering. I am not sure whether we can assume that any Fixes-like keyword followed by `#(decimal)` or by `(word)/(word)/(decimal)` is valid and everything else is not, but it is probably enough.

apapirovski:
We could get fancy and check the API to make sure it's an actual issue 😆 I know, I know...

It seems fine to me as is and we can revisit later, maybe including validating that these are actual open issues.

joyeecheung:
BTW github should have some kind of API about the status of issues they consider can be automatically closed(https://help.github.com/articles/closing-issues-using-keywords/), but it doesn't seem very stable at this point.

joyeecheung:
@tniessen I think this can be merged now?

tniessen:
I will revisit the version without a colon later.

PR #56, opened 2017-11-03
Added todo warn `Warn new commits after reviews`

cPhost:
It warns if there has been new commits after the last review
if so it warns and then outputs two links to one of the commit
link and other link for all the changes that the PR has.

### DIFF
--- a/bin/metadata.js
+++ b/bin/metadata.js
@@ -10,6 +10,7 @@ const PRData = require('../lib/pr_data');
 const PRChecker = require('../lib/pr_checker');
 const MetadataGenerator = require('../lib/metadata_gen');
 const argv = require('../lib/args')();
+const commitAfterReview = require('../lib/commit_after_review');
 
 // const REFERENCE_RE = /referenced this pull request in/
 const OWNER = argv.owner;
@@ -32,8 +33,10 @@ async function main(prid, owner, repo, logger) {
   if (!process.stdout.isTTY) {
     process.stdout.write(`${metadata}${EOL}`);
   }
+
   const checker = new PRChecker(logger, data);
   checker.checkAll();
+  commitAfterReview(data.commits, data.reviews, logger);
 }
 
 const logStream = process.stdout.isTTY ? process.stdout : process.stderr;
--- a/lib/commit_after_review.js
+++ b/lib/commit_after_review.js
@@ -1,0 +1,76 @@
+function commitAfterReview(commit, review, logger) {
+  let commitIndex = commit.length === 0
+    ? 0 : commit.length - 1;
+  let reviewIndex = review.length === 0
+    ? 0 : review.length - 1;
+
+  let lastCommit = commit[commitIndex].commit;
+  let lastReview = review[reviewIndex];
+
+  let commitInfo = getDate(lastCommit.committedDate);
+  let reviewInfo = getDate(lastReview.publishedAt);
+
+  logger.info('checking if new commits were pushed' +
+              ' after last review');
+
+  switch (false) {
+    // very edge case senario
+    // commit after 1 year!
+    case (commitInfo.year < reviewInfo.year ||
+          commitInfo.year === reviewInfo.year):
+      warn();
+      break;
+
+    case (commitInfo.month < reviewInfo.month ||
+          commitInfo.month === reviewInfo.month):
+      warn();
+      break;
+
+    case (commitInfo.date < reviewInfo.date ||
+          commitInfo.date === reviewInfo.date):
+      warn();
+      break;
+
+    case (commitInfo.hour < reviewInfo.hour ||
+          commitInfo.hour === reviewInfo.hour):
+      warn();
+      break;
+
+    case (commitInfo.minutes < reviewInfo.minutes ||
+         commitInfo.minutes === reviewInfo.minutes):
+      warn();
+      break;
+  }
+
+  function warn() {
+    let msg = 'There has been new commits pushed' +
+              ' to Pull Request the since last review';
+    logger.warn(msg);
+    let pullURL = lastReview.url.split('#')[0];
+    let commitURL = `${pullURL}/commits/${lastCommit.oid}`;
+    let allChangesURL = `${pullURL}/files`;
+    logger.warn('New commit url: ', commitURL);
+    logger.warn('see all the changes: ', allChangesURL);
+  }
+}
+
+function getDate(_date) {
+  _date = _date.split('T');
+
+  let day = _date[0];
+  let time = _date[1];
+  day = day.split('-');
+  time = time.split(':');
+
+  let date = {
+    year: +day[0],
+    month: +day[1],
+    date: +day[2],
+    hour: +time[0],
+    minutes: +time[1]
+  };
+
+  return date;
+}
+
+module.exports = commitAfterReview;
--- a/lib/commit_after_review.js
+++ b/lib/commit_after_review.js
@@ -1,7 +1,7 @@
 function commitAfterReview(commit, review, logger) {
-  let commitIndex = commit.length === 0
+  let commitIndex = commit.length === 1
     ? 0 : commit.length - 1;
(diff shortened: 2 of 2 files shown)


cPhost:
The tests are coming soon probably early tomorrow. Just wanted to get it in so you guys could review it. Definitely request changes on naming things differently.

Sample output below. screen shot 2017-11-02 at 9 38 56 pm

lib/commit_after_review.js
```diff
@@ -0,0 +1,76 @@
...
+  let reviewIndex = review.length === 0
+    ? 0 : review.length - 1;
+
+  let lastCommit = commit[commitIndex].commit;
+  let lastReview = review[reviewIndex];
+
+  let commitInfo = getDate(lastCommit.committedDate);
+  let reviewInfo = getDate(lastReview.publishedAt);
+
+  logger.info('checking if new commits were pushed' +
```
apapirovski: Is this necessary? There's no async processing here or anything...

lib/commit_after_review.js
```diff
@@ -0,0 +1,76 @@
...
+  let commitIndex = commit.length === 0
+    ? 0 : commit.length - 1;
+  let reviewIndex = review.length === 0
+    ? 0 : review.length - 1;
+
+  let lastCommit = commit[commitIndex].commit;
+  let lastReview = review[reviewIndex];
+
+  let commitInfo = getDate(lastCommit.committedDate);
+  let reviewInfo = getDate(lastReview.publishedAt);
```
apapirovski: If `review.length === 0` then this will throw

lib/commit_after_review.js
```diff
@@ -0,0 +1,76 @@
+function commitAfterReview(commit, review, logger) {
+  let commitIndex = commit.length === 0
+    ? 0 : commit.length - 1;
+  let reviewIndex = review.length === 0
+    ? 0 : review.length - 1;
+
+  let lastCommit = commit[commitIndex].commit;
```
apapirovski: If `commit.length === 0` then this will throw.

bin/metadata.js
```diff
@@ -32,8 +33,10 @@ async function main(prid, owner, repo, logger) {
   if (!process.stdout.isTTY) {
     process.stdout.write(`${metadata}${EOL}`);
   }
+
   const checker = new PRChecker(logger, data);
   checker.checkAll();
+  commitAfterReview(data.commits, data.reviews, logger);
```
apapirovski: This seems like it should be called from the checker?

lib/commit_after_review.js
cPhost: I will remove that

bin/metadata.js
cPhost: @apapirovski Can you explain more about it.

apapirovski: I just mean that it's conceptually similar in its purpose to the things that are being checked within `checkAll`, so this could be called from there rather than from `bin/metadata.js` and then it could also exist as a method of that `PRChecker` class and have access to the data, instead of needing it passed in.

lib/commit_after_review.js
```diff
@@ -0,0 +1,73 @@
+function commitAfterReview(commit, review, logger) {
+  let commitIndex = commit.length === 1
+    ? 0 : commit.length - 1;
+  let reviewIndex = review.length === 1
+    ? 0 : review.length - 1;
+
+  let lastCommit = commit[commitIndex].commit;
+  let lastReview = review[reviewIndex];
+
+  let commitInfo = getDate(lastCommit.committedDate);
```
joyeecheung: I don't think we need a date object, or get the years/minutes out. ISO date strings are in lexical orders so a simple `lastReview .publishedAt < lastCommit.committedDate` is all we need.

bin/metadata.js
joyeecheung: I agree, this can be implemented as `PRChcker.checkCommitAfterReview`

lib/commit_after_review.js
joyeecheung: The easiest way to grab all oudated reviews would be 

```
const lastCommit = commits.sort((a, b) => descening(a.committedDate, b.committedDate))[0];
if (lastCommit) {
  const oudatedReviews = reviews.filter((r) => r.publishedAt < lastCommit.committedDate);
  // iterate through the outdatedReviews and warn for each
}
```

apapirovski: The one concern I have in general with making this 'super strict' is that currently most people aren't very good about affirming their reviews and often squashes happen with no one coming back to confirm that they still approve. If we log every single review that's outdated then that seems like it might be a tad too much.

I feel like maybe it's almost more helpful to warn for every commit that happened after reviews and show the commits message. That way they can assess whether the PR needs more reviews or it was just maintenance commits.

joyeecheung:
Also can you add some tests in `test/unit/pr_checker.test.js`? Thanks!

lib/commit_after_review.js
```diff
@@ -0,0 +1,73 @@
...
+
+    case (commitInfo.minutes < reviewInfo.minutes ||
+         commitInfo.minutes === reviewInfo.minutes):
+      warn();
+      break;
+  }
+
+  function warn() {
+    let msg = 'There has been new commits pushed' +
+              ' to Pull Request the since last review';
```
tniessen: Grammar: `New changes were pushed since the last review.`

lib/commit_after_review.js
```diff
@@ -0,0 +1,73 @@
...
+    logger.warn(msg);
+    let pullURL = lastReview.url.split('#')[0];
+    let commitURL = `${pullURL}/commits/${lastCommit.oid}`;
+    let allChangesURL = `${pullURL}/files`;
+    logger.warn('New commit url: ', commitURL);
+    logger.warn('see all the changes: ', allChangesURL);
```
tniessen: I don't think this link is particularly helpful as it does not show any correlation between reviews, commits and changes.

tniessen: I would probably just add the link to the list of commits instead of the above lines, we need to check this manually anyway.

lib/commit_after_review.js
joyeecheung: @apapirovski Yes, I think it's more useful to warn about the commits (with their messages) instead of the reviews. If the commits are all like "fix nits" then the people landing them don't have to worry about it.

cPhost:
I will probably close this PR and make a new one with test and all the suggestions and fixes.

lib/commit_after_review.js
cPhost: I did not know that!

apapirovski:
I think a combination of what @tniessen proposed and something like

[WARN] Changes were pushed since the last review:
[WARN] src: improve module loader readability
[WARN] nit: fix error
[WARN] See https://github.com/nodejs/node/pull/16536/commits for more info

I'm not a big fan of adding [commit] or something similar to each line because once you're familiar with the tool it just means you have to look that far further right to start reading. But maybe others have different thoughts on that.

tniessen:
> `[WARN] There has been new commits pushed to Pull Request since the last review`

See https://github.com/joyeecheung/node-core-utils/pull/56#discussion_r148713805. You don't need to use that suggestion but it should conform to English grammar.

> `[WARN] [New Commit] src: improve module loader readability`

Why is `New Commit` enclosed in brackets while `New commit url` is not? Why is `Commit` capitalized in `New Commit` but not in `New commit url`?

> `[WARN] New commit url:  https://github.com/nodejs/node/pull/16536/commits/b619e41d8b2066fcab6459cea0f49f6924200a37`

Why are there two spaces in front of the URL?

cPhost:
@tniessen, I like format by @apapirovski.

apapirovski:
I just tried to ninja edit but I'll post it separately, in case people miss it:

> Maybe if it's just one commit, it links to it and if it's multiple it links to the aggregate commits page.

cPhost:
added new pr

PR #57, opened 2017-11-03
add an intro section with useful info

apapirovski:
So this adds three intro lines of logs that are basically the github PR header (title + pr num, commit author, commit count, branch info + labels). I would like to have this info available and I believe there was a request for it in the issues too.

It's implemented on the PRData class which made the most sense to me as it just logs information from that class. Also, we might want to use this on other targets (such as land-pr).

Tests are included.

### DIFF
--- a/test/fixtures/first_timer_pr.json
+++ b/test/fixtures/first_timer_pr.json
@@ -14,5 +14,8 @@
         "name": "test"
       }
     ]
-  }
+  },
+  "title": "test: awesome changes",
+  "baseRefName": "master",
+  "headRefName": "awesome-changes"
 }
--- a/test/fixtures/test_logger.js
+++ b/test/fixtures/test_logger.js
@@ -19,6 +19,13 @@ class TestLogger {
     return true;
   }
 
+  clear() {
+    this.logs.warn = [];
+    this.logs.info = [];
+    this.logs.trace = [];
+    this.logs.error = [];
+  }
+
   warn(...args) {
     this.logs.warn.push(args);
   }
--- a/test/unit/pr_data.test.js
+++ b/test/unit/pr_data.test.js
@@ -12,7 +12,6 @@ const {
   readme
 } = require('../fixtures/data');
 const TestLogger = require('../fixtures/test_logger');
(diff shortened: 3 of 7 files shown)


joyeecheung:
Refs: https://github.com/joyeecheung/node-core-utils/issues/3

joyeecheung:
Gotta grab a lunch, will review later

test/fixtures/test_logger.js
```diff
@@ -19,6 +19,13 @@ class TestLogger {
     return true;
   }
 
+  clear() {
```
joyeecheung: I think we don't need to clear it because it comes with empty logs when being newed.

test/unit/pr_data.test.js
```diff
@@ -60,4 +60,25 @@ describe('PRData', function() {
     assert.deepStrictEqual(data.commits, simpleCommits);
     assert.deepStrictEqual(data.reviewers, allGreenReviewers);
   });
+
+  it('logIntro', async() => {
+    const logger = new TestLogger();
+    const data = new PRData(16348, 'nodejs', 'node', logger, request);
+    await data.getAll();
+    logger.clear();
```
joyeecheung: See ^

test/fixtures/first_timer_pr.json
```diff
@@ -14,5 +14,8 @@
         "name": "test"
       }
     ]
-  }
-}
\ No newline at end of file
+  },
+  "title": "test: awesome changes",
```
joyeecheung: Can you update the `semver_major_pr.json` as well?

test/fixtures/test_logger.js
joyeecheung: Ah scratch that, I see that it needs to be cleaned because `data.getAll()` will fill the logger.

cPhost: Thanks for this as i was going to implement it for my tests

apapirovski:
Thanks for the reviews, updated. (Surprised codecov doesn't rebase against master when doing a diff...)

PR #60, opened 2017-11-03
Use flat-square style for all badges

targos:


targos:
Before: 
image

After: 
image

PR #61, opened 2017-11-03
doc: add contributing guide and license

joyeecheung:
Refs: https://github.com/joyeecheung/node-core-utils/issues/59

The contributing guide is largely borrowed from the llnode one, although during this patch I noticed that one needs update as well...that one was largely borrowed from the node core one, I dropped a few requirements e.g. wrapping commit message header line in 50 chars because they feel a bit lengthy to me.


README.md
```diff
@@ -74,7 +74,7 @@ $ get-metadata $PRID >> msg.txt
 $ git commit --amend -F msg.txt
 ```
 
-### TODO
+### Features
```
joyeecheung: This is not an accident XD

Tiriel:
It does! Letting a bit more time for reviews before merge

joyeecheung:
I think we should be good to go, if there are any more comments please raise an issue.

PR #62, opened 2017-11-03
mocha, appveyor: add flag --exit so appveryor build does not get stuck

cPhost:
cache node_modules
this still does not fix test failures of auth but rather exits the
test so joyeecheung does not have to cancel the build manually


joyeecheung:
I'll merge this now to make the appveyor finish faster

PR #64, opened 2017-11-03
use ghauth for easier first-time auth flow

apapirovski:
This enables the use of [rvagg/ghauth](https://github.com/rvagg/ghauth) for first-time users instead of a user supplied token that creates the `.ncurc` file. I think this is a slightly nicer user experience. (If they already have a token, creating a `~/.ncurc` file from the terminal seems easy...)

Didn't have to change a ton, especially with the test case — great job on making that code generic enough that it could be so easily repurposed @addaleax!

Fixes: https://github.com/joyeecheung/node-core-utils/issues/1

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -23,8 +23,10 @@ async function auth(getCredentials = ghauth) {
   try {
     ({ username, token } = JSON.parse(await readFile(authFile, 'utf8')));
   } catch (e) {
-    process.stdout.write('Reading configuration for node-core-utils failed:\n' +
-                         e.message + '\n');
+    process.stdout.write('If this is your first time running this command, ' +
+                         'follow the instructions to create an access token. ' +
+                         'If you prefer to create it yourself on Github, ' +
+                         'see https://github.com/joyeecheung/node-core-utils/blob/master/README.md.\n');
   }
 
   // If that worked, yay
@@ -43,8 +45,8 @@ async function auth(getCredentials = ghauth) {
       note: 'node-core-utils CLI tools'
     });
   } catch (e) {
-    process.stdout.write(`Could not get token: ${e.message}\n`);
-    process.exit();
+    process.stderr.write(`Could not get token: ${e.message}\n`);
+    process.exit(1);
   }
 
   const json = JSON.stringify({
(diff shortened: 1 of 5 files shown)


lib/auth.js
```diff
@@ -32,46 +33,46 @@ async function auth() {
...
+  let credentials;
+  try {
+    credentials = await getCredentials({
+      noSave: true,
+      scopes: ['user:email'],
+      note: 'node-core-utils CLI tools'
+    });
+  } catch (e) {
+    process.stdout.write(`Could not get token: ${e.message}\n`);
+    process.exit();
```
addaleax: `process.stderr`, and a non-zero exit code I guess?

lib/auth.js
```diff
@@ -32,46 +33,46 @@ async function auth() {
...
-      'creating-a-personal-access-token-for-the-command-line/]\n');
-  username = await prompt('Github user name');
-  token = await prompt('Github token');
-  check(username, token);
-  const json = JSON.stringify({ username, token }, null, '  ');
+  // Ask the user for input, create a token via github v3 API
+  // then write to ~/.ncurc and try auth() again
+  let credentials;
+  try {
+    credentials = await getCredentials({
```
joyeecheung: Right now this displays:

```
Reading configuration for node-core-utils failed:
ENOENT: no such file or directory, open '/Users/joyee/.ncurc'
Your GitHub username:
```

for first-time users. This looks a bit alarming, maybe pring a line saying something like

```
If this is the first time you run this command,
you can follow the instructions to create the access token.
If you prefer to create the token yourself on Github,
see <readme link>.
```

lib/auth.js
```diff
@@ -32,46 +35,46 @@ async function auth() {
...
     if (cachedValue !== undefined) {
       return cachedValue;
     }
-    cachedValue = fn();
+    cachedValue = fn(...args);
     return cachedValue;
   };
 }
+
+// This is dirty, dirty hack to get around a bug in hyperquest & ghauth
```
tniessen: Nit: `is dirty, dirty hack` → `is a dirty, dirty hack`

PR #65, opened 2017-11-04
todo: warn new commits after review

cPhost:
with all the fixes and suggestions
also added `lint-fix` command


lib/pr_checker.js
```diff
@@ -211,6 +252,36 @@ class PRChecker {
     // 3. is not authored by the people opening the PR
     return true;
   }
+
+  checkCommitsAfterReview(commit, review, logger) {
+    const commitIndex = commit.length - 1;
```
joyeecheung: With this we will need to guarantee that no one alters the order of the arrays. Can you add a comment in the constructor about this?

lib/pr_checker.js
```diff
@@ -211,6 +252,36 @@ class PRChecker {
     // 3. is not authored by the people opening the PR
     return true;
   }
+
+  checkCommitsAfterReview(commit, review, logger) {
```
joyeecheung: `s/commit/commits/` and `s/review/reviews/` because they are arrays?

lib/pr_checker.js
```diff
@@ -211,6 +252,36 @@ class PRChecker {
...
+    const commitIndex = commit.length - 1;
+    const reviewIndex = review.length - 1;
+    const lastCommit = commit[commitIndex].commit;
+    const lastReview = review[reviewIndex];
+
+    const commitDate = lastCommit.committedDate;
+    const reviewDate = lastReview.publishedAt;
+
+    if (commitDate < reviewDate) {
+      logger.info('No new commits after last review');
```
joyeecheung: I think we can just skip this part, only warn when there is new commits, otherwise keep the noise to the minimum.

lib/pr_checker.js
```diff
@@ -211,6 +252,36 @@ class PRChecker {
...
+    }
+
+    function warn() {
+      const multipleCommits = commit.length - 1;
+      const pullURL = lastReview.url.split('#')[0];
+      const commitURL = `${pullURL}/commits/${lastCommit.oid}`;
+      logger.warn('Changes were pushed since the last review:');
+      if (multipleCommits >= 1) {
+        logger.warn(`see ${pullURL}/commits to see all the commits`);
+        logger.warn(`see ${pullURL}/files to see all the changes`);
```
joyeecheung: I think we don't need this URL because it shows all the changes instead of the changes added later.

lib/pr_checker.js
```diff
@@ -211,6 +252,36 @@ class PRChecker {
...
+      warn();
+    }
+
+    function warn() {
+      const multipleCommits = commit.length - 1;
+      const pullURL = lastReview.url.split('#')[0];
+      const commitURL = `${pullURL}/commits/${lastCommit.oid}`;
+      logger.warn('Changes were pushed since the last review:');
+      if (multipleCommits >= 1) {
+        logger.warn(`see ${pullURL}/commits to see all the commits`);
```
joyeecheung: Maybe just log the commit titles into the console instead? If those are all nits, the lander will usually just ignore it. If there are anything alarming, they can do a `git log` and `git show` to see the actual changes locally instead of going to Github (I suppose people usually use this when they already fetch the patch to the disk). If they don't have the patch on disk, then chances are they already open the browser tab so don't need this URL.

lib/pr_checker.js
```diff
@@ -211,6 +252,36 @@ class PRChecker {
...
+        logger.warn(`see ${pullURL}/files to see all the changes`);
+      } else {
+        logger.warn(lastCommit.message.split('\n')[0]);
+        logger.warn(`see ${commitURL} for more info.`);
```
joyeecheung: Same as above.

lib/pr_checker.js
joyeecheung: Also we can suggest people to change their workflow to work better with this tool: only rebase when there are conflicts/relavent breaking updates/it stalls for a long time. Otherwise push new commits with readable titles like `fix nits` `fix styles` `[fixup] <previous commit>` instead of `address comments`.

joyeecheung:
Oh, this reminds me of another TODO: check commits after the last CI run!

apapirovski:
I would like to make sure we standardize around whether we use Prettier or not. I'm noticing a lot of formatting changes in some PRs and IMO it does make it a bit hard to follow the changes. The other reason is that if some people use it and others don't then we'll end up with a really messy history.

Tiriel:
@apapirovski an `npm run lint` script is included in the package. It must be run before committing. I made the same mistake :D

apapirovski:
@Tiriel That's not the issue, the eslint rules for this project are very minimal. These are a result of https://github.com/prettier/prettier.

cPhost:
@apapirovski @Tiriel i came to notice that the code is changed on save that because of my editor as it tried to format code on save. Either is because we don't have enough rules in .eslintrc or something with my editor i disable those setting for now. If With the fixes requested i would update the style to normal.

apapirovski:
@cPhost What editor do you use? I suspect you have something like prettier-atom or its equivalent for another editor: https://atom.io/packages/prettier-atom

cPhost:
@apapirovski it online c9 editor.

Tiriel:
@apapirovski my bad, didn't mean to offense. 

Anyway, that can go with the broader discussion about linting rules in automation. But I'm with you on the necessity!

apapirovski:
@cPhost Interesting. It seems to have some sort of a beautifier that isn't prettier but I can't find much info re: how to disable it. :(

@Tiriel No, no it's fine. I just want to figure this out because there are two PRs open right now both adjusting formatting quite substantially.

apapirovski:
@cPhost Actually, looks like the option would be in `Project > Code Formatters > Format Code on Save`.

cPhost:
@apapirovski that is what i used and i think that what caused it! But i mean it also has option for using .eslintrc and maybe it is because we don't have a rule for it. 

I agree that it bloats the code and make it lot harder to review. If lots of pr came like this i think we should go ahead and open and issue.

apapirovski:
> But i mean it also has option for using .eslintrc and maybe it is because we don't have a rule for it.

I think it's because `eslint` doesn't necessarily have an issue with this new formatting (and neither do I, to be clear). There's nothing inherently wrong with it, it's more that it introduces a lot of churn in the codebase.

lib/pr_checker.js
```diff
@@ -34,6 +35,10 @@ class PRChecker {
     this.reviewers = reviewers;
     this.pr = pr;
     this.comments = comments;
+    // this.review and this.commits must
```
apapirovski: Nit: `this.reviews`

lib/pr_checker.js
```diff
@@ -230,6 +239,33 @@ class PRChecker {
...
+  checkCommitsAfterReview(commits, reviews, logger) {
+    const commitIndex = commits.length - 1;
+    const reviewIndex = reviews.length - 1;
+    const lastCommit = commits[commitIndex].commit;
+    const lastReview = reviews[reviewIndex];
+
+    const commitDate = lastCommit.committedDate;
+    const reviewDate = lastReview.publishedAt;
+    if (commitDate > reviewDate) {
+      warn();
```
apapirovski: I would prefer if both warn and logCommitTitles were unwound (that is, their contents were just within this if statement). Since they're not called more than once, they're just adding to the call stack unnecessarily and each is creating an extra closure.

test/fixtures/commits_after_review.json
```diff
@@ -0,0 +1,56 @@
...
+      "bodyText": "Good idea",
+      "state": "APPROVED",
+      "author": {
+        "login": "foo"
+      },
+      "url": "https://github.com/nodejs/node/pull/16438#pullrequestreview-89923489",
+      "publishedAt": "2017-09-23T11:19:25Z"
+    }]
+  }
+}
```
apapirovski: Could you add a new line here? (We need a lint rule for this, I know.)

cPhost:
@apapirovski @joyeecheung fixed all the changes requested.

lib/pr_checker.js
```diff
@@ -34,6 +35,10 @@ class PRChecker {
     this.reviewers = reviewers;
     this.pr = pr;
     this.comments = comments;
+    // this.review and this.commits must
+    // be in order as recived from github api
```
apapirovski: Nit: `received`

lib/pr_checker.js
```diff
@@ -51,7 +56,11 @@ class PRChecker {
     if (this.authorIsNew()) {
       status.push(this.checkAuthor());
     }
-    // TODO: maybe invalidate review after new commits?
+
+    if (this.reviews.length >= 1) {
```
apapirovski: Nit: could this compare to > 0?

test/fixtures/commits_after_review.json
cPhost: @apapirovski i don't think so for `json`, every other `json` does not have `EOL`. But i added it either way as it looks better.

apapirovski: That's something we'll need to fix. On unix systems a line is only complete if it ends with a new line.

lib/pr_checker.js
```diff
@@ -230,6 +238,25 @@ class PRChecker {
     // 3. is not authored by the people opening the PR
     return true;
   }
+
+  checkCommitsAfterReview(commits, reviews, logger) {
```
joyeecheung: Hmm wait actually, we don't need to pass them here, those are supposed to get assigned in the constructor

test/fixtures/commits_after_review.json
```diff
@@ -0,0 +1,56 @@
+{
+  "singleCommit": {
```
joyeecheung: I would prefer to split this JSON into 4 pieces, `single_commit_after_review_commits.json` & `single_commit_after_review_reviews.json` & `multiple_commits_after_review_commits.json` & `multiple_commits_after_review_commits.json` just like what we do for  the approving reviews/reviewers fitures.

test/unit/pr_checker.test.js
```diff
@@ -300,4 +301,61 @@ describe('PRChecker', () => {
...
+        trace: [],
+        error: []
+      };
+
+      const checker = new PRChecker(logger, {
+        pr: firstTimerPR,
+        reviewers: allGreenReviewers,
+        comments: commentsWithLGTM,
+        reviews: approvingReviews,
+        commits: oddCommits,
```
joyeecheung: Can you pass the fixtures here instead of passing them into `checkCommitsAfterReview`?

cPhost:
@joyeecheung hopefully this is good to go.

test/unit/pr_checker.test.js
```diff
@@ -300,4 +302,92 @@ describe('PRChecker', () => {
...
+
+  describe('checkCommitAfterReview', () => {
+    let logger = null;
+
+    beforeEach(() => {
+      logger = new TestLogger();
+    });
+
+    afterEach(() => {
+      logger.clear();
```
joyeecheung: If we are going to create a new logger for every tests case we don't need to clean them up.

test/unit/pr_checker.test.js
```diff
@@ -300,4 +302,92 @@ describe('PRChecker', () => {
       assert.deepStrictEqual(logger.logs, expectedLogs);
     });
   });
+
+  describe('checkCommitAfterReview', () => {
```
joyeecheung: Can you add a stub in ~L43 and check that `checkCommitsAfterReview` is actually called by `checkAll`?

cPhost:
@joyeecheung done. Added stubs and i meant to clear logs after every `it` so don't have to create new `logger` every time. I got that idea for other PR so copy...paste and forgot to edit it. 😄

lib/pr_checker.js
```diff
@@ -51,7 +55,8 @@ class PRChecker {
     if (this.authorIsNew()) {
       status.push(this.checkAuthor());
     }
-    // TODO: maybe invalidate review after new commits?
+
+    this.checkCommitsAfterReview();
```
apapirovski: Currently this doesn't affect the `status` that checkAll returns (and also has no status of its own). I also think it should very likely be checked right after `checkReviews` rather than after all those other ones... but maybe I'm wrong.

lib/pr_checker.js
```diff
@@ -230,6 +235,33 @@ class PRChecker {
...
     return true;
   }
+
+  checkCommitsAfterReview() {
+    const {
+      commits, reviews, logger
+    } = this;
+
+    if (reviews.length < 1) {
+      return;
```
apapirovski: `return true;`

lib/pr_checker.js
```diff
@@ -230,6 +235,33 @@ class PRChecker {
...
+    if (commitDate > reviewDate) {
+      logger.warn('Changes were pushed since the last review:');
+      commits.forEach((commit) => {
+        commit = commit.commit;
+        if (commit.committedDate > reviewDate) {
+          logger.warn(commit.message.split('\n')[0]);
+        }
+      });
+    }
+  }
```
apapirovski: `return status;`

lib/pr_checker.js
```diff
@@ -230,6 +235,33 @@ class PRChecker {
...
+    const lastReview = reviews[reviewIndex];
+
+    const commitDate = lastCommit.committedDate;
+    const reviewDate = lastReview.publishedAt;
+    if (commitDate > reviewDate) {
+      logger.warn('Changes were pushed since the last review:');
+      commits.forEach((commit) => {
+        commit = commit.commit;
+        if (commit.committedDate > reviewDate) {
+          logger.warn(commit.message.split('\n')[0]);
```
apapirovski: `status = false;`

lib/pr_checker.js
```diff
@@ -230,6 +235,33 @@ class PRChecker {
...
+
+  checkCommitsAfterReview() {
+    const {
+      commits, reviews, logger
+    } = this;
+
+    if (reviews.length < 1) {
+      return;
+    }
+
```
apapirovski: `let status = true;`

cPhost:
Just want to know whats the point for the status. Is is just for the tests?

apapirovski:
Right now? No specific point. Long-term it's going to allow us to do stuff like determine whether land-pr can run or whether we need user input first to address some of the issues, etc.

joyeecheung:
@cPhost I like the idea of `(commit)`, although for consistency with the error logs, maybe `[COMMIT]`? Also we need to sync with https://github.com/joyeecheung/node-core-utils/pull/69 on this.

apapirovski:
I'm still not a big fan of any preface for the commits. I mean it's pretty clear they're commits, especially since our commits have specific format... Just makes the text longer. I could get behind [COMMIT] if it replaced [INFO] or [WARN]...

cPhost:
@joyeecheung the question is would this log be fine 
`[WARN] [COMMIT]`

joyeecheung:
@apapirovski @cPhost What about bullet points?

```
[WARN] Changes were pushed since the last review:
[WARN] - src: fix xxx
[WARN] - test: fix xxx
```

cPhost:
@joyeecheung i just need some clarification on
> Also we need to sync with #69 on this.

do you mean log only last couple of commit to make less noise.

apapirovski:
@cPhost just that we want the format to be consistent between the two PRs. So we'll need to have that one also use bullet points.

cPhost:
@apapirovski @joyeecheung done now logs come out as 
```
[WARN] Changes were pushed since the last review:
[WARN] - fixup: new changes requested
[WARN] - prchecker: new format
```

PR #69, opened 2017-11-04
feat: check commits after the last ci run

buji1993:
Fixes: https://github.com/joyeecheung/node-core-utils/issues/66


lib/pr_checker.js
```diff
@@ -172,6 +203,13 @@ class PRChecker {
     for (const [type, ci] of ciMap) {
       const name = CI_TYPES.get(type).name;
       logger.info(`Last ${name} CI on ${ci.date}: ${ci.link}`);
+      commits.forEach((c) => {
```
apapirovski: So the logic here should be that we only log commits that came after the latest CI run. This code would log commits that came after each CI run type so it might show a lot of duplicates and just generally unhelpful data.

lib/pr_checker.js
```diff
@@ -172,6 +203,13 @@ class PRChecker {
...
+        const commit = c.commit;
+        console.log(commit);
```
apapirovski: This looks like it should be removed?

lib/pr_checker.js
```diff
@@ -172,6 +203,13 @@ class PRChecker {
...
+        if (ci.date < commit.committedDate) {
+          logger.warn(`commit(${commit.message}) after Last ${name} CI`);
```
apapirovski: It might be more helpful if this took a format such as:

```
[WARN] Commits pushed after last CI run:
[WARN] fixup: adjust spelling
[WARN] doc: add documentation of changes
```

Basically an intro and then just output the commit title (first line).

buji1993:
@apapirovski thanks for your suggestion , I will fix it later.

test/fixtures/commits_after_ci.json
```diff
@@ -0,0 +1,16 @@
...
+        "author": {
+          "login": "bar"
+        }
+      }
+    }],
+    "comment": [{
+      "bodyText": "CI: https://ci.nodejs.org/job/node-test-pull-request/10984/",
+      "publishedAt": "2017-10-24T11:19:25Z"
+    }]
+  }
```
apapirovski: Could you add a new line here? (We probably should have a lint rule for this.)

lib/pr_checker.js
```diff
@@ -169,9 +169,30 @@ class PRChecker {
...
+        lastCiDate = ci.date;
+      }
+    }
+
+    const commitsAfterCi = [];
+    if (lastCiDate) {
+      commits.forEach((c) => {
+        const commit = c.commit;
+        if (commit.committedDate > lastCiDate) {
+          commitsAfterCi.push(commit);
```
apapirovski: I believe this could be refactored to be more like:

```js
if (lastCiDate) {
  let didLogHeader = false;
  commits.forEach((c) => {
    const commit = c.commit;
    if (commit.committedDate > lastCiDate) {
      if (!didLogHeader) {
        logger.warn('Commits pushed after the last CI run:');
        status = false;
        didLogHeader = true;
      }
      logger.warn(commit.message);
    }
  });
}
```

(Note also the addition of `status = false` since these new warnings technically mean that the CI check isn't clean.)

lib/pr_checker.js
```diff
@@ -169,9 +169,30 @@ class PRChecker {
       logger.warn('No full CI runs detected');
     }
 
+    let lastCiDate = null;
```
apapirovski: Nit: could this be `let lastCiDate;` — I don't think it strictly needs to be `null`.

test/fixtures/commits_after_ci.json
cPhost: @apapirovski just so you know every other `.json` files does not have EOL.

lib/pr_checker.js
buji1993: if no ci in that pr, the lastCiDate could be null.

lib/pr_checker.js
buji1993: right , I will fix it

lib/pr_checker.js
cPhost: @buji1993 since lastCiDate is not needed to be null has no use in this case and therefore not be assigned it make code simpler.

joyeecheung: @buji1993 I think @apapirovski meant we can just leave it as `undefined` because it's falsy as well.

Before we sort out how to warn about different types of CI runs, I think we should at least change the logic here a bit and make it a `lastCI` which is "the oldest CI (regardless of type) that covers every commit", and then put the CI type in warning.

lib/pr_checker.js
```diff
@@ -169,9 +169,28 @@ class PRChecker {
...
+
+    if (lastCiDate) {
+      let didLogHeader = false;
+      commits.forEach((c) => {
+        const commit = c.commit;
+        if (commit.committedDate > lastCiDate) {
+          if (!didLogHeader) {
+            status = false;
+            didLogHeader = true;
+            logger.warn('Commits pushed after the last CI run:');
```
joyeecheung: Hmm, on a second thought, we should probably warn about commits after each type of CI, e.g. 2 commits after last CITGM CI, 3 commits after the last full CI, although we need to take into account that some CI cover the others as well (e.g. full CI covers linter CI)..no need to tackle this in this PR though.

lib/pr_checker.js
buji1993: @joyeecheung I think either lastCI or oldestCI is not right in this situation. As you said some CI covers others , these CIs are not individual. now I use the oldestCI in ciMap and log the oldest ci type. FYI

lib/pr_checker.js
```diff
@@ -169,9 +169,31 @@ class PRChecker {
...
+
+    if (lastCI) {
+      let didLogHeader = false;
+      commits.forEach((c) => {
+        const commit = c.commit;
+        if (commit.committedDate > lastCI.date) {
+          if (!didLogHeader) {
+            status = false;
+            didLogHeader = true;
+            logger.warn(`Commits pushed after the last ${lastCI.typeName} CI run:`);
```
joyeecheung: Logging this line in every iteration would bring too much noise, maybe just filter out the new commits and do the logging according to the length of the array?

test/fixtures/commits_after_ci.json
```diff
@@ -0,0 +1,16 @@
+  {
+    "commits": [{
+      "commit": {
+        "committedDate": "2017-10-25T11:27:02Z",
```
joyeecheung: Can you extend this fixtures a bit and add one more commits? Just so we can see how it's gonna look like when there are multiple new commits.

lib/pr_checker.js
apapirovski: This logs just once due to `didLogHeader`.

joyeecheung: @apapirovski Ah right, I knew there should be a reason I didn't raise this question before lol

test/fixtures/commits_after_ci.json
buji1993: ok

test/fixtures/commits_after_ci.json
apapirovski: @buji1993 This still needs a new line. Thanks.

joyeecheung:
@buji1993 Hi, this PR needs a rebase. Also since https://github.com/joyeecheung/node-core-utils/pull/65 has landed, can you add a bullet to each commit so it looks something like:

```
Commits pushed after the last Full CI run:
- doc: doc changes
- src: some changes
```

? After that I think we'll be good to merge. Thanks!

buji1993:
@joyeecheung done

lib/pr_checker.js
```diff
@@ -175,9 +175,32 @@ class PRChecker {
...
+      commits.forEach((c) => {
+        const commit = c.commit;
+        if (commit.committedDate > lastCI.date) {
+          if (!didLogHeader) {
+            status = false;
+            didLogHeader = true;
+            logger.warn(`Commits pushed after the last ${lastCI.typeName} ` +
+            'CI run:');
+          }
+          logger.warn(`- ${commit.message}`);
```
cPhost: @buji1993 if you log `commit.message` it would log body of message to.
Nit: `commit.message.split('\n')[0]`

buji1993: ok. thanks for your suggestion

PR #70, opened 2017-11-04
fix the parsing of fixes

apapirovski:
This fixes a very serious bug in parsing & creating Fixes URLs. Noticed it while landing: https://github.com/nodejs/node/commit/291ff72f859df3c4c3849021419f37cd542840a6

This should warrant a new release on npm since this makes `get-metadata` very problematic to use.

### DIFF
--- a/lib/links.js
+++ b/lib/links.js
@@ -10,19 +10,20 @@ const { JSDOM } = require('jsdom');
  * Most of this class is ported from node-review
  */
 class LinkParser {
-  constructor(repo, html) {
+  constructor(owner, repo, html) {
+    this.owner = owner;
     this.repo = repo;
     this.OP = JSDOM.fragment(html);
   }
 
   getFixesUrlsFromArray(arr) {
-    const repo = this.repo;
+    const { owner, repo } = this;
     const result = [];
     for (const item of arr) {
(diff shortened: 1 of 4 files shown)


cPhost:
Great work!

lib/links.js
```diff
@@ -10,19 +10,20 @@ const { JSDOM } = require('jsdom');
...
   getFixesUrlsFromArray(arr) {
-    const repo = this.repo;
+    const { owner, repo } = this;
     const result = [];
     for (const item of arr) {
       const m = item.match(FIX_RE);
       if (!m) continue;
       const fix = m[3];
-      const url = fix.replace(/^#/, `${repo}#`).replace('#', '/issues/');
+      const url = fix.replace(/^#/, `${owner}/${repo}#`).replace('#', '/issues/');
```
joyeecheung: Wait..can we use `getRefUrlFromOP()` for this? That one looks more reliable actually.

joyeecheung: Also this will skip a cross-repository fix, that one seems to solve it as well.

joyeecheung:
So I think we should merge this now and release soon, we can address https://github.com/joyeecheung/node-core-utils/pull/70#discussion_r148946435 later.

PR #71, opened 2017-11-04
add new lines to ends of all files

apapirovski:
What it says...

> You get a new line and you get a new line... eeeeeverybody gets a new line.


joyeecheung:
@cPhost A better way would be to use eslint plugins to check JSONs..

joyeecheung:
BTW I think this is good to merge, but @apapirovski you can decide if you want to do it here or open another one. I've sent you collaborator invites, feel free to merge this yourself!

cPhost:
@joyeecheung i will open an issue and look into the eslint json plugin.

apapirovski:
I tried using the eslint json plugin out there when I made this. It didn't seem to really validate this type of thing though :| Not sure what can be done... On Node core we just check this as part of the review flow and request changes.

Issue #67, opened 2017-11-04
todo: account LGTM in "COMMENTED" reviews


joyeecheung:
Example: https://github.com/nodejs/node/pull/16248#pullrequestreview-74093431

Tiriel:
Just to be sure, you want to count the number of "LGTM" in the comments of a review?

Anyway, I'm taking this one!

joyeecheung:
@Tiriel I mean the LGTM in a COMMENTED review should count as an approval ;)

Tiriel:
Oh yeah I see, sorry. 

Got it! I'll assign myself if it's ok

Tiriel:
Landed in #72 !

PR #72, opened 2017-11-04
Fixing review state to APPROVED whe 'LGTM' in COMMENTED review

Tiriel:
From #67 
Tested with nodejs/node#16248 as given in issue #67 

Ran lint and tests.

Comments and reviews deeply welcomed!

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Options:
   --owner, -o       GitHub owner of the PR repository                   [string]
   --repo, -r        GitHub repository of the PR                         [string]
   --file, -f        File to write the metadata in                       [string]
-  --check-comments  Check for'LGTM' in comments                        [boolean]
+  --check-comments  Check for 'LGTM' in comments                        [boolean]
   --help, -h        Show help                                          [boolean]
 ```
 
--- a/README.md
+++ b/README.md
@@ -52,11 +52,12 @@ get-metadata <identifier>
 Retrieves metadata for a PR and validates them against nodejs/node PR rules
 
 Options:
-  --version    Show version number                                     [boolean]
-  --owner, -o  GitHub owner of the PR repository                        [string]
-  --repo, -r   GitHub repository of the PR                              [string]
-  --file, -f   File to write the metadata in                            [string]
-  --help, -h   Show help                                               [boolean]
+  --version         Show version number                                [boolean]
+  --owner, -o       GitHub owner of the PR repository                   [string]
+  --repo, -r        GitHub repository of the PR                         [string]
+  --file, -f        File to write the metadata in                       [string]
+  --check-comments  Check for'LGTM' in comments                        [boolean]
+  --help, -h        Show help                                          [boolean]
 ```
 
 Examples:
--- a/lib/args.js
+++ b/lib/args.js
@@ -39,7 +39,7 @@ function buildYargs(args = null) {
     })
     .option('check-comments', {
       demandOption: false,
-      describe: 'Check for\'LGTM\' in comments',
+      describe: 'Check for \'LGTM\' in comments',
       type: 'boolean'
     })
     .help()
--- a/lib/args.js
+++ b/lib/args.js
@@ -37,6 +37,12 @@ function buildYargs(args = null) {
       describe: 'File to write the metadata in',
       type: 'string'
     })
+    .option('comments', {
+      alias: 'c',
+      demandOption: false,
+      describe: 'Check for\'LGTM\' in comments',
+      type: 'boolean'
+    })
     .help()
     .alias('help', 'h')
     .argv;
@@ -47,8 +53,8 @@ const PR_RE = new RegExp(
   '([0-9]+)(?:/(?:files)?)?$');
 
 function checkAndParseArgs(args) {
-  const { owner = 'nodejs', repo = 'node', identifier, file } = args;
-  const result = { owner, repo, file };
+  const { owner = 'nodejs', repo = 'node', identifier, file, comments } = args;
+  const result = { owner, repo, file, comments };
   if (!isNaN(identifier)) {
     result.prid = +identifier;
   } else {
--- a/lib/args.js
+++ b/lib/args.js
@@ -37,8 +37,7 @@ function buildYargs(args = null) {
       describe: 'File to write the metadata in',
       type: 'string'
     })
-    .option('comments', {
-      alias: 'c',
+    .option('check-comments', {
       demandOption: false,
       describe: 'Check for\'LGTM\' in comments',
       type: 'boolean'
@@ -53,8 +52,10 @@ const PR_RE = new RegExp(
   '([0-9]+)(?:/(?:files)?)?$');
 
 function checkAndParseArgs(args) {
-  const { owner = 'nodejs', repo = 'node', identifier, file, comments } = args;
-  const result = { owner, repo, file, comments };
+  const {
+    owner = 'nodejs', repo = 'node', identifier, file, checkComments
+  } = args;
+  const result = { owner, repo, file, checkComments };
   if (!isNaN(identifier)) {
     result.prid = +identifier;
   } else {
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -11,7 +11,7 @@ const WEEKDAY_WAIT = 48;
 const WEEKEND_WAIT = 72;
 
 const {
-  REVIEW_SOURCES: { FROM_COMMENT }
+  REVIEW_SOURCES: { FROM_COMMENT, FROM_REVIEW_COMMENT }
 } = require('./reviews');
 const {
   FIRST_TIME_CONTRIBUTOR, FIRST_TIMER
@@ -45,9 +45,9 @@ class PRChecker {
     );
   }
 
-  checkAll() {
+  checkAll(comments = false) {
     const status = [
-      this.checkReviews(),
+      this.checkReviews(comments),
       this.checkCommitsAfterReview(),
       this.checkPRWait(new Date()),
       this.checkCI()
@@ -75,7 +75,7 @@ class PRChecker {
     return hint;
   }
 
-  checkReviews() {
+  checkReviews(comments = false) {
     const {
       pr, logger, reviewers: { rejected, approved }
     } = this;
@@ -98,10 +98,13 @@ class PRChecker {
       let hint = this.getTSCHint(approved);
       logger.info(`Approvals: ${approved.length}${hint}`);
 
-      for (const { reviewer, review } of approved) {
-        if (review.source === FROM_COMMENT) {
-          logger.info(
-            `${reviewer.getName()}) approved in via LGTM in comments`);
+      if (comments) {
+        for (const {reviewer, review} of approved) {
+          if (review.source === FROM_COMMENT ||
+            review.source === FROM_REVIEW_COMMENT) {
+            logger.warn(
+              `${reviewer.getName()}) approved in via LGTM in comments`);
+          }
         }
       }
 
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -77,7 +77,7 @@ class PRChecker {
 
   checkReviews() {
     const {
-      pr, logger, reviewers: { rejected, approved }
+      pr, logger, reviewers: { rejected, approved, commentApproved }
     } = this;
     let status = true;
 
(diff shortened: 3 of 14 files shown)


cPhost:
@Tiriel Great Work! can you add tests in `test/unit/reviews.test.js`

Tiriel:
Oh yeah sorry, that's indeed a good idea. I'm on it!

Tiriel:
Original tests were in fact all that was needed. I just had to add a test-case in fixtures for my first draft to make some tests fail. (pr_data tests in fact).

Corrected work, now tests still pass with proper test case, and command still responds properly.

lib/reviews.js
```diff
@@ -132,14 +138,31 @@ class ReviewAnalyzer {
...
+    return review.state === COMMENTED && this.hasLGTM(review, 'ref');
+  }
+
+  /**
+   * @param object
+   * @param prop: string
+   * @returns {boolean}
+   */
+  hasLGTM(object, prop) {
+    return LGTM_RE.test(object[prop]);
```
refack: So this is too loose. It would pass for `I'm revoking my LGTM`.
IMHO this should be `object[prop].toLowerCase() === 'lgtm'`.

joyeecheung: @refack This is already the case for normal comments, I think we should log an info here, like what we do for those, to remind the landers to double-check. The actual context would be too hard to parse without some kind of natural language processing..

refack: Good idea!

joyeecheung: OK not log an info here, but make a new type of `REVIEW_SOURCES`, maybe `FROM_REVIEW_COMMENT`, and warn them in `checkReviews` like what we do for `FROM_COMMENT`

joyeecheung: @refack re: `I'm revoking my LGTM.` they would've put an red cross there I think?

refack: I'm thinking of something I saw last week - https://github.com/nodejs/node/pull/16612#pullrequestreview-73478662

refack: That case was not a "Request Changes" , just "Don't count me as approver", same as dismissed review.

refack:
BTW: people can also dismiss other's reviews, so need to keep track of the author attribution

joyeecheung:
@refack Yes, we probably need to dig a bit more into the state machine later as well..

lib/reviews.js
joyeecheung: Another idea: if we are uncertain about a LGTM, we should log it as `?Reviewed-By: xxx`, and then core-validate-commit should complain about it.

Tiriel:
By the way, I see that the info telling one PR was approved in comments or commented review is on `console.info`. Wouldn't it be better to pass it to `console.warn`?

Tiriel:
By the way, I see that the info telling one PR was approved in comments or commented review is on `console.info`. Wouldn't it be better to pass it to `console.warn`?

Tiriel:
For now, that's what I got:
image

Tiriel:
Okay, I've read all the comments, bit confusing xD

I think I'll go for both things you suggested if possible @joyeecheung , new source and `?`in report.

@joyeecheung @refack Otherwise, what about changing the RegEx to this: `^(L|S)GTM!?$` ? This would prevent most cases of matches when people are discussing "LGTM" without actually wanting to validate a PR and still allowing some leeway, don't you think?

apapirovski:
I know I'm going to be a minority on this but I would scrape the whole LGTM checking. Everyone uses the green button. If they're not using and they say 'LGTM' or something that extent, it usually means they scanned it but don't have complete confidence to give it firm approval.

apapirovski:
For example, what does this count as?

> Basically LGTM, just note that automatic fixing can introduce bugs if another variable of the same name was declared within the function.

I don't think the author would expect to be listed in `Reviewed-By` even though they gave a LGTM.

Tiriel:
That's a good point, but actually that's why I would put the notice in warning.

In a way, just to say "Some people seem to approve, but be careful".

joyeecheung:
Or....maybe we should skip those people in the actual metadata, just log them as "potential reviewers"?

apapirovski:
@joyeecheung I think that's my favourite suggestion so far. I think any LGTM couldbe be included as ```
Approved: 1, 1 from TSC (mhdawson)
N
```

**Comment by apapirovski** [comment]
Fat fingered the close button. Sorry.

**Comment by Tiriel** [comment]
Maybe then keeping the line "XXX reviewed in comment" while removing the review from review count?

**Comment by joyeecheung** [comment]
@Tiriel Yes that's what I had in mine.

**Comment by joyeecheung** [comment]
@Tiriel Or maybe..

```
The following collaborators are not listed in the metadata:
Foo <foo@example.org>: I am not sure about if I can give a proper LGTM but this is a great idea
Bar <bar@example.org>: LGTM modulo style nits
```

(we can even make this interactive and let the user decide!...although for the moment I think this PR is good with just removing them in the metadata)

**Comment by Tiriel** [comment]
Form now that's what I got:
image
Is that okay?

**Comment by Tiriel** [comment]
I can do like you said in your last comment, but I feel that would need a more important rework of the workflow that would justify another PR

**Comment by Tiriel** [comment]
Actually this way other tests fail.

Going to rework already :)

**Comment by Tiriel** [comment]
Hey!

Finally reworked a bit.
I needed to adapt the tests, since the LGTM in comments don't count as approvals now.
The result looks like this:
image

Hope it's ok now!

File: lib/pr_checker.js [hunk 2]
```diff
@@ -77,7 +77,7 @@ class PRChecker {
 
   checkReviews() {
     const {
-      pr, logger, reviewers: { rejected, approved }
+      pr, logger, reviewers: { rejected, approved, commentApproved }
```
joyeecheung: I don't think we need the extra export here, the information is already in `approved[i].review.source`, so we can derive data from the source of truth here.

lib/pr_checker.js
```diff
@@ -95,8 +95,12 @@ class PRChecker {
       status = false;
       logger.warn(`Approvals: 0`);
```
joyeecheung: We can build something like `approvalFromReview` and use that from here, so we don't need to do the filtering in `getTSCHint`.

lib/pr_checker.js
```diff
@@ -95,8 +95,12 @@ class PRChecker {
...
     } else {
+      let notComm = approved.length;
+      approved.map((r) => {
+        if (r.review.source !== FROM_REVIEW) { notComm--; }
+      });
       let hint = this.getTSCHint(approved);
-      logger.info(`Approvals: ${approved.length}${hint}`);
+      logger.info(`Approvals: ${notComm}${hint}`);
 
       for (const { reviewer, review } of approved) {
```
joyeecheung: This can be merged with `review.source === FROM_REVIEW_COMMENT`

lib/reviews.js
joyeecheung: On a second thought, I am finding the `object[prop].toLowerCase() === 'lgtm'` approach better (although we can just do that by changing `LGTM_RE` to `/^lgtm$/i` and trim the comments before testing them), it's just simpler that way. I think people who don't give LGTM either this way or with a green review would not mind not being listed in the review-by list.

lib/pr_checker.js
Tiriel: Well, I thought so at first, but that's what was breaking the tests, and would have need it to revise them completely.
Plus, IMHO, since I've separated things, I think it's semantically weird to mix properly approved reviews and commented reviews with LGTM. I mean, that's the whole point of getting these reviews out of the metadatas.

lib/pr_checker.js
Tiriel: You mean the two blocks nearly repeating themselves? It probably needs refactoring yes, but I wanted advice on the workflow yet :)

lib/pr_checker.js
Tiriel: Going to look into this!

joyeecheung:
@Tiriel Sorry for so much back-and-forth, but I actually kinda see the point of https://github.com/joyeecheung/node-core-utils/pull/72#discussion_r148944765 now, we might as well just ignore the reviews that is not just `^lgtm$/i`, in that case we don't really need to warn about anything because it's pretty clear that those are legitimate sign-offs.

Tiriel:
No prob! I'll try and rollback to the previous version then.
What do you prefer? `r.bodyText.toLowerCase() === 'lgtm'`, or changing the regex to `/^lgtm\W?$/gmi`?

Personnally I prefer modifying the regex, I've seen people adding punctuation to LGTMs and making a strict equal comparison would count them.

joyeecheung:
@Tiriel Let's go with regexp then. Also we might want to trim the text before testing them as well.

Tiriel:
Done!
Linted and tested as usual (now) :D

lib/reviews.js
```diff
@@ -80,6 +81,12 @@ class ReviewAnalyzer {
             new Review(r.state, r.publishedAt, r.url, FROM_REVIEW)
           );
           break;
+        case COMMENTED:
+          map.set(
+            login,
+            new Review(r.state, r.publishedAt, r.bodyText, FROM_REVIEW_COMMENT)
```
joyeecheung: I suggest we just pass `APPROVED` as the state here, like what we do for approval in normal comments. This way we don't need to call `isApprovedInComment` in `getReviewers`. We still keep the info about its source in `review.source`.

test/fixtures/reviews_approved.json
```diff
@@ -36,6 +36,15 @@
   "publishedAt": "2017-10-24T14:49:52Z"
 },
 {
+  "bodyText": "LGTM",
+  "state": "COMMENTED",
+  "author": {
+    "login": "Baz"
```
joyeecheung: Can you make a new case for this? This and the case above are overlapping each other so one of them would not be visible in the tests. To do this we need to add another collaborator in `collaborators.json` and the readme in the fixtures as well.

joyeecheung:
Also with this we can probably drop the check for `FROM_COMMENT` in the pr checker, or put it behind a command line flag and warn about both `FROM_COMMENT` and `FROM_REVIEW_COMMENT` (but I don't think we really need a warning for that any more..)

lib/reviews.js
Tiriel: Oh thanks, that's what I had done in the first place, forgot to reinstate it. when reverting

test/fixtures/reviews_approved.json
Tiriel: Sure thing!

Tiriel: There are in fact numerous tests using each files. I also had to add an entry in reviewers_approved.json and modify accordingly other tests. Push coming

Tiriel:
There, this should do it. I added the `--comments` flag and some tests cases. I'll let you review, since I had to adapt some existing tests cases

lib/args.js
```diff
@@ -37,6 +37,12 @@ function buildYargs(args = null) {
       describe: 'File to write the metadata in',
       type: 'string'
     })
+    .option('comments', {
```
joyeecheung: I'd rather let this be `--check-comments` with no aliases, in case we need the `c` later. This does not seem to be very in demand.

test/unit/pr_checker.test.js
```diff
@@ -80,12 +80,13 @@ describe('PRChecker', () => {
 
       const expectedLogs = {
         warn: [
+          ['Quux User(Quux)) approved in via LGTM in comments'],
```
joyeecheung: There is a redundant `)` here..

Tiriel: Actually this redundant parens is in every test and is in fact in  the base template :D
I'll remove it everywhere.

lib/args.js
Tiriel: Your desires are my commands!

Tiriel:
Fixed!

cPhost:
@Tiriel can you add this option in readme.

Tiriel:
Very good idea! Incoming!

lib/args.js
```diff
@@ -37,6 +37,11 @@ function buildYargs(args = null) {
       describe: 'File to write the metadata in',
       type: 'string'
     })
+    .option('check-comments', {
+      demandOption: false,
+      describe: 'Check for\'LGTM\' in comments',
```
cPhost: Same as above^

README.md
```diff
@@ -52,11 +52,12 @@ get-metadata <identifier>
...
-  --version    Show version number                                     [boolean]
-  --owner, -o  GitHub owner of the PR repository                        [string]
-  --repo, -r   GitHub repository of the PR                              [string]
-  --file, -f   File to write the metadata in                            [string]
-  --help, -h   Show help                                               [boolean]
+  --version         Show version number                                [boolean]
+  --owner, -o       GitHub owner of the PR repository                   [string]
+  --repo, -r        GitHub repository of the PR                         [string]
+  --file, -f        File to write the metadata in                       [string]
+  --check-comments  Check for'LGTM' in comments                        [boolean]
```
cPhost: Minor Nit:  add a space between for and LGTM `'Check for \'LGTM\' in comments'`

Tiriel:
Going to merge then, unless someone objects!

PR #77, opened 2017-11-05
Collaborators tests

alopezsanchez:
- `lib/collaborators` tests.
- `reorganice fixtures with a README subfolder`. I created some README files in order to test all possible throw cases.

I also have a question: should [`isCollaborator function`](https://github.com/joyeecheung/node-core-utils/blob/master/lib/collaborators.js#L104) return `undefined`/`collaborator` or it should return `true`/`false` instead?


test/unit/collaborators.test.js
```diff
@@ -1,16 +1,120 @@
 'use strict';
 
+const assert = require('assert');
+
 const {
+  isCollaborator,
   getCollaborators
 } = require('../../lib/collaborators');
-const { readme, collaborators } = require('../fixtures/data');
+const { readme, readmeNoTsc, readmeNoTscE, readmeNoCollaborators, readmeNoCollaboratorE, readmeUnordered, collaborators } = require('../fixtures/data');
```
joyeecheung: Can you wrap this line by breaking after the commas?

test/unit/collaborators.test.js
```diff
@@ -1,16 +1,120 @@
...
+  });
+
+  afterEach(() => {
+    logger.clear();
+  });
+
+  describe('Collaborator', () => {
+    describe('isActor', () => {
+      it('should return false if no actor is specified', () => {
+        assert.strictEqual(collaborator.isActor(), false);
```
joyeecheung: This should be `collaborator.isActor(null)`, github returns the actor as `null` if it's a ghost.

test/unit/collaborators.test.js
```diff
@@ -1,16 +1,120 @@
...
+    logger = new TestLogger();
+  });
+
+  afterEach(() => {
+    logger.clear();
+  });
+
+  describe('Collaborator', () => {
+    describe('isActor', () => {
+      it('should return false if no actor is specified', () => {
```
joyeecheung: `...if actor is a ghost`

test/unit/collaborators.test.js
```diff
@@ -1,16 +1,120 @@
...
+  });
+
+  describe('getCollaborators', () => {
+    it('should return all collaborators', async function() {
+      const parsed = await getCollaborators(readme, logger, 'nodejs', 'node');
+      assert.deepStrictEqual(parsed, collaborators);
+    });
+
+    it('should throw error if there is no TSC section in the README', async() => {
+      await assertThrowsAsync(async() => getCollaborators(readmeNoTsc, logger, 'nodejs', 'node'), /Error: Couldn't find ### TSC \(Technical Steering Committee\) in the README/);
```
joyeecheung: Can you wrap the arguments passed to `assertThrowsAsync`? (and the lines below)

test/unit/collaborators.test.js
```diff
@@ -1,16 +1,120 @@
...
+    });
+
+    it('should throw error if there are no collaborators', async() => {
+      await assertThrowsAsync(async() => getCollaborators(readmeUnordered, logger, 'nodejs', 'node'), /Error: Could not find any collaborators/);
+    });
+  });
+
+  describe('isCollaborator', () => {
+    it('should return undefined if the user has no login (ghost)', () => {
+      assert.strictEqual(isCollaborator(collaborators, { name: 'Ghost' }), undefined);
```
joyeecheung: If it's a ghost then the whole thing would be `null`, the login check is just a safeguard, I don't think Github would ever return an actor without `login`..

alopezsanchez:
Updated with requested changes. I also applied the max line length = 80 (https://github.com/joyeecheung/node-core-utils/pull/81)

test/unit/collaborators.test.js
```diff
@@ -1,16 +1,173 @@
...
+    it('should return all collaborators', async function() {
+      const parsed = await getCollaborators(readme, logger, 'nodejs', 'node');
+      assert.deepStrictEqual(parsed, collaborators);
+    });
+
+    it(
+      'should throw error if there is no TSC section in the README',
+      async() => {
+        await assertThrowsAsync(
+          async() => getCollaborators(readmeNoTsc, logger, 'nodejs', 'node'),
```
cPhost: A minor nit: can you add space between `async` and `()`

test/unit/collaborators.test.js
```diff
@@ -1,16 +1,173 @@
...
+          Error);
+      });
+
+    it(
+      'should throw error if there is no TSC Emeriti section in the README',
+      async() => {
+        await assertThrowsAsync(
+          async() => getCollaborators(readmeNoTscE, logger, 'nodejs', 'node'),
```
cPhost: ^ same as above

test/unit/collaborators.test.js
```diff
@@ -1,16 +1,173 @@
...
+            readmeUnordered, logger, 'nodejs', 'node'),
+          /Error/);
+        assert.strictEqual(logger.logs.warn[0][0],
+          'Order of different types of contacts needs update. ' +
+          'Analysis could get wrong.');
+      });
+
+    it('should throw error if there are no collaborators', async() => {
+      await assertThrowsAsync(
+        async() => getCollaborators(readmeUnordered, logger, 'nodejs', 'node'),
```
cPhost: Same as above^ for all.

test/unit/collaborators.test.js
apapirovski: That wouldn't pass the linter, the current version is correct.

PR #78, opened 2017-11-05
editorconfig: set insert_final_newline to true

alopezsanchez:
Now, editorconfig automatically inserts a final new line when saving a file.

Related to: https://github.com/joyeecheung/node-core-utils/pull/71/ and https://github.com/joyeecheung/node-core-utils/issues/76


apapirovski:
Will let this sit for a few hours and then merge since it's a super minor change.

PR #79, opened 2017-11-05
less processing of metadata in the bin target

apapirovski:
Just small changes to do less manipulation of metadata inside the `get-metadata` target so that we don't need to duplicate this code in any other targets that use metadata.

Also simplified the meta array generation to use spread operators — just lets us keep that code a lot shorter and easier to read.

### DIFF
--- a/bin/metadata.js
+++ b/bin/metadata.js
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 'use strict';
 
-const { EOL } = require('os');
 const Request = require('../lib/request');
 const auth = require('../lib/auth');
 
@@ -28,11 +27,11 @@ async function main(prid, owner, repo, logger) {
   const metadata = new MetadataGenerator(data).getMetadata();
   const [SCISSOR_LEFT, SCISSOR_RIGHT] = MetadataGenerator.SCISSORS;
   logger.info({
-    raw: [SCISSOR_LEFT, metadata, SCISSOR_RIGHT].join(EOL)
+    raw: `${SCISSOR_LEFT}${metadata}${SCISSOR_RIGHT}`
   }, 'Generated metadata:');
 
   if (!process.stdout.isTTY) {
-    process.stdout.write(`${metadata}${EOL}`);
+    process.stdout.write(metadata);
   }
   const checker = new PRChecker(logger, data);
   checker.checkAll();
--- a/lib/metadata_gen.js
+++ b/lib/metadata_gen.js
@@ -32,26 +32,21 @@ class MetadataGenerator {
     const fixes = parser.getFixes();
     const refs = parser.getRefs();
 
-    const output = {
-      prUrl, reviewedBy, fixes, refs
-    };
-
     let meta = [
-      `PR-URL: ${output.prUrl}`
+      `PR-URL: ${prUrl}`,
+      ...fixes.map((fix) => `Fixes: ${fix}`),
+      ...refs.map((ref) => `Refs: ${ref}`),
+      ...reviewedBy.map((r) => `Reviewed-By: ${r.reviewer.getContact()}`),
+      '' // creates final EOL
     ];
-    meta = meta.concat(output.fixes.map((fix) => `Fixes: ${fix}`));
-    meta = meta.concat(output.refs.map((ref) => `Refs: ${ref}`));
-    meta = meta.concat(output.reviewedBy.map((r) => {
-      return `Reviewed-By: ${r.reviewer.getContact()}`;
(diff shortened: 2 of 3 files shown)


bin/metadata.js
```diff
@@ -28,11 +27,11 @@ async function main(prid, owner, repo, logger) {
   const metadata = new MetadataGenerator(data).getMetadata();
   const [SCISSOR_LEFT, SCISSOR_RIGHT] = MetadataGenerator.SCISSORS;
   logger.info({
-    raw: [SCISSOR_LEFT, metadata, SCISSOR_RIGHT].join(EOL)
+    raw: `${SCISSOR_LEFT}${metadata}${SCISSOR_RIGHT}`
```
Tiriel: While I understand the intent to prevent too much data manipulation inside the bin, I'm not fond of this particular change, I find it less readable. But maybe that's because I'm less experienced.

lib/metadata_gen.js
```diff
@@ -32,26 +32,21 @@ class MetadataGenerator {
...
-      prUrl, reviewedBy, fixes, refs
-    };
-
     let meta = [
-      `PR-URL: ${output.prUrl}`
+      `PR-URL: ${prUrl}`,
+      ...fixes.map((fix) => `Fixes: ${fix}`),
+      ...refs.map((ref) => `Refs: ${ref}`),
+      ...reviewedBy.map((r) => `Reviewed-By: ${r.reviewer.getContact()}`),
+      '' // creates final EOL
```
Tiriel: Same here.

Il love the use of the spread operator, but the last line puzzles me. Isn't there another way to create the final EOL without this line and comment?

bin/metadata.js
apapirovski: It's just a template literal: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals — there should be no need to create an array for something so basic.

lib/metadata_gen.js
apapirovski: Just imagine that `.join(str)` puts `str` between all elements in the array so [1, 2, 3] becomes `1str2str3`. When you instead make it [1, 2, 3, ''] it'll instead do `1str2str3str` since there's now an extra empty element.

bin/metadata.js
Tiriel: Well, the template litteral I know of, but tbh, at first reading I was wondering how you were gonna include the EOL., that's the only thing that's less readable to me, I thought it was going to be single line ^^

lib/metadata_gen.js
Tiriel: Yup, got that, but since you have to add a comment to make it explicit, I think there must be a more simple and more readable to do it.

I don't like having to rely on comments to understand what code does, I'd prefer you commented why you had to do this, what's the gain ^^

bin/metadata.js
cPhost: @Tiriel it is included in `MetadataGenerator.SCISSORS` in file `lib/metadata_gen.js` line 49-50.

Tiriel: @cPhost Yeah I know, I saw that after, but that's not very explicit or more readable. That's the only thing I'm saying. I fail to see why it's better that way, that's all ^^

apapirovski: That probably has more to do with the fact that the EOL used to be explicitly included in this file. I would expect something like SCISSORS and metadata to already include EOL.

apapirovski: Having EOL declared in this file means that when we want to use metadata or scissors in another target (such as `land-pr`) we need to remember that they don't include EOL and include it there. The more targets, the more times we have to duplicate that code and the more chances that an inadvertent bug sneaks in.

lib/metadata_gen.js
apapirovski: I just didn't think that `meta.join(EOL)${EOL}` is any clearer, I would probably need a comment too to make sure no one removes the final EOL.

Tiriel: That's a fair point :D
Thanks!

bin/metadata.js
apapirovski: Also, if you just think about the metadata content conceptually, it's not complete without the final EOL — that final EOL is something that is needed to make it complete.

Tiriel: I see your point!

Thanks for taking the time!

apapirovski:
I'm going to go ahead and merge since we have approvals from almost everyone. I assume we're ok with doing that but let me know if you would prefer to wait longer in the future.

PR #81, opened 2017-11-05
tools: wrap long lines

joyeecheung:


alopezsanchez:
Should we be strict with this? In my opinion, many times it makes the code less readable, especially in test files, applying BDD... 

Maybe we could add more options to ESLint rule, like patterns, regexp literals... (https://eslint.org/docs/rules/max-len)

What do you think?

joyeecheung:
@alopezsanchez I think we should go with what Node.js core uses...this is just an imitation tho, need to port that into something standalone later.

PR #85, opened 2017-11-06
commit package-lock.json

targos:
Any reason it's not already in git?


joyeecheung:
No particular reason...I don't usually use npm when I am behind the Chinese GFW so didn't notice this...

PR #86, opened 2017-11-06
bin: clean up argument support, rename binary file

joyeecheung:
This PR:

1. Remove the positional `[owner] [repo]` arguments, because that reminds me of the mess of `net.connect()`/`server.listen()` once again..we don't really have to support those, using the dash-prefixed arguments is pretty enough.
2. Rename the binary to `get_metadata.js`, which calls `metadata.js` that does the actual work, so we can reuse `metadata.js` later for other targets
3. Add an `-f` argument that specifies the file to write metadata in, this is also useful for future targets


lib/args.js
```diff
@@ -39,33 +45,29 @@ function buildYargs(args = null) {
...
-      id: +args.identifier
-    };
-  }
-  const match = args.identifier.match(PR_RE);
-  if (match !== null) {
-    if (typeof args.r !== 'undefined' || typeof args.o !== 'undefined') {
-      throw new Error(
-        'Cannot pass second or third argument ' +
-        'when url given as first argument.');
+  const { owner, repo, identifier, file } = args;
```
apapirovski: Can we specify the defaults here in destructuring (including an empty string for identifier) instead and then the line below would be `const result = { owner, repo, file };`?

README.md
```diff
@@ -47,14 +47,15 @@ This one is inspired by Evan Lucas's [node-review](https://github.com/evanlucas/
...
 Retrieves metadata for a PR and validates them against nodejs/node PR rules
 
 Options:
   --version    Show version number                                     [boolean]
-  -o, --owner  GitHub owner of the PR repository                        [string]
-  -r, --repo   GitHub repository of the PR                              [string]
+  --owner, -o  GitHub owner of the PR repository                        [string]
+  --repo, -r   GitHub repository of the PR                              [string]
+  --file, -f   File to write the metadata in                            [string]
   -h, --help   Show help                                               [boolean]
```
apapirovski: Can we switch the order of these to `--help, -h`?

lib/args.js
```diff
@@ -12,25 +12,31 @@ function buildYargs(args = null) {
...
       describe: 'GitHub repository of the PR',
       type: 'string'
     })
+    .option('file', {
+      alias: 'f',
+      demandOption: false,
+      describe: 'File to write the metadata in',
+      type: 'string'
+    })
     .help('h')
```
apapirovski: Can we mimic the above with alias being the short version? so it would become

```js
.help()
.alias('help', 'h')
```

joyeecheung:
@apapirovski I've moved `metadata.js` to steps (the intention is to put files that interact with external system but does not necessarily are binaries here). PTAL.

joyeecheung:
Just linking things: this is part of https://github.com/joyeecheung/node-core-utils/issues/10

PR #88, opened 2017-11-06
pr_checker: add warn about conflicts

alopezsanchez:
Question:
- Should we log something when `mergeable === 'UNKNOWN'`?

Please, review it, I don't know if I put all the things in the correct place.

demo

Fixes:  https://github.com/joyeecheung/node-core-utils/issues/68

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -16,6 +16,9 @@ const {
 const {
   FIRST_TIME_CONTRIBUTOR, FIRST_TIMER
 } = require('./user_status');
+const {
+  CONFLICTING
+} = require('./mergeable_state');
 
 const CIParser = require('./ci');
 const CI_TYPES = CIParser.TYPES;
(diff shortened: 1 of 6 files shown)


cPhost:
@alopezsanchez you can log something like
`
[WARN] GitHub is calculating Mergability
`

joyeecheung:
I wouldn't think it's necessary to warn about uncertain conflicts, people will always find out when they push, let's keep the volume to the minimum.

alopezsanchez:
Okay. Give me some hours to revert the last commit.

alopezsanchez:
Done.

lib/pr_checker.js
```diff
@@ -268,6 +272,20 @@ class PRChecker {
...
+
+  checkMergeableState() {
+    const {
+      pr, logger
+    } = this;
+
+    let status = true;
+    if (pr.mergeable && pr.mergeable === CONFLICTING) {
+      logger.warn('This PR has conflicts that must be resolved');
+      status = false;
```
apapirovski: Let's just simplify this to `return false;`

lib/pr_checker.js
```diff
@@ -268,6 +272,20 @@ class PRChecker {
...
+    }
+
+    return status;
```
apapirovski: This can then be `return true;` and we can remove the `let status = true;`

cPhost:
Merging it now, since i think everybody approved it.

PR #92, opened 2017-11-07
Skip wait check when label 'code-and-learn' is found, testing

Tiriel:
Hey team!

Quick addition, title is explicit enough I think.
Open to suggestions!

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -144,6 +144,13 @@ class PRChecker {
   checkPRWait(now) {
     const { pr } = this;
     const { logger } = this;
+    const labels = pr.labels.nodes.map((l) => l.name);
+    if (labels.includes('code-and-learn')) {
+      logger.info(
+        'This PR was is Code & Learn and doesn\'t have a time limit.'
+      );
+      return false;
+    }
     const wait = this.getWait(now);
     if (wait.timeLeft > 0) {
       const dateStr = new Date(pr.createdAt).toDateString();
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -147,7 +147,7 @@ class PRChecker {
     const labels = pr.labels.nodes.map((l) => l.name);
     if (labels.includes('code-and-learn')) {
       logger.info(
-        'This is a Code & Learn PR, it doesn\'t have a time limit.'
+        'This is a Code & Learn PR, no time limit.'
       );
       return false;
     }
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -144,13 +144,8 @@ class PRChecker {
   checkPRWait(now) {
     const { pr } = this;
     const { logger } = this;
-    const labels = pr.labels.nodes.map((l) => l.name);
-    if (labels.includes('code-and-learn')) {
-      logger.info(
-        'This is a Code & Learn PR, no time limit.'
-      );
-      return false;
-    }
+    const labels = pr.labels.nodes.some((l) => l.name === 'code-and-learn');
+    if (labels) { return true; }
     const wait = this.getWait(now);
     if (wait.timeLeft > 0) {
       const dateStr = new Date(pr.createdAt).toDateString();
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -147,7 +147,7 @@ class PRChecker {
     const labels = pr.labels.nodes.map((l) => l.name);
     if (labels.includes('code-and-learn')) {
       logger.info(
-        'This PR was is Code & Learn and doesn\'t have a time limit.'
+        'This is a Code & Learn PR, it doesn\'t have a time limit.'
       );
       return false;
     }
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -144,8 +144,10 @@ class PRChecker {
   checkPRWait(now) {
     const { pr } = this;
     const { logger } = this;
-    const labels = pr.labels.nodes.some((l) => l.name === 'code-and-learn');
-    if (labels) { return true; }
+    const labels = pr.labels.nodes;
+    const fast = labels.some((l) => l.name === 'code-and-learn') ||
+      (labels.length === 1 && labels[0].name === 'doc');
+    if (fast) { return true; }
     const wait = this.getWait(now);
     if (wait.timeLeft > 0) {
       const dateStr = new Date(pr.createdAt).toDateString();
--- a/test/unit/pr_checker.test.js
+++ b/test/unit/pr_checker.test.js
@@ -199,7 +199,7 @@ describe('PRChecker', () => {
 
       const expectedLogs = {
         warn: [],
-        info: [['This is a Code & Learn PR, it doesn\'t have a time limit.']],
+        info: [['This is a Code & Learn PR, no time limit.']],
         error: [],
         trace: []
       };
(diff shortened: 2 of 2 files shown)


test/unit/pr_checker.test.js
```diff
@@ -193,6 +193,42 @@ describe('PRChecker', () => {
       assert(!status);
       assert.deepStrictEqual(logger.logs, expectedLogs);
     });
+
+    it('should skip wait check for Code & Learn PR', () => {
+      const logger = new TestLogger();
+
+      const expectedLogs = {
+        warn: [],
+        info: [['This PR was is Code & Learn and doesn\'t have a time limit.']],
```
cPhost: Maybe just `This is a Code & Learn PR` to make it simple.

test/unit/pr_checker.test.js
```diff
@@ -193,6 +193,42 @@ describe('PRChecker', () => {
...
+        error: [],
+        trace: []
+      };
+
+      const now = new Date('2017-10-28T13:00:41.682Z');
```
cPhost: @Tiriel is there a need for this.
Nit: Just pass `new Date()` to the `checker.checkPRWait`

test/unit/pr_checker.test.js
cPhost: Maybe just This is a Code & Learn PR to make it simple.

Tiriel:
Refined ;)

test/unit/pr_checker.test.js
```diff
@@ -193,6 +193,42 @@ describe('PRChecker', () => {
...
+        labels: {
+          nodes: [
+            {
+              name: 'code-and-learn'
+            }
+          ]
+        }
+      });
+
+      const checker = new PRChecker(logger, {
```
cPhost: @Tiriel just a suggestion move this JSON to its own file `code_and_learn_pr.json` in `test/fixtures` like every other files.

test/unit/pr_checker.test.js
```diff
@@ -193,6 +193,42 @@ describe('PRChecker', () => {
       assert(!status);
       assert.deepStrictEqual(logger.logs, expectedLogs);
     });
+
+    it('should skip wait check for Code & Learn PR', () => {
+      const logger = new TestLogger();
+
+      const expectedLogs = {
+        warn: [],
+        info: [['This is a Code & Learn PR, it doesn\'t have a time limit.']],
```
cPhost: @Tiriel May be simpify this to `This is Code & Learn PR.`

Tiriel: Well, it wouldn't be very explicit as to what it's supposed to replace. In that case, we might just as well remove completely the phrase.
But since it's supposed to skip the time-before-landing limit, I thought it would be better to state clearly what we were skipping, or else silently skip it.

In terms of ux, I'd say more information is better than less :)

lib/pr_checker.js
```diff
@@ -144,6 +144,13 @@ class PRChecker {
   checkPRWait(now) {
     const { pr } = this;
     const { logger } = this;
+    const labels = pr.labels.nodes.map((l) => l.name);
+    if (labels.includes('code-and-learn')) {
+      logger.info(
+        'This is a Code & Learn PR, it doesn\'t have a time limit.'
+      );
+      return false;
```
apapirovski: `return true;` since there's no time limit which means it's valid to land whenever

lib/pr_checker.js
```diff
@@ -144,6 +144,13 @@ class PRChecker {
   checkPRWait(now) {
     const { pr } = this;
     const { logger } = this;
+    const labels = pr.labels.nodes.map((l) => l.name);
+    if (labels.includes('code-and-learn')) {
+      logger.info(
+        'This is a Code & Learn PR, it doesn\'t have a time limit.'
```
apapirovski: I don't think we need to log anything, to be honest, but maybe others disagree. We already log so much, the less noise, the better.

test/unit/pr_checker.test.js
Tiriel: Sold! I like your last ;)

lib/pr_checker.js
```diff
@@ -144,6 +144,13 @@ class PRChecker {
   checkPRWait(now) {
     const { pr } = this;
     const { logger } = this;
+    const labels = pr.labels.nodes.map((l) => l.name);
+    if (labels.includes('code-and-learn')) {
```
apapirovski: This could be simplified to

```js
if (pr.labels.nodes.some((l) => l.name === 'code-and-learn'))
```

test/unit/pr_checker.test.js
Tiriel: I was talking about the other tests cases of this file, in fact, sorry, porr english ^^

apapirovski:
(Please see my 'outdated' reviews too, bad timing on my part but the feedback is still valid.)

lib/pr_checker.js
cPhost: @Tiriel i we should not log anything because we do not log anything about time when it is ready to land and our goal is to have as less noise as possible.

addaleax:
By the way, the missing time limit isn’t because a PR is made for Code & Learn, it’s because it’s touching only a specific part of the tests or documentation – maybe we can work towards checking whether a PR does that, as a future goal :)

cPhost:
@addaleax meaning if it has `test` or `doc` label.

Tiriel:
@addaleax is there specific labels to look for then?

@cPhost @apapirovski thanks for the input, gonna implement them asap!

I'm going to end up with the most commented and reviewed PRs each time it seems... Dunno if it's good or bad, but it feels like I have lots to learn yet

cPhost:
> I'm going to end up with the most commented and reviewed PRs each time it seems... Dunno if it's good or bad, but it feels like I have lots to learn yet.

The more comments it has the better the PR Gets. 👍

addaleax:
I think `test` and `doc` are good labels to check for, but checking whether the commit only affects one or two test/doc files would be ideal (also much harder to implement I guess, so I wouldn’t suggest it for this PR?)

Tiriel:
Thanks @addaleax !
I feel it would be weird to ship the new labels condition without the number of files, so I'll just open a new issue so we can agree on the max number of files to check.

joyeecheung:
@addaleax I would be careful about test-only because some tests are trickier than the others (I am looking at `async-wrap-uncaughtexception`, sign), but doc-only changes seem to be safe enough, I think "having only the doc label" is a pretty good indication?

joyeecheung:
@addaleax Also we can just add a new label like "fast-track"...I think I've put it in the source code as a todo somewhere, lol

Tiriel:
I added the parsing for 'doc' label only, just in case. Otherwise, I just wait for your reviews ;)

addaleax:
@joyeecheung I’m fine with anything :) I don’t think we can remove human judgement from these cases anyway

Tiriel:
@apapirovski @cPhost If you guys are ok to review again... ^^

Tiriel:
@nodejs/automation I'd like at least one more review before merging, just to be sure. Otherwise I'm going to merge it anyway in few hours since there seems to be no objection.

Tiriel:
@joyeecheung No problem at all! Sorry, I'm a bit bored at work these days, so I got too much time for this :D

Merging!

PR #93, opened 2017-11-07
PRChecker: Update `commitsAfterReview` to warn only last 3 commits

cPhost:
To keep noise level to minimum, only warn about last 3 commits since the last review.

Same exact work for `checkCI` is coming in next pr.
idea: we can also have flag for how much commits to log.

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -209,7 +209,7 @@ class PRChecker {
             cli.warn(`Commits pushed after the last ${lastCI.typeName} ` +
             'CI run:');
           }
-          cli.warn(`- ${commit.message.split('\n')[0]}`);
+          cli.warn(`- ${commit.messageHeadline}`);
         }
       });
     }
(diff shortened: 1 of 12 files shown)


lib/pr_checker.js
```diff
@@ -266,30 +266,41 @@ class PRChecker {
...
+        afterCommits.push(commit);
+      }
+    });
+
+    if (afterCommits.length > 0) {
+      let warnMsg =
+        `${afterCommits.length} commits were pushed since the last review:`;
+      logger.warn(warnMsg.replace('1 commits were', 'Single commit was'));
+
+      if (afterCommits.length < 4) {
```
joyeecheung: In that case maybe make this number a variable first? (`maxCommits + 1` here)

test/unit/pr_checker.test.js
```diff
@@ -380,9 +381,38 @@ describe('PRChecker', () => {
...
+      assert.deepStrictEqual(status, false);
+      assert.deepStrictEqual(logger.logs, expectedLogs);
+    });
+
+    it('should only warn last three commits if more than 3 commits', () => {
+      const { commits, reviews } = moreThanThreeCommitsAfterReview;
+      const expectedLogs = {
+        warn: [
+          [ '4 commits were pushed since the last review:' ],
+          [ 'Last 3 commits:' ],
```
joyeecheung: I think we can replace this line with a `...` line at the end so we don't have double colon-ended sentences.

cPhost:
@joyeecheung what do you think about flags to log the number of commits?

cPhost:
@joyeecheung i update the current version to use `.splice` to avoid unnecessary `if`, `else`  and be more clear.

joyeecheung:
@cPhost Yes a flag sounds like a good idea, maybe `--warn-commits=n`? (or any thing that you think would be better)

lib/pr_checker.js
```diff
@@ -269,30 +269,39 @@ class PRChecker {
...
+      let warnMsg =
+        `${totalCommits} commits were pushed since the last review:`;
+
+      logger.warn(warnMsg.replace('1 commits were', 'Single commit was'));
+      if (totalCommits > 3) {
+        logger.warn('last 3 commits...');
+      }
+
+      afterCommits.splice(0, totalCommits - 3);
+      afterCommits.forEach(commit => {
```
joyeecheung: Or just `afterCommits.slice(totalCommits - 3).forEach(...)`?

test/unit/pr_checker.test.js
```diff
@@ -381,9 +382,38 @@ describe('PRChecker', () => {
...
+      assert.deepStrictEqual(status, false);
+      assert.deepStrictEqual(logger.logs, expectedLogs);
+    });
+
+    it('should only warn last three commits if more than 3 commits', () => {
+      const { commits, reviews } = moreThanThreeCommitsAfterReview;
+      const expectedLogs = {
+        warn: [
+          [ '4 commits were pushed since the last review:' ],
+          [ 'last 3 commits...' ],
```
joyeecheung: Maybe merge this line with the one above? `4 commits were pushed since the last review, the last 3 commits are:`

lib/pr_checker.js
cPhost: @joyeecheung this does not work because it would return the removed items rather than what we need the last three commits.

joyeecheung: `splice` would return the removed items while `slice` will return the sliced items...no?

joyeecheung: @cPhost `slice`ing an array with length < 3 would result in an empty array, if you are `forEach`ing the array I think it's gonna be fine.

cPhost: @joyeecheung the test are failing. I will give it a try again.

joyeecheung: @cPhost I think I know why it's failing: instead of `afterCommits.slice(totalCommits - 3)`, do `afterCommits.slice(-3)`, so you don't end up with `-2` or `-1`

cPhost:
@joyeecheung done. why are the CIs disabled?

joyeecheung:
@cPhost I think it's because I've moved it into the foundation...

cPhost:
Did not notice that at all!

cPhost:
Going to merge this now!

package-lock.json
```diff
@@ -1347,7 +1347,7 @@
     },
     "figures": {
       "version": "2.0.0",
-      "resolved": "http://registry.npm.taobao.org/figures/download/figures-2.0.0.tgz",
```
joyeecheung: This  reminds me that I should switch my registry when creating the lock file...

cPhost: I think it got updated because i used `npm prune`.

PR #104, opened 2017-11-10
PRChecker: support new label fast-track

cPhost:
Now core has the label `fast-track` and a PR that uses that label so i think it is time to add support for new label that @joyeecheung  opened a issue for.

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -53,8 +53,8 @@ class PRChecker {
     const status = [
       this.checkReviews(comments),
       this.checkCommitsAfterReview(),
-      this.checkPRWait(new Date()),
       this.checkCI(),
+      this.checkPRWait(new Date()),
       this.checkMergeableState()
     ];
 
@@ -150,16 +150,29 @@ class PRChecker {
    * @param {Date} now
    */
   checkPRWait(now) {
-    const { pr } = this;
-    const { cli } = this;
+    const {
+      pr, cli, reviewers, CIStatus
+    } = this;
     const labels = pr.labels.nodes;
 
     const fast =
-      labels.some((l) => ['code-and-learn', 'fast-track'].includes(l.name)) ||
-      (labels.length === 1 && labels[0].name === 'doc');
+      labels.some((l) => ['fast-track'].includes(l.name));
     if (fast) {
-      cli.info('This PR is being fast-tracked.');
-      return true;
+      const { approved } = reviewers;
+      if (approved.length > 1 && CIStatus) {
+        cli.info('This PR is being fast-tracked');
+        return true;
+      } else {
+        const msg = ['This PR is being fast-tracked, but awating '];
+        if (approved.length < 2) msg.push('approvals of 2 contributors');
+        if (!CIStatus) msg.push('a CI run');
+
+        let warnMsg = msg.length === 2
+          ? msg.join('') : `${msg[0] + msg[1]} and ${msg[2]}`;
+        cli.warn(warnMsg);
+      }
+
+      return false;
     }
 
     const wait = this.getWait(now);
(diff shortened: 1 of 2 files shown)


refack:
Optional: still emit a warning that it's a fast-tracked PR and the lander should double check, and use discretion.

cPhost:
@refack i think that was a good suggestion to add warn msg for `fast-track` commits. done

joyeecheung:
Hmmm, do we need to warn about that tho? Maybe info?

joyeecheung:
(This reminds me I forgot to open the PR to update collaborator's guide!....)

lib/pr_checker.js
```diff
@@ -152,9 +152,17 @@ class PRChecker {
...
     const { cli } = this;
     const labels = pr.labels.nodes;
-    const fast = labels.some((l) => l.name === 'code-and-learn') ||
+
+    let isFastTrack = false;
+    const fast = labels.some((l) => isFast(l.name)) ||
       (labels.length === 1 && labels[0].name === 'doc');
+
+    if (isFastTrack) {
+      cli.warn('No wait time! "fast-track" labelled PR.');
```
joyeecheung: Maybe just: `This PR is being fast-tracked`, they can always look up the labels in the `logIntro()` messages. Also I prefer an info over warning.

lib/pr_checker.js
```diff
@@ -164,6 +172,11 @@ class PRChecker {
       return false;
     }
 
+    function isFast(label) {
+      isFastTrack = label === 'fast-track';
```
joyeecheung: Why do we need a closure here? All the checks can be done in the function passed to `labels.some()`

cPhost: @joyeecheung To check afterwards if the label was fast-track and log the message out.

cPhost:
@joyeecheung fixed it.

cPhost:
Adding `do-not-land` label until the PR is ready to get merged.

lib/pr_checker.js
joyeecheung: @cPhost But now that we don't log that in the messages...we don't really need this right? All we need is

```js
const fast = labels.some((l) => (['code-and-learn', 'fast-track'].includes(l)) ||
  (labels.length === 1 && labels[0].name === 'doc');
if (fast) {
  cli.info(...)
  return true;
}
```

cPhost: Oh i though we do not count `code-and-learn` fixing it right now.

cPhost: done.

joyeecheung:
BTW: https://github.com/nodejs/node/pull/17056 has landed, the current rule is:

> When a pull request is deemed suitable to be fast-tracked, label it with fast-track. The pull request can be landed once 2 or more collaborators approve both the pull request and the fast-tracking request, and the necessary CI testing is done.

cPhost:
@joyeecheung just want to know if the `code-and-learn` and/or `doc` PRs counts as fast-tracked because we currently count them as being fast-tracked.

BTW this is what i currently have: 
``` bash
# if it can be fast-tracked, have two approvals a CI cli.info
This PR is being fast-tracked
# else if missing approvals or ci or both cli.warn
This PR is being fast-tracked, but awaiting approvals of 2 contributors # or
This PR is being fast-tracked, but awaiting a CI run # or
This PR is being fast-tracked, but awaiting approvals of 2 contributors and a CI run
````

and so it will warn if there is CI for docs (IIRC there is no CI testing for docs or maybe its linter)

joyeecheung:
@cPhost I would suggest to skip the code-and-learn and doc rules and just check for fast-track. If it should be fast-tracked, people should label it.

cPhost:
@joyeecheung done.

PR #105, opened 2017-11-10
PRChecker: add flag `--warn-commits`.

cPhost:
Added flag `warn-commits` that defaults to 3.  Documented the flag in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Options:
   --repo, -r        GitHub repository of the PR                         [string]
   --file, -f        File to write the metadata in                       [string]
   --check-comments  Check for 'LGTM' in comments                       [boolean]
-  --warn-commits    Number of commits to warn              [number] [default: 3]
+  --max-commits     Number of commits to warn              [number] [default: 3]
   --help, -h        Show help                                          [boolean]
 ```
 
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ Options:
   --owner, -o       GitHub owner of the PR repository                   [string]
   --repo, -r        GitHub repository of the PR                         [string]
   --file, -f        File to write the metadata in                       [string]
-  --check-comments  Check for 'LGTM' in comments                        [boolean]
+  --check-comments  Check for 'LGTM' in comments                       [boolean]
+  --warn-commits    Number of commits to warn              [number] [default: 3]
   --help, -h        Show help                                          [boolean]
 ```
 
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -29,10 +29,10 @@ class PRChecker {
    * @param {{}} cli
    * @param {PRData} data
    */
-  constructor(cli, data) {
+  constructor(cli, data, argv) {
     this.cli = cli;
     const {
-      pr, reviewers, comments, reviews, commits, collaborators, warnCommits
+      pr, reviewers, comments, reviews, commits, collaborators
     } = data;
     this.reviewers = reviewers;
     this.pr = pr;
@@ -43,7 +43,7 @@ class PRChecker {
     // the last review.
     this.reviews = reviews;
     this.commits = commits;
-    this.warnCommits = warnCommits;
+    this.argv = argv;
     this.collaboratorEmails = new Set(
       Array.from(collaborators).map((c) => c[1].email)
     );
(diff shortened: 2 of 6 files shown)


steps/metadata.js
```diff
@@ -35,6 +35,7 @@ module.exports = async function getMetadata(argv, cli) {
   cli.write(metadata);
   cli.separator();
 
+  data.warnCommits = argv.warnCommits;
   const checker = new PRChecker(cli, data);
```
joyeecheung: A better way to do this is just passing `argv` to the `PRChecker` constructor instead of passing them/monkey-patching them here and there...

cPhost: @joyeecheung done and added a test since, now can pass a flag.

test/unit/pr_checker.test.js
```diff
@@ -474,11 +489,40 @@ describe('PRChecker', () => {
...
         commits: simpleCommits,
         collaborators
-      });
+      }, argv);
 
       const status = checker.checkCommitsAfterReview();
       assert.deepStrictEqual(status, true);
     });
+
+    it('should print out three commits if passed NaN', () => {
```
joyeecheung: Shouldn't we test cases like `warnCommits: 0`?

joyeecheung: Also I am kinda curious if `warnCommits: -1`...should be `slice(1)`, which kinda make sense

README.md
```diff
@@ -56,7 +56,8 @@ Options:
   --owner, -o       GitHub owner of the PR repository                   [string]
   --repo, -r        GitHub repository of the PR                         [string]
   --file, -f        File to write the metadata in                       [string]
-  --check-comments  Check for 'LGTM' in comments                        [boolean]
+  --check-comments  Check for 'LGTM' in comments                       [boolean]
+  --warn-commits    Number of commits to warn              [number] [default: 3]
```
joyeecheung: On a second thought, maybe `--max-commits` otherwise this sounds like a boolean to turn on/off commit warning.

test/unit/pr_checker.test.js
cPhost: `maxCommits: -1` Would warn first commit after review and so on. `0` would warn all commits! which it should not. Here is what i am thinking...

```javascript
maxCommits = -maxCommits;
// In this case if negative number or zero was passed
// it should not warn out commits. mainly because it seems pretty 
// buggy to log first commits in case of negatives!
maxCommits = maxCommits > -1 ? commits.length : maxCommits;  
```

README.md
cPhost: Using `max-commits` flag now.

test/unit/pr_checker.test.js
cPhost: @joyeecheung updated as follows. if `0` is passed it should not log anything. if `-1` or any negative number convert it to positive, and if `NaN` in case of string is passed it shoule default to `3`.

cPhost:
CI looks Green:
https://travis-ci.org/nodejs/node-core-utils/builds/301146913
https://ci.appveyor.com/project/joyeecheung/node-core-utils/build/168.$%7Bpatch%7D

test/unit/pr_checker.test.js
```diff
@@ -510,11 +556,63 @@ describe('PRChecker', () => {
...
 
       const status = checker.checkCommitsAfterReview();
       assert.deepStrictEqual(status, true);
     });
+
+    it('should log as expected if passed 1 as flag', () => {
+      const { commits, reviews } = moreThanThreeCommitsAfterReview;
+      const expectedLogs = {
+        warn: [
+          [ '4 commits were pushed since the last review, ' +
```
joyeecheung: Can we just make it

```
Commits pushed since the last review:
- foo: bar
- quo: qux
- bar: foo
...(use `--max-commits ${totalLength}` to see the full list of commits)
```

cPhost: Sure, i think its a good idea.

cPhost: @joyeecheung changed it.

lib/pr_checker.js
```diff
@@ -210,19 +212,22 @@ class PRChecker {
...
-        afterCommits.slice(-3)
+        let sliceLength = maxCommits === 0 ? totalCommits : -maxCommits;
+        afterCommits.slice(sliceLength)
           .forEach(commit => {
             cli.warn(`- ${commit.messageHeadline}`);
           });
+
+        if(totalCommits > maxCommits) {
+          let infoMsg = '...(use `' +
+            `--warn-commits ${totalCommits}` +
```
joyeecheung: This should be `--max-commits`..

cPhost:
Appveyor: https://ci.appveyor.com/project/joyeecheung/node-core-utils/build/201.master

PR #110, opened 2017-11-14
lib: add request-v3, for GitHub API v3 support

cPhost:
We can use this whenever we need it. Let me know if you think there should a diffrent name for this.

### DIFF
--- a/lib/request-v3.js
+++ b/lib/request-v3.js
@@ -1,0 +1,43 @@
+'use strict';
+
+const rp = require('request-promise-native');
+
+// create a valid user agent required by GitHub API v3,
+const userAgent =
+  'node-core-utils/v1 (https://github.com/nodejs/node-core-utils)';
+
+class Request {
+  constructor(credentials) {
+    // GitHub API v3 does not require authorization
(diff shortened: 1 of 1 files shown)


targos:
I think we should use a library like https://github.com/octokit/node-github (it supports promises) for API v3 calls.

cPhost:
@targos that would be good idea if this whole repo was on v3 but it based on v4 it just required for one functionality or maybe some in feature.

apapirovski:
I think... I tend to agree with @targos here. Using `node-github` means we don't need to worry about testing some of that stuff and it allows us to easily expand functionality in the future, as opposed to refactoring everything when we suddenly need it at a later time.

Tiriel:
I do agree too. For now we need only for one thing, but I think we're doomed to have more uses for it later unless GitHub implements new nodes in the Graph API.

I think adding the `node-github` package would be a safe bet now.

joyeecheung:
A few thoughts:

1. We can just put a v3 method/property on the `Request` class in `lib/request.js`.
2. I tend to agree with @targos , we can instantiate the node-github client in `Request` and expose it as `request.v3`

PR #126, opened 2017-11-25
checkCI: log out CI status of a PR.

cPhost:


lib/pr_checker.js
```diff
@@ -249,10 +252,30 @@ class PRChecker {
...
+        switch (status.state) {
+          case EXPECTED:
+          case PENDING:
+            cli.warn('CI test results are still pending');
+            break;
+          case SUCCESS:
+            cli.info('All CI tests are successful');
+            break;
+          case FAILURE:
+            cli.warn('Some or all CI tests seems to be failing or flasky');
```
joyeecheung: Nit: `s/seems/seem/`, `s/flasky/flaky/`

lib/pr_checker.js
```diff
@@ -249,10 +252,30 @@ class PRChecker {
...
+            break;
+          case ERROR:
+            cli.error('CI testing seems to be error out');
```
joyeecheung: I would prefer a description like `CI tests have failed.`

lib/pr_checker.js
cPhost: After thinking about it like a brief message for case `FAILURE` because the goal was if the ci is all green they do not have to go and check the CI but when it is `FAILURE` it either because of flasky or unrelated error so i think this should work:
`CI tests have some failures, perhaps maybe unrelated or flasky error.`

lib/pr_checker.js
cPhost: done.

cPhost:
Closing this since the CI system does not seem to reliable yet...

PR #134, opened 2018-01-03
Show actual value in error message

tniessen:
The old code just displayed `[object Object]` instead of the actual argument.


joyeecheung:
Thanks for catching that!

Issue #137, opened 2018-01-04
Detect merged/closed PR


gibfahn:
Currently if you run get-metadata on a closed PR, you don't get an obvious indication that it's closed.

```bash
▶▶▶ get-metadata 17899                                                                                                                                                                                                                ~/wrk/com/node-core-utils (xdg-config✦)
✔  Done loading data for nodejs/node/pull/17899
----------------------------------- PR info ------------------------------------
Title   doc: remove x86 from os.arch() options #17899
Author  gibfahn
Commits 1
Branch  gibfahn:process-arch -> nodejs:master
Labels  doc
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/17899
Reviewed-By: Ben Noordhuis <info@bnoordhuis.nl>
Reviewed-By: Luigi Pinca <luigipinca@gmail.com>
Reviewed-By: Anatoli Papirovski <apapirovski@mac.com>
Reviewed-By: James M Snell <jasnell@gmail.com>
Reviewed-By: Colin Ihrig <cjihrig@gmail.com>
Reviewed-By: Tiancheng "Timothy" Gu <timothygu99@gmail.com>
Reviewed-By: Daniel Bevenius <daniel.bevenius@gmail.com>
--------------------------------------------------------------------------------
✔  Requested Changes: 0
✔  Approvals: 7, 3 from TSC (jasnell, cjihrig, danbev)
⚠  Commits were pushed since the last review:
⚠  - doc: remove x86 from os.arch() options
ℹ  Last Full CI on 2018-01-03T22:59:42Z: https://ci.nodejs.org/job/node-test-commit/15215/
⚠  Commits were pushed after the last Full CI run:
⚠  - doc: remove x86 from os.arch() options
```

It should probably be made clear that the PR is closed.

tniessen:
Would a warning suffice or should this be an error?

gibfahn:
>Would a warning suffice or should this be an error?

I'm not sure, I was thinking that a non-zero exit code would be useful if you were scripting this, but I haven't actually scripted it so I'm not sure...

I was also thinking that maybe it shouldn't generate metadata, but again, there may be use-cases where you actually want to get metadata for closed PRs, so IDK about that either.

Basically IDK, that's why I opened this as an issue, to see what people think 😁 .

Maybe just a big 🔴 would be adequate.

joyeecheung:
[...]

PR #138, opened 2018-01-04
Warn if PR if closed or merged

cPhost:
Fixes: https://github.com/nodejs/node-core-utils/issues/137

### DIFF
--- a/test/unit/pr_checker.test.js
+++ b/test/unit/pr_checker.test.js
@@ -25,7 +25,9 @@ const {
   firstTimerPR,
   firstTimerPrivatePR,
   semverMajorPR,
-  conflictingPR
+  conflictingPR,
+  closedPR,
+  mergedPR
 } = require('../fixtures/data');
 
 const argv = { maxCommits: 3 };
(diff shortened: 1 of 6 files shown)


test/unit/pr_checker.test.js
```diff
@@ -800,4 +806,74 @@ describe('PRChecker', () => {
...
+      const checker = new PRChecker(cli, options, argv);
+      const status = checker.checkPRState();
+      assert.strictEqual(status, false);
+      cli.assertCalledWith(expectedLogs);
+    });
+
+    it('should warn if PR is merged', () => {
+      const expectedLogs = {
+        warn: [
+          [ 'This PR is merged' ]
```
tniessen: We use `[abc]` instead of `[ abc ]` (no spaces around brackets) in node.js core. Personally, I am not a fan of that convention, just note that this differs.

cPhost: Ah.., we do not have any lint rules for this and so we do have mixed up array with and without spaces.

joyeecheung: I think the current rule is: as long as you are being consistent, i.e. not something like `[ 'something']`, you are fine.

Issue #144, opened 2018-01-18
from the doc…


srl295:
```
$ ncu-config upstream your-remote-name
$ ncu-config branch your-branch-name 
```

* should this be `ncu-config` **set** `upstream your-remote-name` ?
* also, `Error: ENOENT: no such file or directory, open '/Users/srl/src/node/.ncu/config'`

PR #147, opened 2018-01-18
Fix ncu-config

joyeecheung:
Fixes: #144


joyeecheung:
Anyone wants to review it?

Issue #156, opened 2018-01-23
get-metadata: missing "Refs:" in metadata section


vsemozhetbyt:
For https://github.com/nodejs/node/pull/18310, "Refs:" was omitted in metadata section.

cPhost:
@vsemozhetbyt can provide more details on this or do you have any clue why that happend?

vsemozhetbyt:
@cPhost I cannot understand the cause.
Commit message:
```txt
doc: remove confusing signature in fs.md

Fixes: https://github.com/nodejs/node/issues/18305
Refs: https://github.com/nodejs/node/pull/13424
```
<details>
<summary>get-metadata output:</summary>

```console
$ winpty get-metadata.cmd 18310
√  Done loading data for nodejs/node/pull/18310
----------------------------------- PR info ------------------------------------
Title      doc: remove confusing signature in fs.md (#18310)
Author     Vse Mozhet Byt <vsemozhetbyt@gmail.com> (@vsemozhetbyt)
Branch     vsemozhetbyt:doc-fs-del-err-signature -> nodejs:master
Labels     doc, fast-track, fs
Commits    1
 - doc: remove confusing signature in fs.md
Committers 1
 - Vse Mozhet Byt <vsemozhetbyt@gmail.com>
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/18310
Fixes: https://github.com/nodejs/node/issues/18305
Reviewed-By: Joyee Cheung <joyeec9h3@gmail.com>
Reviewed-By: Gireesh Punathil <gpunathi@in.ibm.com>
Reviewed-By: Richard Lau <riclau@uk.ibm.com>
--------------------------------------------------------------------------------
√  Requested Changes: 0
√  Approvals: 3, 1 from TSC (joyeecheung)
i  Last Full CI on 2018-01-23T03:47:53Z: https://ci.nodejs.org/job/node-test-pull-request-lite/115/
i  This PR is being fast-tracked
‼  This PR is closed
```
</details><br>
Seems to be a common case, [`REFS_RE`](https://github.com/nodejs/node-core-utils/blob/8e3d831dc5deee9b6e053f5a9f26b30b410cfaa0/lib/links.js#L5) is matched.

cPhost:
Aren't you suppose to add `Refs: <url>` in for `get-metadata` it to detect it?

gibfahn:
[...]

PR #159, opened 2018-01-27
Generate metadata from commits too.

priyank-p:
This also does not add duplicate PR in LinkParser.
Fixes: https://github.com/nodejs/node-core-utils/issues/156

### DIFF
--- a/lib/metadata_gen.js
+++ b/lib/metadata_gen.js
@@ -10,11 +10,16 @@ class MetadataGenerator {
    * @param {PRData} data
    */
   constructor(data) {
-    const { owner, repo, pr, reviewers } = data;
+    const { owner, repo, pr, reviewers, commits } = data;
     this.owner = owner;
     this.repo = repo;
     this.pr = pr;
     this.reviewers = reviewers;
+    this.op = pr.bodyHTML;
+
+    commits.forEach(commit => {
+      this.op += commit.commit.messageBodyHTML;
+    });
   }
 
   /**
@@ -23,9 +28,10 @@ class MetadataGenerator {
   getMetadata() {
     const {
       reviewers: { approved: reviewedBy },
-      pr: { url: prUrl, bodyHTML: op },
+      pr: { url: prUrl },
       owner,
-      repo
+      repo,
+      op
     } = this;
 
     const parser = new LinkParser(owner, repo, op);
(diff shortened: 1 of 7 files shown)


joyeecheung:
This could result in duplicate entries in `git-node`...`LandingSession#amend()` needs to remove those from the original commit message during merging.

lib/metadata_gen.js
```diff
@@ -10,11 +10,16 @@ class MetadataGenerator {
...
-    const { owner, repo, pr, reviewers } = data;
+    const { owner, repo, pr, reviewers, commits } = data;
     this.owner = owner;
     this.repo = repo;
     this.pr = pr;
     this.reviewers = reviewers;
+    this.op = pr.bodyHTML;
+
+    commits.forEach(commit => {
+      this.op += commit.commit.messageBodyHTML;
```
joyeecheung: If there are multiple commits in one PR, a refs in commit A does not necessarily means it is referenced by commit B...need to figure out how to make this work

cPhost: So here i guess you mean some times a PR's commits needs to land in not just one but multiple commits (mostly 2-3 commits) so we want to output them differently, per commit.

joyeecheung: @cPhost Yes. Thinking about that, the best we could do is printing these differently for different commits, but don't write them into the actual output consumed by git-node

cPhost: > but don't write them into the actual output consumed by git-node

This a part i am bit confused on where exactly to print them (sure not it generated metadata) right.

joyeecheung: @cPhost we could do some refactoring with the link parser and get it to return those links separately, say one for the OP, and a map for each commit.

cPhost: @joyeecheung if you want to takeover this PR go ahead. I think you will be the right person to solve this. I cannot work on this anymore until after March 10 i got test to prepare for :)

joyeecheung: @cPhost Sure, I also plan to write some tests for git-node first so patches can be properly tested.

Good luck on your tests!

cPhost: Thanks.

codebytere:
@priyank-p do you ever plan to come back to this or can it be closed?

priyank-p:
I will close this out! I wished I wrote better code back ~2 years ago that I could quickly understand sigh.

PR #162, opened 2018-02-06
doc: split doc for different tools

joyeecheung:
Split the docs for different tools to make the README more concise.


joyeecheung:
Anyone wants to review it? Thanks! @nodejs/automation

targos:
>  but should we rename doc directory to docs?

+1. This would allow us to enable github pages for the docs directory

joyeecheung:
@cPhost Node core is using `doc`, so I figured why not..

joyeecheung:
Updated and merged. Thanks!

cPhost:
> @cPhost Node core is using doc, so I figured why not..

Oh, i never noticed that. Normally all project stick to docs since all docs are in that place, so it makes sense to name it docs rather than doc.

targos:
I was wrong about "docs" for gh pages. It only works if there is an index page.
Instead I enabled it on master: https://nodejs.github.io/node-core-utils/

cPhost:
@targos i think if you would have enabled it as `master branch /docs folder` and visited the name of the `https://nodejs.github.io/node-core-utils/get-metadata.html` it should work.

targos:
@cPhost Yeah, I initially tried that, but then there is no index page and since the README is automatically used as an index page, I think it's simpler to just enable `master branch`.

cPhost:
Yeah, thats true

PR #164, opened 2018-02-06
ncu-team: new tool to sync members of a Github team

joyeecheung:


lib/team_info.js
```diff
@@ -0,0 +1,98 @@
...
+      const info = new TeamInfo(cli, request, org, team);
+      const teamData = await info.getMemberContacts();
+      const opening = `<!-- ncu-team-sync.team(${org}/${team}) -->`;
+      const ending = `<!-- ncu-team-sync end -->`;
+      blocks.set(mapKey, `${opening}\n\n${teamData}\n\n${ending}`);
+    };
+  }
+
+  if (blocks.size === 0) {
+    throw new Error(`Could not find block matching ${RE}`);
```
cPhost: Can we instead log out the comments instead here i guess the regular expression would be kind of daunting to log out to user?

lib/queries/Team.gql
```diff
@@ -0,0 +1,17 @@
+query team($org: String!, $team: String!) {
+  organization(login: $org) {
+    login
+    team(slug: $team) {
+      name,
+      description,
+      members(first: 100) {
```
cPhost: Does this limit how many team members we receive?

joyeecheung: @cPhost Yeah, I plan to fix that in a later PR

cPhost: Cool! then :)

PR #168, opened 2018-02-09
Add `bin` folder to linting

alopezsanchez:
Hello all!
Sorry for being inactive the last few months.

`bin` folder was not being included with the `eslint .` command (I think because files aren't `.js`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,11 +13,11 @@
     "test": "npm run test-unit && npm run lint",
     "test-unit": "mocha --require intelli-espower-loader test/unit/*.test.js --exit",
     "test-all": "mocha --require intelli-espower-loader test/**/*.test.js --exit",
-    "lint-fix": "eslint . --fix",
     "coverage": "nyc --reporter=html --reporter=text --reporter=text-summary npm test",
     "coverage-all": "nyc --reporter=lcov --reporter=text --reporter=text-summary npm run test-all",
-    "lint": "eslint . --cache",
-    "report-coverage": "nyc report --reporter=lcov > coverage.lcov && codecov"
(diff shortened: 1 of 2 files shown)


cPhost:
@alopezsanchez thanks for catching this, and welcome back :)

package.json
```diff
@@ -13,11 +13,11 @@
...
     "test-unit": "mocha --require intelli-espower-loader test/unit/*.test.js --exit",
     "test-all": "mocha --require intelli-espower-loader test/**/*.test.js --exit",
-    "lint-fix": "eslint . --fix",
     "coverage": "nyc --reporter=html --reporter=text --reporter=text-summary npm test",
     "coverage-all": "nyc --reporter=lcov --reporter=text --reporter=text-summary npm run test-all",
-    "lint": "eslint . --cache",
-    "report-coverage": "nyc report --reporter=lcov > coverage.lcov && codecov"
+    "report-coverage": "nyc report --reporter=lcov > coverage.lcov && codecov",
+    "lint": "eslint . 'bin/*' --cache",
+    "lint-fix": "npm run lint -- --fix"
```
cPhost: Can we avoid this double piping of npm script here. It would output npm stuff two time which does look right?

alopezsanchez: Yes, we can remove it if you consider that this is not good.

However, I changed it to avoid repeating code.

cPhost: I'd do prefer no having the piped script, so if you can change this it would be great.

joyeecheung: I prefer to write the eslint command out as well...that'll be clearer IMO. Thanks

alopezsanchez: Done!

PR #171, opened 2018-02-10
CI Fix node-commit-linter-lite not being detected 

cPhost:
I still need to add test for this.

### DIFF
--- a/lib/ci.js
+++ b/lib/ci.js
@@ -12,17 +12,19 @@ const LIBUV = 'LIBUV';
 const NOINTL = 'NOINTL';
 const V8 = 'V8';
 const LINTER = 'LINTER';
+const LITE = 'LITE';
 
 const CI_TYPES = new Map([
   [CITGM, { name: 'CITGM', re: /citgm/ }],
   [FULL,
     { name: 'Full',
-      re: /node-test-pull-request|node-test-commit\// }],
+      re: /node-test-pull-request\/|node-test-commit\// }],
   [BENCHMARK, { name: 'Benchmark', re: /benchmark/ }],
   [LIBUV, { name: 'libuv', re: /libuv/ }],
   [NOINTL, { name: 'No Intl', re: /nointl/ }],
   [V8, { name: 'V8', re: /node-test-commit-v8/ }],
-  [LINTER, { name: 'Linter', re: /node-test-linter/ }]
+  [LINTER, { name: 'Linter', re: /node-test-linter/ }],
+  [LITE, { name: 'Lite', re: /node-test-.+-lite/ }]
 ]);
 
 class CIParser {
@@ -78,7 +80,7 @@ class CIParser {
 
 CIParser.TYPES = CI_TYPES;
 CIParser.constants = {
-  CITGM, FULL, BENCHMARK, LIBUV, V8, NOINTL, LINTER
+  CITGM, FULL, BENCHMARK, LIBUV, V8, NOINTL, LINTER, LITE
 };
 
 module.exports = CIParser;
--- a/test/fixtures/comments_with_ci.json
+++ b/test/fixtures/comments_with_ci.json
@@ -22,5 +22,8 @@
   {
     "publishedAt": "2017-10-22T04:16:36.458Z",
     "bodyText": "Linter CI: https://ci.nodejs.org/job/node-test-linter/13127/"
+  },{
+    "publishedAt": "2018-02-09T21:38:30Z",
+    "bodyText": "CI https://ci.nodejs.org/job/node-test-commit-lite/246/"
   }
 ]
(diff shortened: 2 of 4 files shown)


devsnek:
this already seems to work with the latest npm release

lib/ci.js
```diff
@@ -22,7 +23,8 @@ const CI_TYPES = new Map([
   [LIBUV, { name: 'libuv', re: /libuv/ }],
   [NOINTL, { name: 'No Intl', re: /nointl/ }],
   [V8, { name: 'V8', re: /node-test-commit-v8/ }],
-  [LINTER, { name: 'Linter', re: /node-test-linter/ }]
+  [LINTER, { name: 'Linter', re: /node-test-linter/ }],
+  [LITE, { name: 'Lite Commit CI', re: /node-test-commit-lite/ }]
```
richardlau: Shouldn't this be `/node-test-commit-light/`?

cPhost: Thanks for the review, i got that fixed locally with tests.

lib/ci.js
```diff
@@ -22,7 +23,8 @@ const CI_TYPES = new Map([
   [LIBUV, { name: 'libuv', re: /libuv/ }],
   [NOINTL, { name: 'No Intl', re: /nointl/ }],
   [V8, { name: 'V8', re: /node-test-commit-v8/ }],
-  [LINTER, { name: 'Linter', re: /node-test-linter/ }]
+  [LINTER, { name: 'Linter', re: /node-test-linter/ }],
+  [LIGHT, { name: 'Lite Commit', re: /node-test-commit-light/ }]
```
joyeecheung: There is also node-test-pull-request-lite which is the parent job of node-test-commit-light

richardlau: That's currently caught by the regexp for `FULL` above.

cPhost: I will add that in too then, funny enough my PR title has this in it since i was confused between lite and light.

@richardlau is not caught by `FULL` regex since is closed at the end by a slash (`\/`)

richardlau: The trailing slash is only on the right side of `|`. The left side would match.

cPhost: Yep, its being detected i can confirm (using devtools):
```javascript
> /node-test-pull-request|node-test-commit\//.test('https://ci.nodejs.org/node-test-pull-request-lite/123')
true
```

cPhost:
@joyeecheung @richardlau are there any other ci's we need to include here?

richardlau:
Looking at the list in https://github.com/nodejs/node/blob/master/COLLABORATOR_GUIDE.md#useful-ci-jobs I think we're now covered -- I don't think we would want to merge PR's where `node-stress-single-test` (not currently covered) was the only CI run.

lib/ci.js
richardlau: FYI @gibfahn has just renamed the `node-test-commit-lite` job: https://github.com/nodejs/node-core-utils/issues/214#issue-303916863

cPhost: I will update the PR accordingly. Thanks, @richardlau for the heads up

cPhost:
I updated the PR now to use `node-test-commit-lite` so the PR should be good to go, also I don't remember why was this PR stalled before did it need something else to be done if not this PR needs a review.

lib/ci.js
```diff
@@ -22,7 +23,8 @@ const CI_TYPES = new Map([
   [LIBUV, { name: 'libuv', re: /libuv/ }],
   [NOINTL, { name: 'No Intl', re: /nointl/ }],
   [V8, { name: 'V8', re: /node-test-commit-v8/ }],
-  [LINTER, { name: 'Linter', re: /node-test-linter/ }]
+  [LINTER, { name: 'Linter', re: /node-test-linter/ }],
+  [LITE, { name: 'Lite', re: /node-test-commit-lite/ }]
```
joyeecheung: It is node-test-pull-request-lite and node-test-commit-light

richardlau: @gibfahn renamed `node-test-commit-light` to `node-test-commit-lite`: https://github.com/nodejs/build/issues/1178

joyeecheung: Oh, missed that one. But the PR job should still be included

richardlau: https://github.com/nodejs/node-core-utils/pull/171#discussion_r167386568

joyeecheung: @richardlau It doesn’t seem right to capture that as a Full CI when we have a new category named Lite?

richardlau: In that case the regexp for full would also need to be modified to stop it matching.

cPhost:
I updated it as @joyeecheung suggested.

cPhost:
Merged, Thanks everyone for reviews.

Issue #148, opened 2018-01-18
tests for ncu-config


joyeecheung:
There are no tests for the `ncu-config` command at the moment. Related files:

- `bin/ncu-config`
- `lib/file.js`
- `lib/config.js`

alopezsanchez:
I'm on it.

PR #176, opened 2018-02-11
test: add `ncu-config` tests

alopezsanchez:
Fixes: https://github.com/nodejs/node-core-utils/issues/148


test/unit/config.test.js
```diff
@@ -0,0 +1,168 @@
...
+const localFileContent =
+  `{
+    "username": "${localUsername}"
+  }`;
+const newUsername = 'bar';
+const newToken = 'asdfghj';
+
+describe('config', () => {
+  beforeEach(() => {
+    mockFs({
```
cPhost: Does this really needs to get re-mocked before every single test? I think it doesn't need to be so can we move this outside of describe and restore at the end or instead of `beforeEach` and `afterEach` make it `before` and `after`.

test/unit/file.test.js
```diff
@@ -0,0 +1,110 @@
...
+const mockFileName = 'file.json';
+const filePath = path.join(mockDirName, mockFileName);
+
+const newUsername = 'bar';
+const newToken = 'asdfghj';
+
+describe('file', () => {
+  beforeEach(() => {
+    mockFs({
+      [mockDirName]: {
```
cPhost: Right here same comments as above, since the values do not change there is no need to reinit the fs mock so this should only be done once right.

test/unit/config.test.js
alopezsanchez: The file is changing in every `writeFile`, `writeJson`, `writeConfig` and `updateConfig`, so we need to isolate test cases.

I could put the `mockFs.restore()` in the specific `describe`, but I think that makes the code less readable.

cPhost: Ok that make sense.

test/unit/config.test.js
```diff
@@ -0,0 +1,168 @@
...
+      // Mock global .ncurc
+      [path.join(os.homedir(), '.ncurc')]: globalFileContent,
```
joyeecheung: If the user has `XDG_CONFIG_HOME` set when running the tests, this is not enough right? We need to manually set `XDG_CONFIG_HOME` to undefined in the before hook to make sure it does not read from there.

package.json
```diff
@@ -53,6 +53,7 @@
     "eslint-plugin-standard": "^3.0.1",
     "intelli-espower-loader": "^1.0.1",
     "mocha": "^4.0.1",
+    "mock-fs": "^4.4.2",
```
joyeecheung: I just took a look at this library and it is definitely going to be broken by Node.js master (v10)...it might worth adding it to citgm-all and see how we could fix it. I also had some conversation with @jasnell yesterday about providing a hook so testing libraries can monkey-patch this kind of thing reliably...

joyeecheung: BTW to mock the fs methods, simply overriding `fs.something` on the require cache should be enough

test/unit/config.test.js
alopezsanchez: Yes, you are right. Thanks @joyeecheung!

package.json
alopezsanchez: Okay, I didn't know. Then, what we should do with this?

joyeecheung: @alopezsanchez I think simply overriding `fs.something = mock` in the hooks is enough?

alopezsanchez: @joyeecheung, done!

But... I think I made a mistake. There was a conflict with `package.json` and I resolved it with GitHub and it obviously generated 5d3d0005867b30b20b9f5d215a59c66c5a6ab7e5.

This is not correct, right? How can I solve it without worsening the log?

cPhost: Hmm. Sorry for merge conflict's after npm packages update. In future you definitely want to handle this locally by pull latest master through rebase way and not using github's editor.

Any just pull the latest master by `git pull upstream master --rebase` and fix all the conflict and push it and this should be fixed. Also while you at it you should probably remove the last commit `Merge branch ...` and then do this.

alopezsanchez: Hello @joyeecheung @cPhost. Sorry, I totally forgot this PR. I resolved the commit issue!

alopezsanchez:
Hello @joyeecheung, can we merge this?

BridgeAR:
This needs a rebase. @joyeecheung could you take another look?

package.json
```diff
@@ -52,6 +52,7 @@
     "eslint-plugin-promise": "^3.6.0",
     "eslint-plugin-standard": "^3.0.1",
     "intelli-espower-loader": "^1.0.1",
+    "lodash.clonedeep": "^4.5.0",
```
joyeecheung: We already depend on `lodash` now

codebytere:
Does this have a path forward or can it be closed?

PR #178, opened 2018-02-12
 git-node: add core-validate-commit to dependencies

targos:
This also includes a commit to fix git-node on Windows (did not work at all before)

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -1,8 +1,5 @@
 'use strict';
 
-const path = require('path');
-const isWindows = require('os').platform() === 'win32';
-
 const {
   runAsync, runSync, forceRunAsync, exit
 } = require('./run');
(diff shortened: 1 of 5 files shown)


lib/landing_session.js
```diff
@@ -159,7 +162,11 @@ class LandingSession extends Session {
     const branch = this.branch;
     const notYetPushed = this.getNotYetPushedCommits();
     const notYetPushedVerbose = this.getNotYetPushedCommits(true);
-    await runAsync('core-validate-commit', notYetPushed);
+    const validateCommand = path.join(
```
joyeecheung: I think it makes sense to just go with `npx core-validate-commit`?

targos:
updated to use npx

package.json
```diff
@@ -32,6 +32,7 @@
   "license": "MIT",
   "dependencies": {
     "chalk": "^2.2.0",
+    "core-validate-commit": "^3.5.0",
```
mcollina: I don't think we should include it as a dependency if we are using npx. It would download it automatically.

cPhost:
Merged after squashing third commit as 03c1ff528fd68ec53b544cf8681af973ababe2bf and 75f2b210d59d28a52a6bbd3c5f6df0c546af65e8, Thanks everyone!

package.json
cPhost: Oops i missed this..., and landed this maybe do a followup.

targos: I prefer to have as a dependency. Otherwise it's slower and inefficient (always reinstalling the package)

mcollina: then what is the benefit of using npx?

cPhost: I think i this case it was to avoid adding `.cmd` to script in `./node_modules/,bin/core-validate-commit` adn writing out the path to script.

targos: Yes, it was my intention but I think I was wrong. The command is executed from the `node` directory so it cannot see `core-validate-commit`.

PR #181, opened 2018-02-16
warn if the PR Author does not have valid config setup

cPhost:
This displays a warning if PR Author does not have valid config setup using `git config`.


lib/pr_summary.js
```diff
@@ -34,8 +34,16 @@ class PRSummary {
...
     const authorHint =
       this.data.authorIsNew() ? ', first-time contributor' : '';
-    cli.table('Author',
-      `${author.name} <${author.email}> (@${author.login}${authorHint})`);
+
+    if (author.name && author.email) {
+      cli.table('Author',
+        `${author.name} <${author.email}> (@${author.login}${authorHint})`);
+    } else {
+      // user do not have correct `git config` setup
```
joyeecheung: I think the information here comes from their github account settings? The `git config` setup are only visible in the commit information.

Take a look at the profile page https://github.com/antoine-amara and they don't have the name nor the email setup.

cPhost: Yeah actually not too sure about i guessed it since if you look at the issues the commit the user reference have correct config of commit.

cPhost: > Take a look at the profile page https://github.com/antoine-amara and they don't have the name nor the email setup.

I don't think that matters see the issue link and the commits that are referenced they seem to perfect commit except the last one (which was the one in the PR).

joyeecheung: I don't think it's possible to get an empty `git config`, since git should prompt for a username and a email when you commit for the first time. So the commits should always come with a name and a email (could be totally bogus, but still). It is possible to have empty github profile settings if you don't put those down during github account registration, or deletes them afterwards.

cPhost: See that something i am unsure about too. I have no idea how that occurred at the first place.

cPhost: It seems like the user squashed a that commit so it might be that he changed his email causing the commit to empty details. Since it does have users profile picture.

joyeecheung: @cPhost Hmm..maybe the user has set the email to private? I believe that would cause the email queried in the API to be `null` but Github still knows how to associate the commits with the account since they know about the private email address. Either way the warning should be something like "Could not retrieve the email from the PR author's GitHub profile"

cPhost: Ok, cool i can update the warning, also the name is also not present so i will add that in too.

test/fixtures/data.js
```diff
@@ -59,6 +59,7 @@ const firstTimerPrivatePR = readJSON('first_timer_pr_with_private_email.json');
 const semverMajorPR = readJSON('semver_major_pr.json');
 const fixAndRefPR = readJSON('pr_with_fixes_and_refs.json');
 const conflictingPR = readJSON('conflicting_pr.json');
+const incorrectConfigPR = readJSON('incorrect_config_pr.json');
```
joyeecheung: Maybe renaming this to something like `emptyProfilePR`

cPhost: Ah, yeah makes sense since this seems to be something other than incorrect config i was thinking.

PR #185, opened 2018-02-20
ncu-team: retrieve the whole team

joyeecheung:
Do not cap on the 100 limit.

### DIFF
--- a/lib/queries/Team.gql
+++ b/lib/queries/Team.gql
@@ -1,10 +1,15 @@
-query team($org: String!, $team: String!) {
+query team($org: String!, $team: String!, $after: String) {
   organization(login: $org) {
     login
     team(slug: $team) {
       name,
       description,
-      members(first: 100) {
+      members(first: 100, after: $after) {
+        totalCount
+        pageInfo {
+          hasNextPage
+          endCursor
+        }
         nodes {
           login,
           name,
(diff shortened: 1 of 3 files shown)


cPhost:
I assume that you plan to check on `hasNextItems` and if so request the list for last `totalCount - 100`. 

BTW i notices something from last time members can't retrieves the list i think (at-least i can't) so we might need to have a not for this somewhere. Sorry i should have brought this up before.

joyeecheung:
@cPhost The whole pagination thing has already implemented and used by other queries, see

https://github.com/nodejs/node-core-utils/blob/a15715ad1434ed12287adf15d00ed35123a654a8/lib/request.js#L62-L85

This PR just uses that by passing the path params and fix the query.

joyeecheung:
> BTW i notices something from last time members can't retrieves the list i think (at-least i can't) so we might need to have a not for this somewhere. Sorry i should have brought this up before.

Do you mean you don't have the permission to get the list? Have you configured the `read:org` scope in your token?

cPhost:
Maybe thats it, it says `it looks like you have correct authorization ... but not allow to access data`. My token is since the repo started so its only for the `email` one.

joyeecheung:
@nodejs/automation Anyone wants to review this?

cPhost:
@joyeecheung this lgtm to me but was waiting for you to fix the tests.

riyadhalnur:
The coverage reporter failed and not the tests actually

joyeecheung:
@cPhost Thanks for catching that..I am probably too used to red CIs these days :(

cPhost:
@joyeecheung no problem i though you were on it :), anyways the test lgtm since it just involved some promise fixed.

PR #189, opened 2018-02-23
docs: deprecate get-metadata -> git node metadata

joyeecheung:


cPhost:
Are we planning to remove `get-metadata` in near future or this is what core calls document only deprecation?

targos:
We should probably runtime-deprecate it as well before removing.

Issue #130, opened 2017-12-12
Exit code non zero


BridgeAR:
I think it would be nice if the call would return a non-zero exit code in case a requirement is missing.

cPhost:
Sure, do you want to make a PR it just adding non-zero exit if status is false.

ryzokuken:
@BridgeAR @cPhost I'd love to have a shot at this one, I'd just need a little more detail regarding what exactly needs to be done.

joyeecheung:
I think this has been fixed now? AFAICT, the commands are using `runPromise` to run things and that exits with 1 in `.catch` (which reminded me there seem to be another bug...all the `.catch` attached to `runPromise` or `runAsync` calls will not be called because of that)

ryzokuken:
@joyeecheung If you're positive about it, please proceed to close this issue. Also, let me know if you think the redundant `.catch` calls should be removed, and I'll look into it.

cPhost:
This is still not fixed yet, all that needs to be done is, add a `then` to `getMetadata` and if the status if false exit with non-zero exit code.

ryzokuken:
@cPhost isn't `get-metadata` deprecated?

PR #199, opened 2018-03-05
get-metadata: fail with a non-zero exit status

ryzokuken:
Fixes: https://github.com/nodejs/node-core-utils/issues/130

@joyeecheung @cPhost Please take a look. I hope this is what was needed to be done.

### DIFF
--- a/components/git/metadata.js
+++ b/components/git/metadata.js
@@ -86,7 +86,7 @@ function handler(argv) {
   return getMetadata(Object.assign({}, config, argv, parsed), cli)
     .then(({status}) => {
       if (status === false) {
-        process.exit(1);
+        throw new Error('PR checks failed');
       }
     })
     .catch((err) => {
@@ -94,7 +94,7 @@ function handler(argv) {
         cli.spinner.fail();
       }
       cli.error(err);
-      process.exit(-1);
+      process.exit(1);
     });
 }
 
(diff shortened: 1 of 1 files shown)


components/git/metadata.js
```diff
@@ -83,15 +83,19 @@ function handler(argv) {
...
-    if (cli.spinner.enabled) {
-      cli.spinner.fail();
-    }
-    cli.error(err);
-    process.exit(-1);
-  });
+  return getMetadata(Object.assign({}, config, argv, parsed), cli)
+    .then(({status}) => {
+      if (status === false) {
+        process.exit(1);
```
joyeecheung: I think we can just let this throw here so it goes to the catch below, and replace `process.exit(-1)` below to `process.exit(-1)`? (I don't even know what I was thinking when I put `-1` there :/)

ryzokuken: I think you mean `process.exit(1)`?

joyeecheung: Yes, like so

```
    .then(({status}) => {
      if (status === false) {
        throw new Error(`PR checks failed`);
      }
    })
    .catch((err) => {
      if (cli.spinner.enabled) {
        cli.spinner.fail();
      }
      cli.error(err);
      process.exit(1);
    });

```

ryzokuken: Cool. That was exactly what I had been thinking about.

ryzokuken:
@joyeecheung take a look now.

BridgeAR:
@ryzokuken thanks a lot!

Issue #198, opened 2018-03-05
abort doesn't undo reset


bmeck:
When you run `git node land --abort` it resets the branch but then does not undo that once it aborts. This can lead to some confusion since it leaves the branch without the commits it included previously. So this can lead to odd looking behavior when you sync your branch back up to github and/or detached HEAD. See https://github.com/nodejs/node/pull/18914#issuecomment-370455755 for example of confusion caused by this.

gibfahn:
Copy-pasting @joyeecheung 's comment from https://github.com/nodejs/node/pull/18914#issuecomment-370487639:

>>    @joyeecheung bradley mentioned ncu, was node land final for single commits ever fixed? it could be related
>
>I could not reproduce that so no sure how to get it fixed though....
>
>@bmeck Did you run git node land --abort on your PR branch? That would reset your PR branch to upstream/master (but it will prompt you first before doing the reset). I guess if you push the upstream/master head to this branch, then github might auto close it since there is nothing to merge - but I am pretty sure you'll have to force-push to this branch to make that happen. Have you force pushed?

bmeck:
this source of my confusion might be addressed in https://github.com/nodejs/node-core-utils/pull/200

PR #200, opened 2018-03-05
git-node: refuse to run without configurations/on wrong revs

joyeecheung:
It now refuses to run if

1. User has not configured upstream or branch - no more defaults.
2. User is not on the branch configured
3. User is on detached HEAD

Refs: https://github.com/nodejs/node/pull/18914
Fixes: https://github.com/nodejs/node-core-utils/issues/198


gibfahn:
Maybe add a `Fixes: https://github.com/nodejs/node-core-utils/issues/198`

PR #201, opened 2018-03-05
git-node: warn on unexpected remote repository url

bmeck:
This is just a double check so it emits warnings rather than doing anything. It doesn't parse `git+ssh` URLs since they are not valid URLs according to http://url.spec.whatwg.org/ .

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const path = require('path');
+const { URL } = require('url');
 
 const {
   runAsync, runSync, forceRunAsync, exit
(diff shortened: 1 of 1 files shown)


lib/landing_session.js
```diff
@@ -14,6 +15,33 @@ class LandingSession extends Session {
...
+      `remote.${upstream}.url`]);
+    let warned = false;
+    try {
+      const { hostname, pathname } = new URL(upstreamHref);
+      if (hostname !== 'github.com') {
+        warned = true;
+        cli.warn('Remote repository URL does not point to the expected ' +
+          'host "github.com"');
+      }
+      if (/^\/nodejs\/node(?:\.git)?$/.test(pathname) !== true) {
```
joyeecheung: Can you change this to `new RegExp(`^\/${this.owner}\/${this.repo}(?:\.git)?$/`)` so these can be configured? Also the warning below (`this.owner` and `this.repo` come from the `Session` class this is inheriting, they default to `nodejs` and `node`).

PR #202, opened 2018-03-05
git-node: do not require https:// urls or github.com urls

joyeecheung:
It currently outputs a warning for SSH urls because WHATWG URL does not support that. Also does not trim the output from `git config` so there might be false warnings. This might create confusion to users who have SSH urls since we do not disencourage that in core.

Update to only check the owner and repo instead.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const path = require('path');
-const { URL } = require('url');
 
 const {
   runAsync, runSync, forceRunAsync, exit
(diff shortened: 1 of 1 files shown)


lib/landing_session.js
```diff
@@ -16,31 +15,12 @@ class LandingSession extends Session {
...
-    }
-    if (warned) {
-      cli.warn('You might want to change the remote repository URL with ' +
-        `\`git remote set-url ${
-          upstream
-        } "https://github.com/${owner}/${repo}"\``);
+    const upstreamHref = runSync('git', [
+      'config', '--get',
+      `remote.${upstream}.url`]).trim();
+    if (!upstreamHref.endsWith(`${owner}/${repo}.git`)) {
```
cPhost: Hmm, maybe not check for `.git` in the end, since you can emit adding `.git`?

joyeecheung: @cPhost Is it possible though? At least the GitHub UI only offers `git@github.com:nodejs/node.git` and `https://github.com/nodejs/node.git` in https://github.com/nodejs/node

cPhost: Yep, when sometimes I am lazy I do `git remote add upstream git@github.com:user/repo` and it works fine, and I just tested if you do `git config --get` it will return URL without `.git` part thus causing a false

joyeecheung:
@cPhost updated, PTAL, thanks!

PR #207, opened 2018-03-08
metadata: remove file hashes from passed url

cPhost:
The `/files` at the end of URL is handled by `PR_RE` already, so this pr just removes hashes is passed as url.


cPhost:
This needs a quick review @nodejs/automation, thanks.

cPhost:
Merged, @gibfahn thanks for the review.

PR #208, opened 2018-03-08
lib,src: fixes for Windows and error output

kfarnung:
* On Windows the node executable needs to be explicitly invoked.
* Make sure the data property exists before trying to use it when
  printing error output.


cPhost:
This was also fixed back then in #178, shebang lines (of unix style) does not work on windows.

cPhost:
Merged, Thanks @kfarnung!

Issue #213, opened 2018-03-09
Allow git node land with full PR urls


gibfahn:
https://github.com/nodejs/node-core-utils/issues/206#issuecomment-371827740

>Just to add, it looks like you can't currently do `git node land https://github.com/nodejs/node/pull/19130`.

@cPhost

aks-:
@gibfahn can you please assign it to me? I'll take a look in free time

PR #219, opened 2018-03-16
allow git-node land to work with full PR url (#213)

aks-:
Fixes: https://github.com/nodejs/node-core-utils/issues/213

### DIFF
--- a/components/git/epilogue.js
+++ b/components/git/epilogue.js
@@ -11,6 +11,7 @@ $ ncu-config set branch master   # Assuming you are landing commits on master
 $ git checkout master
 $ git node land --abort          # Abort a landing session, just in case
 $ git node land $PRID            # Start a new landing session
+$ git node land $URL             # Start a new landing session
 
 $ git rebase -i upstream/master  # Put "edit" on every commit that's gonna stay
 
--- a/components/git/epilogue.js
+++ b/components/git/epilogue.js
@@ -11,7 +11,7 @@ $ ncu-config set branch master   # Assuming you are landing commits on master
 $ git checkout master
 $ git node land --abort          # Abort a landing session, just in case
 $ git node land $PRID            # Start a new landing session
-$ git node land $URL             # Start a new landing session
+$ git node land $URL             # Start a new landing session using the PR URL
 
 $ git rebase -i upstream/master  # Put "edit" on every commit that's gonna stay
 
--- a/components/git/land.js
+++ b/components/git/land.js
@@ -7,7 +7,6 @@ const { runPromise } = require('../../lib/run');
 const LandingSession = require('../../lib/landing_session');
 const epilogue = require('./epilogue');
 const yargs = require('yargs');
-const isUrl = require('is-url');
 
 const landOptions = {
   apply: {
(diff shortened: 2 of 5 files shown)


components/git/land.js
```diff
@@ -59,7 +61,7 @@ const CONTINUE = 'continue';
 const ABORT = 'abort';
 
 function handler(argv) {
-  if (argv.prid && Number.isInteger(argv.prid)) {
+  if (argv.prid && (Number.isInteger(argv.prid) || isUrl(argv.prid))) {
```
joyeecheung: We don't really need this module, do we? We can simply use a regexp to validate & capture the id: `argv.prid.match(/github.com/\w+/\w+/pull/(\d+)/)`

aks-: @joyeecheung makes sense, also helps having less direct dependencies as they are 👍. making changes in few

components/git/epilogue.js
```diff
@@ -11,6 +11,7 @@ $ ncu-config set branch master   # Assuming you are landing commits on master
 $ git checkout master
 $ git node land --abort          # Abort a landing session, just in case
 $ git node land $PRID            # Start a new landing session
+$ git node land $URL             # Start a new landing session
```
joyeecheung: Can you put something down in the comment indicating this is an *alternative* way to land a PR, not a necessary step? Something like

```
git node land $URL             # Or, start a new landing session using the PR URL
```

aks-:
@joyeecheung updated. i think it should ready to land now

components/git/epilogue.js
```diff
@@ -11,6 +11,7 @@ $ ncu-config set branch master   # Assuming you are landing commits on master
 $ git checkout master
 $ git node land --abort          # Abort a landing session, just in case
 $ git node land $PRID            # Start a new landing session
+$ git node land $URL             # Start a new landing session using the PR URL
```
gibfahn: Just a suggestion, but it might make more sense to merge this into the line above, and just say that `$PRID` can be the PR number or the PR URL.

aks-: @gibfahn hm, i don't think it makes big difference, i think it's fine to have them on separate line? i'm okay with changing it though. just let me know...

gibfahn: I also don't think it makes a big difference, which is why it was just a suggestion. If you think it's fine as is, then that's fine by me.

aks-: cool, then ready to land i guess?

docs/git-node.md
```diff
@@ -32,6 +32,7 @@ $ ncu-config set branch master   # Assuming you are landing commits on master
 $ git checkout master
 $ git node land --abort          # Abort a landing session, just in case
 $ git node land $PRID            # Start a new landing session
+$ git node land $URL             # Start a new landing session
```
joyeecheung: Can you update this line as well?

aks-: @joyeecheung done

aks-:
@joyeecheung @gibfahn changes applied. can you please check now?

gibfahn:
I already approved, if @joyeecheung approves then it can be landed.

PR #221, opened 2018-03-21
lib: fix XDG_CONFIG_HOME not being set

joyeecheung:
`git-node` is currently broken if `XDG_CONFIG_HOME` is not set becauseit's a real `undefined` if it's not set in the environment (whereas in tests, where it's explicitly set undefined, it will be coerced to `'undefined'`).

### DIFF
--- a/lib/ncurc_path.js
+++ b/lib/ncurc_path.js
@@ -2,7 +2,8 @@ const path = require('path');
 const os = require('os');
 
 function getNcurcPath() {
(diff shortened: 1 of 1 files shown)


cPhost:
Thanks @joyeecheung for the fix, as you can tell i was bit confused with `process.env` coerceing thing now it all makes sense.

gibfahn:
@joyeecheung if we did `delete process.env.XDG_CONFIG_HOME` in the tests instead of just setting it to `undefined` would that have caught this issue?

If so we should do that 😁 .

PR #222, opened 2018-03-21
lib: make NODE_PENDING_DEPRECATION happy

joyeecheung:
This commit replaces request with node-fetch and jsdom (which
depends on request) with cheerio to get rid of DEP0005 warnings
under `export NODE_PENDING_DEPRECATION=1`.

Also renames `request.promise` to `request.text` since that's
the only use case it has in the current code base

Also removes the tests on the spinners because changes in ora
would break it. We should not be testing the output of ora in
our own tests.


joyeecheung:
Yikes, forgot to switch mirrors when I generate package-lock.json. Regenerated with npmjs.org.

PR #224, opened 2018-04-01
[WIP] git-hooks: add ncu-git-hooks to install pre-commit hook

priyank-p:
This still needs some work on some parts of code but mostly is ready. Also needs docs and perhaps tests.

Basic Way to use:
```bash
ncu-git-hooks install <hook>
```

And after that, it should run `./configure && make -j4 test` on Unix while `vcbuild test` on windows when you make a commit. It should also work on windows, if not we can use PowerShell later.

### DIFF
--- a/bin/ncu-git-hooks
+++ b/bin/ncu-git-hooks
@@ -1,0 +1,44 @@
+#!/usr/bin/env node
+'use strict';
+
+const yargs = require('yargs');
+const gitHooks = require('../lib/git-hooks');
+
+/* eslint-disable no-unused-expressions */
+yargs
+  .command({
+    command: 'install <hook>',
+    desc: 'Install a git hook',
+    aliases: ['i'],
+    builder: (yargs) => {
+      yargs
+        .positional('hook', {
+          describe: 'The hook to install',
+          choices: gitHooks.possibleGitHooks
+        });
+    },
+    handler: installHandler
+  })
+  .command({
+    command: 'uninstall <hook>',
+    desc: 'Uninstall a git hook',
+    builder: (yargs) => {
+      yargs
+        .positional('hook', {
+          describe: 'The hook to uninstall',
+          choices: gitHooks.possibleGitHooks
+        });
+    },
+    handler: uninstallHandler
+  })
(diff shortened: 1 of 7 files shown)


joyeecheung:
Thanks for picking this up!

> And after that, it should run ./configure && make -j4 test on Unix while vcbuild test on windows when you make a commit. 

I don't think that's what most people do though? Also I've seen many people developing on a local branch or in their own fork before submitting a PR and the code there is usually not ready for a full test before it gets out of WIP.

cPhost:
That makes `pre-commit` easier then, so it would be just `make lint` or `vcbuild lint`.

cPhost:
Currently works like this:
git-hooks

joyeecheung:
@cPhost From what I can tell most people would not want running linters on commit in their workflow, which is why ‘make test-only’ exists. Mind that commit is not reserved for changes that will be summitted right away, a lot of times it’s just about checking WIP code into the source control but at that point you don’t really care about linters or tests since it is just a WIP.

cPhost:
So we want to run `make test-only` for each commit instead of linter or test?

It looks like the workflow of node is quite not the same as other projects, which is because in the PR mostly some commit are not standalone sometimes and so they are squashed into one while landing.

Is it that we don't need `pre-commit`? I used it because was pretty easy to do than `pre-push` where we would need to get commit hash of each commit made and send it through `core-validate-commit`. (+ some more things to consider too. so didn't quite want to take it just yet).

joyeecheung:
The primary reason I would say is that `make test` takes a lot time to run, especially when you are switching between branches and need to recompile. Also yes because node takes the git commit history seriously and most of the times the commits should be able to pass standalone.

joyeecheung:
> So we want to run make test-only for each commit instead of linter or test?

IMO nothing should be done in the pre-commit hooks. The ideal would be to only run those things when they are being pushed to a branch with PR opened.

cPhost:
OK, makes sense, I will start implementing `pre-push` hook then.

codebytere:
Can this be closed?

priyank-p:
I will close this out!

Issue #228, opened 2018-04-06
git-node: `git node status`


joyeecheung:
- Similar to `git status`
- The status of `git-node land` should be in `.ncu/land` already. We just need to display it properly.

GithubGoldMiner:
Hi @joyeecheung, I can try to create a PR for this :)

PR #232, opened 2018-04-07
Git node status

GithubGoldMiner:
This commit is related to #228 , if solution is in the right direction, I will then add tests accordingly. :)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,7 +5,8 @@
     "space-before-function-paren": ["error", "never"],
     "no-multi-spaces": ["error", { "ignoreEOLComments": true }],
     "camelcase": "off",
-    "max-len": [2, 80, 4, {"ignoreUrls": true}]
+    "max-len": [2, 80, 4, {"ignoreUrls": true}],
+    "object-property-newline": "off"
   },
   "env": {
     "node": true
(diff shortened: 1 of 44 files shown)


joyeecheung:
Thanks for picking this up. Although I am afraid logging the JSON out probably is going to be quite  confusing to users who don't really understand the structure of it..git-node currently also tries to serialize the access token there, which makes the output even more weird. I think some textual output generating from the `state`, `prid`, `config.upstream`, `config.branch` would be better, although I don't quite remember what the possible combinations are without digging into that part of the code base.

GithubGoldMiner:
@joyeecheung Yes I agree with your concern. I will look more into that part of the codebase to find a more proper solution and modify my pr. Thank you so much for your valuable response!

GithubGoldMiner:
@joyeecheung Hi, I am wondering, for `git node status`. We just need to display `state`, `prid`, `config.upstream` and `config.branch` in a meaningful way?

joyeecheung:
@GithubGoldMiner That would be a good start. The information in the config can be useful as well. Also it would be nice to display additional state of git if it’s in the middle of a git am or git rebase (no docs for that but we have code relying on some special files created by git for state management

GithubGoldMiner:
@joyeecheung Absolutely! I will look into those in the codebase and commit soon! Thank you!

codebytere:
@GithubGoldMiner would you still want to work on this or can it be closed?

codebytere:
Closing as outdated/abandoned. If someone wishes to pick this up again, please reopen/leave a message!

PR #236, opened 2018-04-17
lib: don't print dash for users without names

MylesBorins:
when logging team data we currently include a dash between
the profile name / link and the account name. If the account
name is blank, we still include the dash. This patch changes
the behavior to remove the dash in cases that the name
is blank.


MylesBorins:
🎉

PR #237, opened 2018-04-19
git-node: suggest -x and --autosquash for git rebase

joyeecheung:
- Suggest to add `-x "git node land --amend"` which will put
  `exec git node land --amend` on every picked commit.
- Suggest to add `--autosquash` if `fixup!` or `squash!` appears
  in the commit messages, which will add fixup or squash actions
  to corresponding commits if they match the autosquash rules.

Used this to land https://github.com/nodejs/node/pull/20121 . Worked pretty well.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -3,7 +3,7 @@
 const path = require('path');
 
 const {
-  runAsync, runSync, forceRunAsync, exit
+  runAsync, runSync, forceRunAsync
 } = require('./run');
 const Session = require('./session');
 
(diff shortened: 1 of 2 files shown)


joyeecheung:
BTW the `-x` idea came from @addaleax

srl295:
Using git-interactive-rebase-tool, I got this result:

```
$ git rebase upstream/master -i -x "git node land --amend"
Error reading file, /Users/srl/src/node/.git/rebase-merge/git-rebase-todo
Reason: Invalid action: exec
Could not execute editor
```

It's not a problem with `git-node`, but it might be worth noting in case someone runs into this.

PR #243, opened 2018-05-01
git-node v8: remove eu-strip during major upgrades

targos:
Ref: https://github.com/nodejs/node/pull/20304

### DIFF
--- a/lib/update-v8/util.js
+++ b/lib/update-v8/util.js
@@ -3,8 +3,6 @@
 const fs = require('fs-extra');
 const path = require('path');
 
-const constants = require('./constants');
-
 function getNodeV8Version(cwd) {
   try {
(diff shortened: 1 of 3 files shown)


targos:
Note: Node's gitignore was updated in https://github.com/nodejs/node/pull/20304 to ignore the `eu-strip` directory. It is not enough because the rule is overwritten by V8's gitignore.
The new task removes the V8 rule and the local directory.

lib/update-v8/util.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 
-const fs = require('fs');
+const fs = require('fs-extra');
+const path = require('path');
 
 const constants = require('./constants');
```
priyankp10: This line needs be removed, the lint fails here since its not required anymore.

PR #244, opened 2018-05-03
package.json: remove codecov from devDependencies

joyeecheung:
travis: remove node v9 and add v10
package.json: remove codecov from devDependencies

There is no need to install codecov on developer machines.
To enable it in Travis, we only need to install it in the Travis
setup. This would help us remove some unnecessary entries in
package-lock.json as well.


.travis.yml
```diff
@@ -1,11 +1,14 @@
...
-  - "9"
   - "8"
+  - "10"
 sudo: false
 cache:
   directories:
     - node_modules
+install:
+  - npm install
+  - npm install codecov
```
targos: suggestion: remove `install` and move the `npm install codecov` line to `after_success`

joyeecheung:
@targos I've moved the `npm install codecov` to `after_success` and updated nyc to get rid of the vulnerability warning of `npm audit`. PTAL, thanks!

PR #247, opened 2018-05-26
package: update package-lock.json

joyeecheung:
Mostly to udpate core-validate-commit to 3.5.1

Refs: https://github.com/nodejs/core-validate-commit/pull/26


joyeecheung:
cc @nodejs/automation

joyeecheung:
@priyankp10 

package.json lists that as `"core-validate-commit": "^3.5.0",` so it does not make much difference to update the patch version.

PR #249, opened 2018-06-02
remove outdated instructions

Trott:
h/t @joyeecheung


joyeecheung:
The CI fails due to timeout on auth...probably unrelated, but can you amend and force-push to trigger the CI again? @Trott

PR #250, opened 2018-06-03
ncu-ci: handle successful CI runs

joyeecheung:
Apparently I have never tried this on a green CI run before because I couldn't find one, but I just had one https://ci.nodejs.org/job/node-test-commit/18960/ (yay!) so `ncu-ci` turned out to be broken on actual green builds. This patch fixes that.


priyankp10:
Merged, thanks @joyeecheung!

Issue #257, opened 2018-06-19
ncu-ci: authenticate before accessing jenkins


joyeecheung:
`ncu-ci` is now broken because we stopped all unauthenticated access to ci.nodejs.org now. We should:

1. Add a `jenkins_token` to the config
2. Use that token to authenticate requests to jenkins
3. Add docs on how to create the jenkins token

Issue #259, opened 2018-06-21
git-node: count jenkins pipeline and/or Travis as effective CI runs


joyeecheung:
Refs: https://github.com/nodejs/github-bot/pull/178 that runs the jenkins pipeline for collaborator PRs. The bot currently comments on the PR with a link to the job after the build is started.
Refs: https://github.com/nodejs/node/pull/21059 which runs the Travis CI for non-collabortor PRs. The existence of the job can only be tracked in the Github status API or via Travis API.

Both of them only serve the same purpose as the original Lite CI. For non-doc-only PRs we still require a full CI triggered by human (for now at least).

One way to fix this easily is, instead of only looking at the thread for CI links, also look at the status API to find jobs run for the last effective commit of the PR.
(Personally, I would want a warning to remind people to post the link to the PR because when the PR is landed, no one will ever know about those links if they only exist in the status API. I think someone raised concern about that before)

joyeecheung:
This is how the pipeline link looks: https://github.com/nodejs/node/pull/21430#issuecomment-398910078

vsemozhetbyt:
CI-lite detection is also broken now. Refs: https://github.com/nodejs/node/pull/21938

PR #260, opened 2018-06-21
auth & ncu-ci: add jenkins authentication, refactor ci parsers

joyeecheung:
Fixes: https://github.com/nodejs/node-core-utils/issues/257

After this https://github.com/nodejs/node-core-utils/issues/259 should be easier to fix as well


lib/auth.js
```diff
@@ -49,27 +38,61 @@ async function auth(getCredentials = ghauth) {
...
+    }
+    check(username, token);
+    result.github = encode(username, token);
+  }
+
+  if (options.jenkins) {
+    let { username, jenkins_token } = getMergedConfig();
+    if (!username || !jenkins_token) {
+      process.stdout.write(
+        'Get your Jenkins API token in https://ci.nodejs.org/me/configure ' +
```
joyeecheung: I have not added any documentation on how to get the Jenkins API token but it's pretty obvious if you follow this error message. It's only used when you use `ncu-ci` anyway.

PR #269, opened 2018-08-10
ci: handle node-test-commit-custom-suites-freestyle

joyeecheung:


bin/ncu-ci
```diff
@@ -16,7 +16,7 @@ const {
 
 const {
   PRBuild, BenchmarkRun, CommitBuild, listBuilds
-  , jobCache
+  // , jobCache
```
joyeecheung: This was previously accidentally committed and should be reverted

joyeecheung:
CI is green so I'll go ahead and merge this.

PR #275, opened 2018-08-17
ci: tweak patterns, improve markdown output

joyeecheung:
New format is in https://github.com/nodejs/reliability/issues/14 Added a few more patterns and linkify the machine names as well as console of example stacks

### DIFF
--- a/lib/ci/ci_result_parser.js
+++ b/lib/ci/ci_result_parser.js
@@ -416,7 +416,7 @@ class FailureAggregator {
       for (const item of aggregates[type]) {
         const { reason, type, prs, failures, machines } = item;
         if (prs.length < 2) { continue; }
-        todo.push(reason)
+        todo.push(reason);
         output += markdownRow('Reason', `<code>${reason}</code>`);
         output += markdownRow('-', ':-');
         output += markdownRow('Type', type);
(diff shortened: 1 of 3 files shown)


lib/ci/ci_result_parser.js
```diff
@@ -392,29 +401,48 @@ class FailureAggregator {
...
+    output += `[${jobName}/${last.jobid}](${last.link}) `;
+    output += `that failed more than 2 PRs\n`;
+    const todo = [];
     for (const type of Object.keys(aggregates)) {
       output += `\n### ${FAILURE_TYPES_NAME[type]}\n\n`;
       for (const item of aggregates[type]) {
         const { reason, type, prs, failures, machines } = item;
         if (prs.length < 2) { continue; }
-        output += markdownRow('Reason', `\`${reason}\``);
+        todo.push(reason)
```
priyankp10: Linting error: missing semicolon here.

joyeecheung:
There used to be a (sort-of) bug in the parser, where a pattern returns the failure object without wrapping it in the array - it's fine in practice since the arrays will be eventually flattened, but returning an array now breaks the test because the `expected.json` contains an object instead of an array. Fixed it by making the expected output an array.

Issue #279, opened 2018-09-22
git-node land: author does not have correct git config?


targos:
not sure what's wrong with this one:

```
$ git node land 22754
✔  Done loading data for nodejs/node/pull/22754
----------------------------------- PR info ------------------------------------
Title      deps: update V8 to 7.0 (#22754)
Author     Michaël Zasso <targos@protonmail.com> (@targos)
Branch     targos:v8-7.0 -> nodejs:master
Labels     V8 Engine, author ready, semver-major
Commits    8
 - deps: update V8 to 7.0.276.20
 - deps: update v8.gyp
 - src: use HeapStatistics to get external memory
 - deps: cherry-pick 2363cdf from upstream V8
 - test: update postmortem metadata test for V8 7.0
 - src: update postmortem constants
 - build: reset embedder string to "-node.0"
 - src: update NODE_MODULE_VERSION to 67
Committers 1
 - Michaël Zasso <targos@protonmail.com>
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/22754
Fixes: https://github.com/nodejs/node-v8/issues/77
Reviewed-By: Matteo Collina <matteo.collina@gmail.com>
Reviewed-By: James M Snell <jasnell@gmail.com>
Reviewed-By: Refael Ackermann <refack@gmail.com>
Reviewed-By: Ali Ijaz Sheikh <ofrobots@google.com>
Reviewed-By: Colin Ihrig <cjihrig@gmail.com>
--------------------------------------------------------------------------------
✔  Requested Changes: 0
✔  Approvals: 5, 4 from TSC (mcollina, jasnell, ofrobots, cjihrig)
ℹ  Last Full PR CI on 2018-09-22T12:47:56Z: https://ci.nodejs.org/job/node-test-pull-request/17370/
ℹ  Last V8 CI on 2018-09-22T12:47:56Z: https://ci.nodejs.org/job/node-test-commit-v8-linux/1710/
✖  Author does not have correct git config!
--------------------------------------------------------------------------------
```

priyankp10:
It's due to the https://github.com/nodejs/node/pull/22754/commits/58a2932064f752ad36185c392cf4ae698d1991ab commit; the author didn't configure their name and email using `git config`.

(I think we should come up with good descriptive title for this!)

joyeecheung:
This seems more like warning, not an error to me..

PR #280, opened 2018-09-22
git: warn instead of error in case of wrong git config

joyeecheung:
Fixes: https://github.com/nodejs/node-core-utils/issues/279


priyankp10:
Merged, thanks @joyeecheung for this followup!

Issue #283, opened 2018-10-02
When squashing commits node-core-utils should fail if subsystem prefix is missing 


trivikr:
Hi,

I faced this issue while landing commit https://github.com/nodejs/node/commit/a82ac6eedf925b8922ca2e577d0364a62dea5ad7

Creating quick issue for follow-up, will add details later.
Summary: When there are multiple commits to land and node-core-utils suggests squashing, it doesn't fail if subsystem prefix (and probably other requirements) is missing.

I'll reproduce the steps and add details later.

Regards,
Trivikram

joyeecheung:
It will fail when you run `git node land --final` though (that's when `core-validate-commit` gets invoked). Have you run that? Or just run `git node land --continue` until it invokes the validations?

trivikr:
I don't think I ran `git node land --final` before landing.
I'd just run the command which was suggested for squashing commits. I remember it as I didn't have to manually replace `pick` to `f` as I'd to do in the past.

The validation should be done after the squashing happens, either the user:
* doesn't have to run third command `git node land --final`
* or `git node land --final` runs automatically after squashing

This quick issue was created for logging purposes, I'll recreate a scenario to provide details.

trivikr:
Tested with landing fake change https://github.com/nodejs/node/pull/23235, and noticed that `git node land --final` does catch issues with commit as @joyeecheung mentioned in her comment

Outputs while trying to land:

<details><summary>$ git node land 23235</summary>
<p>

```
✔  Done loading data for nodejs/node/pull/23235
----------------------------------- PR info ------------------------------------
Title      [DO NOT LAND] Testing squashing while landing with git-node (#23235)
⚠  Could not retrieve the email or name of the PR author's from user's GitHub profile!
Branch     trivikr:trivikr-patch-1 -> nodejs:master
Labels     doc
Commits    4
 - Testing squashing while landing with git-node
 - Another commit to add to the PR
 - One more commit!
 - Adding sample text
Committers 2
 - GitHub <noreply@github.com>
 - Kamat, Trivikram <16024985+trivikr@users.noreply.github.com>
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/23235
--------------------------------------------------------------------------------
✔  Requested Changes: 0
✖  Approvals: 0
✖  No CI runs detected
ℹ  This PR was created on Tue Oct 02 2018 (weekday in UTC)
⚠  47 hours left to land
[...]

PR #285, opened 2018-10-03
git-node: message to run --final after --amend

trivikr:
Refs: https://github.com/nodejs/node-core-utils/issues/283
Suggestion: https://github.com/nodejs/node/pull/23235#issuecomment-426669640

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -187,6 +187,9 @@ class LandingSession extends Session {
     const takeMessage = await cli.prompt('Use this message?');
     if (takeMessage) {
       await runAsync('git', ['commit', '--amend', '-F', messageFile]);
+      cli.log('Please run the following command to verify' +
+        ' and fix commit message before landing\n\n' +
+        '$ git node land --final');
       return true;
     }
 


lib/landing_session.js
```diff
@@ -187,6 +187,9 @@ class LandingSession extends Session {
     const takeMessage = await cli.prompt('Use this message?');
     if (takeMessage) {
       await runAsync('git', ['commit', '--amend', '-F', messageFile]);
+      cli.log('Please run the following command to verify' +
```
joyeecheung: This would show up after every `git node land --amend`, is that intended?

joyeecheung: Also, why this is only printed when `takeMessage` (use the automatically generated message) is true?

trivikr: About message appearing after every `git node land --amend`
* Yes, it's advisable to do this so that new collaborators or ones using node-core-utils infrequently remember to run `git node land --final` before pushing commits.

About why this is only printed when `takeMessage` is true:
* I noticed that when `takeMessage` is false, collaborator is recommended to run `git commit --amend -F ${messageFile}` - so the message to run `git node land --final` will be shown when that command is run.

joyeecheung: @trivikr 

> I noticed that when takeMessage is false, collaborator is recommended to run git commit --amend -F ${messageFile} - so the message to run git node land --final will be shown when that command is run.

I don't think any code from ncu is going to run after the user runs `git commit --amend -F ...`? Because we don't have any git hook for ordinary git commits.

trivikr: How about recommending user to run following command instead?
`git node land --amend -F ${messageFile}`

which internally calls `git commit --amend -F ${messageFile}`

joyeecheung: Maybe we can just suggest them to call

```
git commit --amend -F ${messageFile}
git node land --continue
```

?

BridgeAR:
Closing as outdated. If that should still be applied, please reopen / leave a message.

Issue #131, opened 2017-12-12
Error out on young PRs without fast-track label


BridgeAR:
If a PR has a `fast-track` label on it, it should not be complained about. Right now there is a warning sign in case the PR is younger than 48 hours (do we check 72 hours on the weekend by the way?). But if the label is applied, that should not warn out of my perspective. And if the label is not applied, I would expect a error instead of just a warning.

cPhost:
@BridgeAR we do check for 72 hours on weekdays. 

> And if the label is not applied, I would expect a error instead of just a warning.

I dont think we can detect if a PR needs fast track 

Can you elaborate more on what you need error for or don not have warning for.

apapirovski:
> I dont think we can detect if a PR needs fast track

We definitely can, that's what the `fast-track` label is for.

cPhost:
@apapirovski we do warn in case when PR has `fast-track` label i think @BridgeAR was talking about detecting if PR needs `fast-tracking`

(Correct me if am wrong)

apapirovski:
In that case there doesn't seem to be much to do with this issue? I think @BridgeAR is just asking for the functionality that was implemented in https://github.com/nodejs/node-core-utils/pull/104/files

BridgeAR:
What I am asking for is a `error` instead of a `warning`. It is of course trivial to change that but I would really like to enforce using the label.

mmarchini:
`ncu` handles fast-track properly today

PR #287, opened 2018-10-05
Update pr_checker.js

srl295:
Clarify that the wait time is a *minimum* not a maximum. Also see  #131

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -185,7 +185,7 @@ class PRChecker {
       const dateStr = new Date(pr.createdAt).toDateString();
       const type = wait.isWeekend ? 'weekend' : 'weekday';
       cli.info(`This PR was created on ${dateStr} (${type} in UTC)`);
-      cli.warn(`${wait.timeLeft} hours left to land`);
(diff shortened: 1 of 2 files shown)


srl295:
failed because of tests…

lib/pr_checker.js
```diff
@@ -185,7 +185,7 @@ class PRChecker {
       const dateStr = new Date(pr.createdAt).toDateString();
       const type = wait.isWeekend ? 'weekend' : 'weekday';
       cli.info(`This PR was created on ${dateStr} (${type} in UTC)`);
-      cli.warn(`${wait.timeLeft} hours left to land`);
+      cli.warn(`Wait at least ${wait.timeLeft} hours before landing`);
```
Trott: Optional nit: Maybe to avoid more confusion make it "Wait at least N **more** hours before landing"? So that someone doesn't see that it says to wait 4 hours before landing, look at their PR and think "it's been open for two days, I can land it now because I waited 4 hours!"

srl295: 4530c4692e0b613d76f4db014549498afa520f79

PR #290, opened 2018-10-07
git-node: wait time for PR with single approval

trivikr:
One Collaborator approval is enough only if the pull request
has been open for more than 7 days

Refs: https://github.com/nodejs/node/pull/22255

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -4,8 +4,8 @@ const SECOND = 1000;
 const MINUTE = SECOND * 60;
 const HOUR = MINUTE * 60;
 
-const WAIT_TIME = 48;
-const WAIT_TIME_SINGLE_APPROVAL = 168;
+const WAIT_TIME_MULTI_APPROVAL = 24 * 2;
+const WAIT_TIME_SINGLE_APPROVAL = 24 * 7;
 
 const {
   REVIEW_SOURCES: { FROM_COMMENT, FROM_REVIEW_COMMENT }
@@ -49,12 +49,11 @@ class PRChecker {
     );
   }
 
-  checkAll(comments = false) {
+  checkAll(checkComments = false) {
     const status = [
-      this.checkReviews(comments),
       this.checkCommitsAfterReview(),
       this.checkCI(),
-      this.checkPRWait(new Date()),
+      this.checkReviewsAndWait(new Date(), checkComments),
       this.checkMergeableState(),
       this.checkPRState(),
       this.checkGitConfig()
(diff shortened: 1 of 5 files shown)


lib/pr_checker.js
```diff
@@ -132,12 +133,16 @@ class PRChecker {
    */
   getWait(now) {
     const createTime = new Date(this.pr.createdAt);
-    const timeLeft = WAIT_TIME - Math.ceil(
+    const hoursFromCreateTime = Math.ceil(
```
joyeecheung: This seems rather complicated. I believe we don't have to special case if we combine `checkReviews()` and `checkPRWait()` so that the number of approvals and wait times are checked together with conditionals instead of independently, the logic would be much clearer that way.

joyeecheung: (The two methods are separated only because previously a human would check these two conditions separately, but now there is a relationship between the two we should do what a human brain would do here instead and combine the logic.)

joyeecheung:
EDIT: Just realized I don't know how `fast-track` plays out if there is only one approvals..(also, should we check the thumbs up message? I don't even know how we can identify the reply that needs emojis programically...

lib/pr_checker.js
```diff
@@ -5,6 +5,7 @@ const MINUTE = SECOND * 60;
 const HOUR = MINUTE * 60;
 
 const WAIT_TIME = 48;
+const WAIT_TIME_SINGLE_APPROVAL = 168;
```
targos: Nit: `7 * 24` or add comment to make it  obvious that this is 7 days

targos:
> Also, should we check the thumbs up message?

I'd say no for now. I think we should take advantage of the bot for that:
- The person who would like to fast-track adds the label
- The bot posts a well-known message to ask for thumbs up
- `ncu` can find that message and count the number of thumbs up by collaborators (can it?)

joyeecheung:
> The bot posts a well-known message to ask for thumbs up

I don't think at the moment the bot has any hooks to do this after a label is added by a human?

targos:
The bot receives all pull request events, which apparently include labels:
image

Trott:
> cc @Trott Is that correct?

@joyeecheung Yes, and to answer this part:

```js
    if (fastTrack) {
        // ????
    }
   // check 7 day
```

No need to check `fastTrack` there. If you remove that check, it checks that the PR is open for 7 days. If you only have one approval, you cannot fast track. So, it's 7 days or else `return false`.

joyeecheung:
Ping @trivikr do you still want to pursue this? Ideally I want to go with a big branch as described in https://github.com/nodejs/node-core-utils/pull/290#issuecomment-427670704 so it's easier to update later if we are ever going the change the rules again

targos:
Ping :)

trivikr:
@joyeecheung @targos I wanted to make a simple change instead of refactoring the code.
I'm little busy, feel free to pick it up.

devsnek:
from this image:



it would be nice if  was on one line, and the `91` hours was made like `3 days 18 hours`

joyeecheung:
@devsnek Thanks for the feedback, I merged the two logs into oneline, regarding the date time improvements that should probably be don in a follow-on PR

trivikr:
Thanks @joyeecheung for picking up the refactor task!

PR #296, opened 2018-11-01
git-node: load supported WPT modules from file system

joyeecheung:
Load the supported tests from `test/wpt/status` foler for flexibility.
Also removes a redundant file.


joyeecheung:
cc @nodejs/automation

joyeecheung:
On how the status files work, see https://github.com/joyeecheung/node/blob/wpt-any/test/wpt/README.md

PR #299, opened 2018-11-04
[WIP] git-node: prompt and conditionally auto push

refack:
Seeking feedback and testing guidance.

Refs: https://github.com/nodejs/node/pull/23941#issuecomment-435678992

PR #300, opened 2018-11-04
cli: refactor out cli.promptForInput

joyeecheung:


joyeecheung:
cc @refack

joyeecheung:
Refs: https://github.com/nodejs/node-core-utils/pull/299

PR #301, opened 2018-11-09
cli: indent results, make warning and errors bold

joyeecheung:
Also removes the formatting in CLI tests since there isn't too much point to test the exact output


joyeecheung:
screen shot 2018-11-09 at 8 15 56 pm


The warning and errors are now less negligible

PR #302, opened 2018-11-09
deps: update package-lock.json

joyeecheung:


targos:
Can I suggest to instead:

```console
rm -rf node_modules package-lock.json
ncu -au
npm install
```

joyeecheung:
@targos You mean locally or regenerate the file with the command above?

targos:
I mean recreate the PR with this command. I do it like that on my projects and find it very effective.

BTW `ncu` comes from the package `npm-check-updates`. This step is not required. It upgrades all versions in the package.json

PR #306, opened 2018-11-11
deps: upgrade all dependencies

targos:
Alternative to https://github.com/nodejs/node-core-utils/pull/302

I verified `npm test` passes with Node 8 and 10.

`10 files changed, 1235 insertions(+), 2486 deletions(-)` 🎉


package.json
```diff
@@ -31,34 +31,34 @@
   ],
```
refack: IMO this should come with a version bump to this module 
https://github.com/nodejs/node-core-utils/blob/f033641787f3595e4fedb184744ec6f7e52ca15c/package.json#L3

targos: What do you mean? This module's version will be bumped when the next release is done

package.json
```diff
@@ -31,34 +31,34 @@
...
   },
   "devDependencies": {
-    "eslint": "^4.17.0",
-    "eslint-config-standard": "^11.0.0",
-    "eslint-plugin-import": "^2.8.0",
-    "eslint-plugin-node": "^6.0.1",
-    "eslint-plugin-promise": "^3.6.0",
-    "eslint-plugin-standard": "^3.0.1",
+    "eslint": "^5.9.0",
+    "eslint-config-standard": "^12.0.0",
```
joyeecheung: I actually start to think we should use something else (like the one ported from node core) instead..

targos: I agree. Using Standard with semicolons is... unusual 😄

Issue #303, opened 2018-11-09
`git node v8 backport` aborts on merge conflict


hashseed:
`git node v8 backport` attempts at some point to do a `git apply -3`, which fails often. The `-3` doesn't help since Node's copy of V8 doesn't have a git history.

How about not aborting, but wait for the user to fix the conflicts and `git add` the fixes and then continue?

targos:
+1000 if you know how to do that! I was not able to get conflict markers with `git apply` when I developed the command.

joyeecheung:
I think the thing to implement is more like to have a way to persist the failure and then be able to restart from here once the user fixes the conflict?

https://github.com/nodejs/node-core-utils/blob/092cbe5d997eb565e967eed5d9f44beb0f958dde/lib/update-v8/backport.js#L12

`git node land` is able to do that by persisting states to `.ncu/land` and always reading this file to understand the previous state so that users can `git node land --continue` all the way down while fixing errors

targos:
The real problem is that there is no conflict to fix. The patch fails and that's all. No files are modified

joyeecheung:
@targos One way to tackle that is to use `git apply --reject`, and prompt the user to use [wiggle](https://linux.die.net/man/1/wiggle) on the `.rej` files, though that cannot guarantee to work either (I guess it's better than outright failure?) Then when the conflicts are fixed, read some persisted state from disk to bake the commit message and so on.

hashseed:
I made an attempt to solve this: use `patch -p1 --merge --no-backup-if-mismatch` so that we get conflict markers in the files, and then prompt the user to continue once conflicts are resolved.

PR #304, opened 2018-11-11
v8: make backport command better

targos:
- Add support for incomplete SHA as input
- Add support for multiple SHAs to be combined

I'd like this to be merge quickly so I can work on a fix for https://github.com/nodejs/node-core-utils/issues/303 on top of it.

PR #307, opened 2018-11-13
git-node v8: prompt on merge conflict

hashseed:
Instead of performing `git apply` and aborting if it fails (which happens often), perform `patch -p1 --merge` instead to produce merge artifacts and then prompt for user input to continue.

It should look something like this:

example output

### DIFF
--- a/lib/update-v8/backport.js
+++ b/lib/update-v8/backport.js
@@ -5,6 +5,7 @@ const path = require('path');
 const execa = require('execa');
 const fs = require('fs-extra');
 const Listr = require('listr');
+const input = require('listr-input');
 
 const common = require('./common');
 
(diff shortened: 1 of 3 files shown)


hashseed:
This fixes https://github.com/nodejs/node-core-utils/issues/303

targos:
Nice, thank you! I tested it locally and it works :+1:

targos:
I will follow up with another PR that uses "backport" instead of "cherry-pick" in the commit message when a conflict occurs, but after https://github.com/nodejs/node-core-utils/pull/304 is landed.

hashseed:
I don't actually have power to merge this :)

targos:
I do ;)

Issue #313, opened 2018-11-22
Do we still need the appveyor config?


targos:
- We have Travis
- The badge reports errors
- The link errors: Project not found or access denied.

joyeecheung:
appveyor is supposed to test Windows, but I think we can probably try out the new Windows support of travis: https://blog.travis-ci.com/2018-10-11-windows-early-release

priyank-p:
Updated the title and removed the `stale` label since this is still relevant for GitHub Actions we currently use.

PR #314, opened 2018-11-22
travis: test on Windows

targos:
Fixes: https://github.com/nodejs/node-core-utils/issues/313


targos:
Let's see if it works

joyeecheung:
I think adding these to `.gitattributes` should fix the Windows issue?

```
# auto detect text files
* text=auto

# we don't have actual binary files in this repo
*.* text eol=lf
```

targos:
I changed the matrix and created `.gitattributes`: https://github.com/nodejs/node-core-utils/pull/314/checks

targos:
I cannot reproduce on my Windows machine

targos:
Rebased. @joyeecheung do you think we still need the `.gitattributes`?

targos:
There seems to still be issues with CRLF

joyeecheung:
Can you rebase and force push to see if Travis can pick the other fix up?

joyeecheung:
@targos ah, sorry, didn't see the history (was on mobile)

joyeecheung:
The previous green Windows I got is on travis-ci.org instead of travis-ci.com, I wonder if that makes a difference... https://travis-ci.org/joyeecheung/node-core-utils/jobs/462560512

Also, looks like the failing tests are all comparing output to some files contents

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Auto-detect text files
+* text=auto
```
joyeecheung: Can you try simply using

```
* text=auto eol=lf
```

in the whole file?

PR #325, opened 2019-02-07
session: don't call new upstream commits "stray"

sam-github:
Stray means to wander away from the right place or group, but upstream
commits are exactly where they should be, ready to be pulled. Use "new"
instead of "stray" to describe them.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -299,7 +299,8 @@ class Session {
     if (!stray.length) {
       return;
     }
-    cli.log(`Found new commits in ${branchName}:\n` +
+    cli.log(`${branch} is out of sync with ${branchName}. ` +
+            'Mismatched commits:\n' +
       ` - ${stray.join('\n - ')}`);
     const shouldReset = await cli.prompt(`Reset to ${branchName}?`);
     if (shouldReset) {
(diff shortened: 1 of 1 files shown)


lib/session.js
```diff
@@ -299,7 +299,7 @@ class Session {
     if (!stray.length) {
       return;
     }
-    cli.log(`Found stray commits in ${branchName}:\n` +
+    cli.log(`Found new commits in ${branchName}:\n` +
```
joyeecheung: This could also happen when the local branch contains commits not present in the upstream, I think `different` would be a more appropriate word?

```suggestion
    cli.log(`Found different commits in ${branchName}:\n` +
```

sam-github: "different" would confuse me, I'm not sure what they are different from. How about something like

> master doesn't match upstream/master

(subbing in the actual master branch name)

That seems both accurate and explicit.

targos:
Ping @joyeecheung

lib/session.js
joyeecheung: Maybe:

```suggestion
    cli.log(`${branch} is out of sync with ${branchName}. ` +
            'Mismatched commits:\n' +
```

would be clearer

joyeecheung:
Linter is happy, so I'll take it as it works..

PR #327, opened 2019-03-22
v8: fix reading DEPS file

targos:
Remove Python comments from the object before evaluating it.

Refs: https://github.com/v8/v8/blob/85017f042855d78ec6d87f00f75e12d966241f1b/DEPS#L112


refack:
It would be great to have a test that covers this.

PR #329, opened 2019-03-22
Tweak CI walk filter order

refack:
Moved more specific pattern higher up

### DIFF
--- a/lib/ci/ci_failure_parser.js
+++ b/lib/ci/ci_failure_parser.js
@@ -203,6 +203,39 @@ const FAILURE_FILTERS = [{
 }, {
   filter(ctx, text) {
     const patterns = [{
+      pattern: /FATAL: .+/g,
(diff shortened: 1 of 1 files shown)


lib/ci/ci_failure_parser.js
```diff
@@ -238,23 +236,25 @@ const FAILURE_FILTERS = [{
 }, {
   filter(ctx, text) {
     const patterns = [{
-      pattern: /FATAL: .+/g,
-      context: { index: -1, contextBefore: 0, contextAfter: 5 }
+      pattern: /There is not enough space in the file system./g,
```
joyeecheung: The ones moved here are actually more specific? Can you split those instead?

refack: Yeah I wrote this patch on GitHub, hoping I hit it in one try ¯\\_(ツ)_/¯
I need to dig into this locally a bit more...

BridgeAR:
Ping @refack

PR #331, opened 2019-03-29
v8: fix V8 upgrades from 7.4

targos:
- Just increment NODE_MODULE_VERSION automatically by one.
  Now that this constant can also be used by embedders, it is no longer
  predictable
- Add a --no-version-bump so we can skip this task for canary builds.
- From V8 7.4 on, gypfiles are located in tools/ instead of deps/v8.
  Stop trying to move them around.


lib/update-v8/gypfiles.js
```diff
@@ -5,7 +5,11 @@ const path = require('path');
 const fs = require('fs-extra');
 
 function nodeOwnsGypfiles(ctx) {
-  if (ctx.currentVersion[0] === 6 && ctx.currentVersion[1] < 6) {
+  if (
```
refack:
```suggestion
  const hasGypFiles = fs.exists(path.join(ctx.nodeDir, 'deps/v8/gypfiles'));
  if (!hasGypFiles) {
    return false;
  }
  const [v8_major, v8_minor, ...] = ctx.currentVersion;
  if (v8_major >= 7 || (v8_minor === 6 && v8_minor > 5)) {
    return true;
  }
  return false;
```

targos: I went for a more drastic change and removed this code entirely. We won't do major upgrades anymore on the current release lines so in the future gypfiles will always be in `tools`

refack: > removed this code entirely.

💚

PR #335, opened 2019-04-17
package: update nyc to 14.x

sam-github:
It's not a big deal, because these "vulnerabilities" are only in dev deps, but might as well update.

There is another "vulnerability" in js-yaml is via mocha, https://github.com/nodejs/security-wg/issues/520, but it has to be fixed there because mocha is pinning its dependency's versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "eslint-plugin-standard": "^4.0.0",
     "intelli-espower-loader": "^1.0.1",
     "mocha": "^6.0.2",
-    "nyc": "^13.3.0",
+    "nyc": "^14.0.0",
     "power-assert": "^1.6.1",
     "sinon": "^7.3.1"
   }


sam-github:
Not that when run with coverage 3 tests fail, but its the same 3 tests before or after updating nyc:

```
  1) auth
       asks for auth data if no ncurc is found:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/sam/w/core/node-core-utils/te
st/unit/auth.test.js)
      at listOnTimeout (internal/timers.js:535:17)
      at processTimers (internal/timers.js:479:7)

  2) auth
       returns ncurc data if valid in XDG_CONFIG_HOME:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/sam/w/core/node-core-utils/te
st/unit/auth.test.js)
      at listOnTimeout (internal/timers.js:535:17)
      at processTimers (internal/timers.js:479:7)

  3) auth
       prints an error message if it can't generate a token:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/sam/w/core/node-core-utils/te
st/unit/auth.test.js)
      at processTicksAndRejections (internal/process/task_queues.js:86:5)
```

PR #336, opened 2019-04-18
test: create Mocha config file; remove unneeded "exit" option

boneskull:
- Created `.mocharc.json` where current & future flags/options can be placed
- Removed the `exit` option.  Recommended practice is to only use `--exit` if a) your test process won't quit without it, and b) it's too difficult or time-consuming to fix your tests to cleanup properly.  Since the test process exits cleanly, I removed it.

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,0 +1,3 @@
+{
+  "require": "intelli-espower-loader"
+}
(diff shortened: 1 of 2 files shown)


priyank-p:
We should remove the AppVeyor CI config file `appveyor.yml` since we are no longer using it and add windows os on Travis CI instead for windows testing.

PR #345, opened 2019-05-13
lib: change order of "Post in" message

sam-github:
Change order to be slightly more consistent with previous message,
and so that the URL of the PR does not have to be followed directly
by a `:`. Some terminals consider the `:` to be part of the URL,
so clicking on it ends up on an invalid github page. Rearranging the
message so the URL is bounded by whitespace should work with all
terminals, and is no less readable.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -213,7 +213,7 @@ class LandingSession extends Session {
     cli.log('To finish landing:');
     cli.log(`1. Run \`git push ${upstream} ${branch}\``);
     const url = `https://github.com/${owner}/${repo}/pull/${prid}`;
-    cli.log(`2. Post \`Landed in ${willBeLanded}\` in ${url}`);
+    cli.log(`2. Post "Landed in ${willBeLanded}" in ${url}`);
 
     const shouldClean = await cli.prompt('Clean up generated temporary files?');
     if (shouldClean) {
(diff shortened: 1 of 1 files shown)


sam-github:
For me, a URL of [`https://github.com/nodejs/node-core-utils/pull/345:`](https://github.com/nodejs/node-core-utils/pull/345:) gets transformed by github into [`https://github.com/nodejs/node-core-utils/compare/345:?expand=1`](https://github.com/nodejs/node-core-utils/compare/345:?expand=1). Maybe its very specific to my terminal setup, but switching the order allows me to click on the link in the `git node land` message, and I don't think its any less readable than the original. This:

```
To finish landing:
1. Run `git push upstream master`
2. Post in https://github.com/nodejs/node/pull/27631: `Landed in 1df3080...b6bfc19`
```

after this PR becomes:

```
To finish landing:
1. Run `git push upstream master`
2. Post `Landed in 1df3080...b6bfc19` in https://github.com/nodejs/node/pull/27631
```

lib/landing_session.js
```diff
@@ -213,7 +213,7 @@ class LandingSession extends Session {
     cli.log('To finish landing:');
     cli.log(`1. Run \`git push ${upstream} ${branch}\``);
     const url = `https://github.com/${owner}/${repo}/pull/${prid}`;
-    cli.log(`2. Post in ${url}: \`Landed in ${willBeLanded}\``);
+    cli.log(`2. Post \`Landed in ${willBeLanded}\` in ${url}`);
```
targos:
```suggestion
    cli.log(`2. Post "Landed in ${willBeLanded}" in ${url}`);
```

While you're here. This isn't code.

PR #348, opened 2019-05-29
test: test collaborator names with parentheses

joyeecheung:
张三 is roughly the Chinese version of [John Doe](https://en.wikipedia.org/wiki/John_Doe) (modulo it is not used for corpses)


ChALkeR:
Is there an ncu test with rtl language text somewhere?

priyank-p:
> Is there an ncu test with rtl language text somewhere?

As far as I can remember, there are no tests for rtl text.

PR #349, opened 2019-06-11
git-node: add a prompt for force landing commits with invalid message

joyeecheung:
Makes it possible to land a commit that does not pass
the core-validate-commit checks in case there is a bug
in the validator.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -196,7 +196,23 @@ class LandingSession extends Session {
       __dirname,
       '../node_modules/.bin/core-validate-commit' + (isWindows ? '.cmd' : '')
     );
-    await runAsync(validateCommand, stray);
+
+    try {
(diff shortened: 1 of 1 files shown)


richardlau:
If this is related to https://github.com/nodejs/core-validate-commit/pull/62 I'd like some consideration as to whether that really is a bug in the validator or that the V8 commits have been formatted incorrectly all along.

joyeecheung:
@richardlau Partially related to https://github.com/nodejs/core-validate-commit/pull/62 but this has been on my TODO list for a while since I've encountered bugs in core-validate-commit several times in the past. IMO bugs like that should not block the whole landing of a patch so it is nice to have a way to keep landing a commit using the tool - even without this the user can just ignore the warning and `git push` so it's just formality.

Issue #344, opened 2019-05-01
ncu-ci stopped working on Windows jobs


joyeecheung:
The results are reported as the fallback `Failed in Tests phase` becuase ncu-ci does not know how to parse the results properly. There's probably some upstream changes in the JSON respose (the `-2` suffix?) that are not yet handled by ncu-ci.

```
-------------------------- node-test-binary-windows-2 --------------------------
URL        job/node-test-binary-windows-2/521/
Type       BUILD_FAILURE
Reason
  Failed in Tests phase (node-test-binary-windows-2)
```

The job looks like this:

https://ci.nodejs.org/job/node-test-binary-windows-2/521/

<img width="671" alt="Screen Shot 2019-05-01 at 5 08 39 PM" src="https://user-images.githubusercontent.com/4299420/57011338-ca057d00-6c33-11e9-87bf-c6bd1bc3a203.png">

Theoratically we could also use the URL https://ci.nodejs.org/job/node-test-binary-windows-2/521/testReport/api/json to get results but sometimes the TAP report of Windows job is off
(https://ci.nodejs.org/job/node-test-binary-windows-2/512/)

<img width="263" alt="Screen Shot 2019-05-01 at 5 13 19 PM" src="https://user-images.githubusercontent.com/4299420/57011481-70518280-6c34-11e9-826a-2437c2afda6e.png">

PR #351, opened 2019-06-12
ncu-ci: parse console text of the first failed phase

joyeecheung:
Previously we try to detect the type of the failed phase
and return a generic build failure when the job failed in a
non-test phase. This is not really necessary - the console
text parser is capable of detecting the actual build failure
from even the console text of the build phase. Also hard-coding
the test phase name makes the detection fragile when there
are configuration changes in the Jenkins. So we should just
skip the detection.

Fixes: https://github.com/nodejs/node-core-utils/issues/344

### DIFF
--- a/lib/ci/ci_result_parser.js
+++ b/lib/ci/ci_result_parser.js
@@ -37,7 +37,6 @@ const FAILURE = 'FAILURE';
 const ABORTED = 'ABORTED';
 const UNSTABLE = 'UNSTABLE';
 
-const TEST_PHASE = 'Binary Tests';
 // com.tikal.jenkins.plugins.multijob.MultiJobBuild
 const BUILD_FIELDS = 'builtOn,buildNumber,jobName,result,url';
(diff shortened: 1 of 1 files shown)


joyeecheung:
cc @nodejs/automation

joyeecheung:
@evenstensberg What do you mean by json side of things?

evenstensberg:
I’m just trying to understand the source of truth. New to the automation team so trying to get insight

joyeecheung:
@evenstensberg The source of truth is Jenkins. This tool adapts to the configurations there.

evenstensberg:
Okay so can't change output instead of removing the test? 

CC @mhdawson for reviews

joyeecheung:
@evenstensberg There isn't anything actually removed here? It simply removes an false early return and always run the logic below no matter how the phases are named.

evenstensberg:
Okay gotcha, thanks for elaborating!

Issue #353, opened 2019-06-13
detection of lite ci is broken again?


sam-github:
I recall this was broken, then @richardlau fixed it, but now it seems to be broken again, I get `No ci runs detected` on https://github.com/nodejs/node/pull/28178 and https://github.com/nodejs/node/pull/28162

```
core/node (master $% u=) % node --version
v12.4.0
core/node (master $% u=) % git node --version
1.17.1
```

(just updated)

Am I(we?) doing something wrong?

richardlau:
Currently testing out with https://github.com/nodejs/node/pull/28178 but it looks like it currently needs `/pipeline` at the end of the commented url: https://github.com/nodejs/node-core-utils/blob/8d6455327069abf5d366809177d9625c268946d3/lib/ci/ci_type_parser.js#L86-L91

I think when the bot was able to autostart jobs it would post the lite ci's with `/pipeline` at the end. In any case we should probably try to make the regexp not require the `/pipeline` suffix.

richardlau:
```console
-bash-4.2$ node ../node-core-utils/bin/git-node land 28162
✔  Done loading data for nodejs/node/pull/28162
----------------------------------- PR info ------------------------------------
Title      doc: update kernel and glibc reqs for PPCle (#28162)
Author     Michael Dawson <michael_dawson@ca.ibm.com> (@mhdawson)
Branch     mhdawson:ppcversions -> nodejs:master
Labels     author ready, build, doc
Commits    3
 - doc: update kernel and glibc reqs for PPCle
 - squash: address comments
 - squash: also update toolchains section.
Committers 2
 - Michael Dawson <michael_dawson@ca.ibm.com>
 - GitHub <noreply@github.com>
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/28162
Reviewed-By: Richard Lau <riclau@uk.ibm.com>
Reviewed-By: Sam Roberts <vieuxtech@gmail.com>
Reviewed-By: Rich Trott <rtrott@gmail.com>
Reviewed-By: Beth Griggs <Bethany.Griggs@uk.ibm.com>
--------------------------------------------------------------------------------
   ℹ  Last Lite PR Pipeline CI on 2019-06-13T15:11:00Z: https://ci.nodejs.org/job/node-test-pull-request-lite-pipeline/3678/
   ℹ  This PR was created on Mon, 10 Jun 2019 23:28:35 GMT
   ✔  Approvals: 4
[...]

PR #354, opened 2019-06-13
ci-parser: fix detection of lite CI pipelines

richardlau:
Fixes: https://github.com/nodejs/node-core-utils/issues/353

cc @sam-github

### DIFF
--- a/lib/ci/ci_type_parser.js
+++ b/lib/ci/ci_type_parser.js
@@ -86,7 +86,7 @@ const CI_TYPES = new Map([
   [LITE_PR_PIPELINE, {
     name: 'Lite PR Pipeline',
     jobName: 'node-test-pull-request-lite-pipeline',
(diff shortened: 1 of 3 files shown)


richardlau:
> 
> 
> RSLGTM, though will it match with or with the trailing pipeline? Should both cases be tested? I don't know, just asking, and thanks for fixing this.

There's still a test that has the `/pipeline` suffix that I left alone (and still passes): https://github.com/nodejs/node-core-utils/blob/433293aefc38d72ab64832f137321be0bcafea36/test/unit/pr_checker.test.js#L574

sam-github:
I'm not sure what the policy is here, if it is like node core, but I'm in support of fast-tracking and publishing this fix.

joyeecheung:
@sam-github The policy is basically "this can land whenever it gets an approval from someone who can write to this repo".

PR #364, opened 2019-10-01
ncu-config: introduce project config file

mmarchini:
Project config file is intended to be committed to a project. This will
allow other projects in the org to use NCU without users having to set
`owner` and `repo` by themselves. In the future we can also use this add
other project-specific configurations (number of hours to wait, minimum
approvals and main CI, for example).

### DIFF
--- a/bin/ncu-config
+++ b/bin/ncu-config
@@ -3,7 +3,7 @@
 'use strict';
 
 const {
-  getConfig, updateConfig
+  getConfig, updateConfig, GLOBAL_CONFIG, PROJECT_CONFIG, LOCAL_CONFIG
 } = require('../lib/config');
 
 const yargs = require('yargs');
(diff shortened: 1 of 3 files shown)


priyank-p:
Shouldn't we just rather look for `~/.ncurc` by default without having to pass in `-g` or `--global` flag? and use that if it exists. And not look for it if `-p` is passed and rather use `./.ncurc`.

(I don't use the tool so can't tell for sure that this is indeed better implementation; this is just a idea from looking at the code)

mmarchini:
Today `ncu-config` will look at `$PWD/.ncu/config` (known as local configs in the source code) if no flags are passed, I didn't want to change this behavior. Also, `ncu-config` is used only for config manipulation, so it makes sense for the user to explicitly say which config file they want to change/look. *Maybe* it would make sense for `ncu-config list` to show configs from all files (like `git config` does), but I don't think many users use `list` to make a difference.

All other tools rely on `getMergedConfigs`, which will fetch configs from all three locations and merge them, prioritizing local config, then project config, then global config.

joyeecheung:
Why do we need arguments for these? I think we can simply let `./.ncu/config` override `./.ncurc` which overrides `~/.ncurc`?

mmarchini:
> Why do we need arguments for these?

The arguments are only for `ncu-config`, which is how it worked before (--global to modify `~/.ncurc`, no arguments to modify `./.ncu/config`). Without the arguments we wouldn't be able to set values (`ncu-config set`) for a specific file.

> I think we can simply let `./.ncu/config` override `./.ncurc` which overrides `~/.ncurc`?

That's the behavior for `git-node` and other commands, no extra arguments needed.

PR #366, opened 2019-10-01
land: add non-interactive mode

mmarchini:
Non-interactive mode will try to land a Pull Request without asking any
questions. This is meant to be used by scripts and CI landing tools.

### DIFF
--- a/components/git/land.js
+++ b/components/git/land.js
@@ -30,6 +30,13 @@ const landOptions = {
   abort: {
     describe: 'Abort the current landing session',
     type: 'boolean'
+  },
+  yes: {
+    type: 'boolean',
+    default: false,
+    describe: 'Assume "yes" as answer to all prompts and run ' +
+    'non-interactively. If an undesirable situation occurs, such as a pull ' +
+    'request or commit check fails, then git node land will abort.'
   }
 };
 
@@ -93,6 +100,9 @@ function handler(argv) {
 
 function land(state, argv) {
   const cli = new CLI(process.stderr);
+  if (argv.yes) {
+    cli.setAssumeYes();
+  }
   const req = new Request();
   const dir = process.cwd();
 
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -22,10 +22,14 @@ class LandingSession extends Session {
     const { cli } = this;
     this.startLanding();
     const status = metadata.status ? 'should be ready' : 'is not ready';
+    // NOTE(mmarchini): default answer is yes. If --yes is given, we need to be
+    // more careful though, and we change the default to the result of our
+    // metadata check.
+    const defaultAnswer = !cli.assumeYes ? true : metadata.status;
     const shouldContinue = await cli.prompt(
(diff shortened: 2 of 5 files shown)


lib/landing_session.js
```diff
@@ -20,7 +20,7 @@ class LandingSession extends Session {
     this.startLanding();
     const status = metadata.status ? 'should be ready' : 'is not ready';
     const shouldContinue = await cli.prompt(
-      `This PR ${status} to land, do you want to continue?`);
+      `This PR ${status} to land, do you want to continue?`, metadata.status);
```
mmarchini: I wasn't sure about this. Do we want to change the default for this prompt? We *could* only change it if the command is running on non-interactive mode. On the other hand, it kinda makes sense for this prompt to default to false.

mmarchini:
I'll try to add some tests.

components/git/land.js
```diff
@@ -38,6 +38,8 @@ function builder(yargs) {
     .options(landOptions).positional('prid', {
       describe: 'ID or URL of the Pull Request'
     })
+    .boolean('non-interactive')
```
joyeecheung: I'd actually prefer `--yes`/`-y` here...I think that's a more universal option name?

lib/landing_session.js
joyeecheung: If it's --yes, then it makes more sense to make set this to true in that mode.

components/git/land.js
mmarchini: Humm, that makes sense. I got `non-interactive` from debconf's non-interactive mode, but `--yes/-y` seems to be used by more tools. 

But assuming `yes` to all questions is not the behavior I was looking for. The idea is to run all actions from `git node land` without asking on the best case scenario (all validations passed, only one commit, etc.). So, for example, if CI failed, it shouldn't continue.

We could still use `--yes` for that though.

lib/landing_session.js
mmarchini: In that case it would also make sense to change the default of the last check (which uses `core-validate-commit`) to true. I'm not sure we want that behavior though.

components/git/land.js
joyeecheung: Can we add some kind of config file for the behavior of `--yes`? That would also make it easier to tweak stuff in bots I suppose.

joyeecheung: Maybe `--silent` is also an appropriate name, and the config argument could be specified by something like `--silent-answers`.

lib/landing_session.js
mmarchini: I kept this with the previous behavior (default to Y) unless --yes is given, in which case it will default to the result of the metadata check.

mmarchini:
@targos done!

mmarchini:
CI is passing :tada: 

@targos do you mind landing this? I don't have commit access here.

priyank-p:
@mmarchini merged!

targos:
released in 1.19.0

PR #367, opened 2019-10-01
git-node: add config option to set wait times

mmarchini:
Add two new config options: `waitTimeSingleApproval` and
`waitTimeMultiApproval`, which can be used to set how long ncu should
wait before landing a PR with 1 approval or with 2+ approvals.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -50,6 +50,8 @@ class Session {
       upstream: this.upstream,
       branch: this.branch,
       readme: this.readme,
+      waitTimeSingleApproval: this.waitTimeSingleApproval,
+      waitTimeMultiApproval: this.waitTimeMultiApproval,
       prid: this.prid
     };
   }
@@ -76,6 +78,14 @@ class Session {
 
   get readme() {
     return this.config.readme;
+  }
+
+  get waitTimeSingleApproval() {
+    return this.config.waitTimeSingleApproval;
+  }
+
+  get waitTimeMultiApproval() {
+    return this.config.waitTimeMultiApproval;
   }
 
   get pullName() {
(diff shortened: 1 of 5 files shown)


joyeecheung:
Can you add some docs? From the look of the code I assume it's in hours?

mmarchini:
Sure. What's the best place to document this? [git-node.md](https://github.com/nodejs/node-core-utils/blob/master/docs/git-node.md) or [ncu-config.md](https://github.com/nodejs/node-core-utils/blob/master/docs/ncu-config.md)?

joyeecheung:
It's git-node specific config so I think it should be documented there.

mmarchini:
Done

lib/session.js
```diff
@@ -78,6 +80,14 @@ class Session {
     return this.config.readme;
   }
 
+  get waitTimeSingleApproval() {
+    return this.config.waitTimeSingleApproval;
```
joyeecheung: Umm, on a second thought, why are the defaults set in the PR checker, instead of here?

mmarchini: Ah yes, that makes more sense. I'll move it here.

mmarchini: On second thought, this won't work because `git node metadata` doesn't use a Session. So `git node land` would get the defaults correctly, but `git node metadata` wouldn't.

We could create a MetadataSession, but it seems a bit too much. Also doesn't make sense since the idea of a session (to my understanding at least) is to save the state of a command to be resumed later.

Should we keep this here then?

mmarchini: ping @joyeecheung

joyeecheung: Ah, makes sense, I forgot that the PR checker is also run by the metadata generator.

mmarchini:
@joyeecheung do you mind landing this? I don't have commit access to this repo

PR #369, opened 2019-10-02
git-node: add GitHub status as a CI option

mmarchini:
Some project in the org don't use Jenkins, which means PRChecker will
never succeed for pull requests on those projects. These projects
usually have Travis, AppVeyor or other CI systems in place, and those
systems will publish the status to GitHub, which can be retrieved via
API. This commit adds GitHub status as an optional way to validate if a
PR satisfies the CI requirement.

We need to check for the CI status in two fields returned by our GraphQL
query: commit.status for services using the old GitHub integration, and
commits.checkSuites for services using the new GitHub integration via
GitHub Apps.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -55,6 +55,7 @@ class Session {
       readme: this.readme,
       waitTimeSingleApproval: this.waitTimeSingleApproval,
       waitTimeMultiApproval: this.waitTimeMultiApproval,
+      ciType: this.ciType,
       prid: this.prid
     };
   }
@@ -89,6 +90,10 @@ class Session {
 
   get waitTimeMultiApproval() {
     return this.config.waitTimeMultiApproval;
+  }
+
+  get ciType() {
+    return this.config.ciType || 'jenkins';
   }
 
   get pullName() {
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -201,9 +201,20 @@ class PRChecker {
     return cis.find(ci => isFullCI(ci) || isLiteCI(ci));
   }
 
+  checkCI() {
+    const ciType = this.argv.ciType || 'jenkins';
+    if (ciType === 'jenkins') {
+      return this.checkJenkinsCI();
+    } else if (ciType === 'github-check') {
+      return this.checkGitHubCI();
+    }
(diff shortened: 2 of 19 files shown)


mmarchini:
I'll add docs later, since they'll go to the same place as the docs in #367.

mmarchini:
Oh no, I thought eslint would fix those lines, but it can't autofix max-len :(

I'll fix it manually and update the PR.

sam-github:
Even though node uses jenkins, the jenkins results are pushed to github, so this approach might work for node as well.

mmarchini:
> Even though node uses jenkins, the jenkins results are pushed to github, so this approach might work for node as well.

Yes, but it's less reliable since sometimes the bot won't update the status correctly.

mmarchini:
@joyeecheung do you mind removing the WIP tag?

joyeecheung:
@mmarchini Can you rebase this branch? The Jenkins status in the Node.js repo are fairly stable now, so it would be good to use this for Node as well. Thanks!

mmarchini:
On my list. Will try to get this done today or next week.

mmarchini:
Rebased

lib/session.js
```diff
@@ -91,6 +92,10 @@ class Session {
     return this.config.waitTimeMultiApproval;
   }
 
+  get ciType() {
+    return this.config.ciType || 'jenkins';
```
joyeecheung: I think we can make `github-check` the default now?

mmarchini: Not yet. If we set `github-check` as the default, ncu will not warn users that Jenkins jobs didn't ran on a `nodejs/node` Pull Request, since Travis and GitHub Actions also publish to GitHub Check.

joyeecheung: I guess we'll need to add extra logic for Travis v.s. Jenkins status updates, then. Until then it always warns unless you do `ncu-config set ciType github-check` since the bot (or people) no longer posts that comment to the PRs

joyeecheung:
Thanks for the PR :)

lib/session.js
mmarchini: > since the bot (or people) no longer posts that comment to the PRs

Which comment? The bot is still commenting when a CI starts.

PR #372, opened 2019-10-24
fix: use 8 characters for commit SHAs

targos:
GitHub already refused to generate a link with 7.

### DIFF
--- a/lib/ci/ci_result_parser.js
+++ b/lib/ci/ci_result_parser.js
@@ -26,7 +26,8 @@ const {
   CI_TYPES
 } = require('./ci_type_parser');
 const {
-  flatten
+  flatten,
+  shortSha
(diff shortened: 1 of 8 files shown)


targos:
See https://github.com/nodejs/node/pull/30064#issuecomment-545937837

This link works: https://github.com/nodejs/node/commit/d53dd8b0
This link doesn't: https://github.com/nodejs/node/commit/d53dd8b

addaleax:
Fwiw, I’ve set this to 12 locally, and I’m happy with that. But since most messages here either aren’t user-facing or are supposed to end up on Github where they will be automatically abbreviated anyway, can we maybe go for something even longer or the full SHA?

targos:
I'm fine with any length

PR #379, opened 2019-12-09
Add functionality for daily-master to NCU-CI

AshCripps:
@sam-github heres my draft for what I have so far.

### DIFF
--- a/lib/ci/ci_result_parser.js
+++ b/lib/ci/ci_result_parser.js
@@ -225,6 +225,13 @@ class TestBuild extends Job {
     this.builtOn = builtOn;
   }
 
+  setDailyBuildData({ result, changeSet, actions, timestamp, builtOn }) {
+    this.change = changeSet.items[0] || {};
+    this.date = new Date(timestamp);
+    this.result = result;
+    this.builtOn = builtOn;
+  }
+
   get sourceURL() {
     const { params } = this;
 
@@ -303,7 +310,9 @@ class TestBuild extends Job {
   displayBuilds() {
     const { cli, failures, builds } = this;
     for (const failure of failures) {
-      this.displayFailure(failure);
+      if (failure !== undefined) {
+        this.displayFailure(failure);
+      }
     }
     cli.separator('Other builds');
     for (const aborted of builds.aborted) {
@@ -329,6 +338,7 @@ class TestBuild extends Job {
     const { failures } = this;
     let output = `Failures in job ${this.jobUrl}\n\n`;
     for (const failure of failures) {
+      if (failure === undefined) continue;
       output += `#### [${getNodeName(failure.url)}](${failure.url})`;
       if (!failure.reason.includes('\n') && failure.reason.length < 20) {
         const builtOn = failure.builtOn ? `On ${failure.builtOn}: ` : '';
@@ -355,6 +365,10 @@ class TestBuild extends Job {
 }
 
 function getHighlight(f) {
+  if (!f.reason) {
+    f.reason = 'failure not found';
+    return f.reason;
+  }
   return f.reason.split('\n')[f.highlight]
     .replace(/not ok \d+ /, '')
     .replace(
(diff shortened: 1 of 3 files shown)


sam-github:
For context: we're looking to get ncu-ci running against daily builds, because, by definition, any failure on master is either a failure in the infrastructure, or a flaky test.

sam-github:
We're trying to figure out if we should push it forward because its useful and achievable, or perhaps whether its a bit of an improvement already and can be merged, or perhaps it was just an interesting experiment and we should abandon it.

@joyeecheung what do you think?

joyeecheung:
@sam-github Is there a summary of what this does? AFAICT it just adds some logic so that the command can be used on daily build data and it adds some logic to make it more fail-safe?

sam-github:
The intention was just this,

> adds some logic so that the command can be used on daily build data

but maybe some other logic was added as needed.

joyeecheung:
@sam-github I think this should be good to get in with the commented code removed

AshCripps:
ive rebased and removed the commented code.

lib/ci/ci_result_parser.js
```diff
@@ -343,24 +354,26 @@ class TestBuild extends Job {
...
+      .replace(
+        /failed: no workspace for .+/,
+        'failed: no workspace for ...'
+      )
+      .replace(
+        /fatal: loose object \w+ \(stored in .git\/objects\/.+\) is corrupt/,
+        'fatal: loose object ... (stored in .git/objects/...) is corrupt')
+      .replace(/hudson\.plugins\.git\.GitException: /, '')
+      .replace(/java\.io\.IOException: /, '');
+  } else return f.reason = "failure not found";
```
sam-github, line 376: you could reduce the diff, and simplify the logic, by leaving the code as it was and returning early when the exceptional condition is found:

```
if (!f.reason)
   return f.reason = 'failure not found'
```

lib/ci/ci_result_parser.js
```diff
@@ -804,7 +839,9 @@ class FannedBuild extends Job {
     const { jobName, buildNumber } = failedPhase;
     const build = new NormalBuild(cli, request, jobName, buildNumber);
     let failures = await build.getResults();
-    failures = flatten(failures);
+    if (failures != undefined){
```
sam-github, line 842: Would it be better to change .getResults(), so that it always returns a results array, even if its empty?

lib/ci/ci_result_parser.js
```diff
@@ -873,48 +910,52 @@ class NormalBuild extends Job {
     }
 
     const { result, runs, builtOn } = data;
+    //console.log(data);
```
sam-github, line 913: remove commented code, or if ncu uses `debug()`, change to a debug statement

lib/ci/ci_result_parser.js
```diff
@@ -873,48 +910,52 @@ class NormalBuild extends Job {
...
 
-    if (!runs.length) {
-      this.failures = [
-        new BuildFailure(
-          { url: this.jobUrl, builtOn }, 'Failed to trigger runs'
-        )
-      ];
-      return this.failures;
-    }
+    if (runs != undefined){
```
sam-github, line 921: this diff is quite hard to interpret, but I think that what is happening is that you have wrapped an existing group of statements into an `if(cond) {unusual}else{prexisting}` block. Better use the return-early idiom: `if(cond) {unusual; return}`. Its not just the diff, it reduces cognitive load - it means from that point on, the reader knows for sure that the unusual "cond" condition is NOT true. With the if/else, you have to remember it is still true, and that code later on after the else condition may still have to deal with it.

lib/ci/ci_result_parser.js
```diff
@@ -320,14 +329,16 @@ class TestBuild extends Job {
...
     for (const failure of failures) {
-      output += `#### [${getNodeName(failure.url)}](${failure.url})`;
-      if (!failure.reason.includes('\n') && failure.reason.length < 20) {
-        const builtOn = failure.builtOn ? `On ${failure.builtOn}: ` : '';
-        output += `\n\n${builtOn}${failure.reason}\n`;
-      } else {
-        output += '\n\n';
-        const builtOn = failure.builtOn ? ` on ${failure.builtOn}:` : '';
-        output += fold(`See failures${builtOn}`, failure.reason) + '\n\n';
+      if (failure !== undefined) {
```
sam-github, line 332: variation of early-return, early-continue:
```
if (failure === undefined)
  continue
```

lib/ci/ci_result_parser.js
```diff
@@ -880,6 +919,8 @@ class NormalBuild extends Job {
       return this.failures;
     }
 
+    if (!runs) return;
```
sam-github, line 922: This is returning `undefined` (effectively), so now the caller has to be able to deal with that, maybe better to return `[]`? Functions that return an array-or-undefined require callers to check. Returning an array always simplifies things, and allows only callers that _really_ care to check if `retval.length === 0`. If the function really needs to distinguish error from zero-length data, that'd be different, but that is probably not the case here.

AshCripps:
Can this be landed? I dont have permission to merge.

richardlau:
> 
> 
> Can this be landed? I dont have permission to merge.

cc @nodejs/automation-collaborators

AshCripps:
Ive rebased this against master so is free to be landed again

PR #380, opened 2019-12-12
v8: ignore zlib test and bench directories

targos:
They are not needed to build V8 and cause issues with our release CI.

I'm intending to fix this issue:
image


targos:
@nodejs/automation-collaborators

PR #381, opened 2019-12-13
chore: fix ESLint errors

targos:


targos:
They were introduced in https://github.com/nodejs/node-core-utils/pull/367 because it landed after https://github.com/nodejs/node-core-utils/pull/371 but the Travis run happened before that.

targos:
@nodejs/automation-collaborators

PR #383, opened 2020-01-23
git-node: add --backport flag to land

andrewhughes101:
This commit adds functionality to land that allows you to land backports
with the correct metadata

example: `git node land --backport <PR number>`

Co-authored-by: @AshCripps <ashley.cripps@ibm.com>

### DIFF
--- a/components/git/land.js
+++ b/components/git/land.js
@@ -37,6 +37,11 @@ const landOptions = {
     describe: 'Assume "yes" as answer to all prompts and run ' +
     'non-interactively. If an undesirable situation occurs, such as a pull ' +
     'request or commit check fails, then git node land will abort.'
+  },
+  backport: {
+    describe: 'Land a backport PR onto a staging branch',
+    default: false,
+    type: 'boolean'
   }
 };
 
@@ -152,8 +157,14 @@ async function main(state, argv, cli, req, dir) {
       cli.log('run `git node land --abort` before starting a new session');
       return;
     }
-    session = new LandingSession(cli, req, dir, argv.prid);
+    session = new LandingSession(cli, req, dir, argv.prid, argv.backport);
     const metadata = await getMetadata(session.argv, cli);
+    if (argv.backport) {
+      const split = metadata.metadata.split('\n')[0];
+      if (split === 'PR-URL: ') {
+        cli.error('Commit message is missing PR-URL');
+      }
+    }
     return session.start(metadata);
   } else if (state === APPLY) {
     return session.apply();
--- a/lib/links.js
+++ b/lib/links.js
@@ -4,6 +4,7 @@ const FIXES_RE = /(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/mgi;
 const FIX_RE = /(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/i;
 const REFS_RE = /Refs?\s*:\s*(\S+)/mgi;
 const REF_RE = /Refs?\s*:\s*(\S+)/i;
+const PR_RE = /PR-URL\s*:\s*(\S+)/i;
 const cheerio = require('cheerio');
 
 /**
@@ -36,7 +37,19 @@ class LinkParser {
       const m = item.match(REF_RE);
       if (!m) continue;
       const ref = m[1];
-      const url = this.getRefUrlFromOP(ref);
+      const url = this.getUrlFromOP(ref);
+      if (url) result.push(url);
+    }
+    return result;
+  }
+
+  getPRUrlsFromArray(arr) {
+    const result = [];
+    for (const item of arr) {
+      const m = item.match(PR_RE);
+      if (!m) continue;
+      const prUrl = m[1];
+      const url = this.getUrlFromOP(prUrl);
       if (url) result.push(url);
     }
     return result;
@@ -44,7 +57,7 @@ class LinkParser {
 
   // Do this so we can reliably get the correct url.
   // Otherwise, the number could reference a PR or an issue.
-  getRefUrlFromOP(ref) {
+  getUrlFromOP(ref) {
     const as = this.$('a');
     const links = as.map((i, el) => this.$(el)).get();
     for (const link of links) {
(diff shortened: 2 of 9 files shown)


sam-github:
@andrewhughes101 I labelled WIP because I think you intend to extend the tests now that the current ones are passing.

components/git/land.js
```diff
@@ -37,6 +37,11 @@ const landOptions = {
     describe: 'Assume "yes" as answer to all prompts and run ' +
     'non-interactively. If an undesirable situation occurs, such as a pull ' +
     'request or commit check fails, then git node land will abort.'
+  },
+  backport: {
```
sam-github: Not blocking, but is it possible to set the default value of this correctly, without the caller having to use `--backport`? It seems to me that it is known to be a backport in most cases because either the target is not master, or the `[backport ]` metadata in the PR subject. Not a big deal, not part of the core functionality, and easy to add later if its useful.

test/fixtures/pr_with_backport.json
```diff
@@ -0,0 +1,22 @@
...
+    "nodes": [
+      {
+        "name": "build"
+      },
+      {
+        "name": "v12.x"
+      }
+    ]
+  }
+}
```
sam-github: set your editor to end all lines with a NL

components/git/land.js
```diff
@@ -37,6 +37,11 @@ const landOptions = {
...
+    describe: '',
```
sam-github: Needs a description, and likely needs some docs added to the README.md, or wherever the docs are.

components/git/land.js
```diff
@@ -152,8 +157,14 @@ async function main(state, argv, cli, req, dir) {
       cli.log('run `git node land --abort` before starting a new session');
       return;
     }
-    session = new LandingSession(cli, req, dir, argv.prid);
+    session = new LandingSession(cli, req, dir, argv.prid, argv.backport);
     const metadata = await getMetadata(session.argv, cli);
+    if (argv.backport) {
+      const split = metadata.metadata.split('\n')[0];
+      if (split === 'PR-URL: ') {
+        cli.error('Commit message is missing PR-URL');
```
sam-github: I don't know if this covers all cases, but it seems to me that any PR that has a commit message with `PR-URL: ` already in it is a backport by definition, so doesn't need the argv switch. Or do we have other cases?

lib/links.js
```diff
@@ -4,6 +4,7 @@ const FIXES_RE = /(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/mgi;
 const FIX_RE = /(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/i;
 const REFS_RE = /Refs?\s*:\s*(\S+)/mgi;
 const REF_RE = /Refs?\s*:\s*(\S+)/i;
+const PR_RE = /PR-URL?\s*:\s*(\S+)/i;
```
sam-github: the `?` is a copy-paste error, in the lines above `Ref` or `Refs` are both valid, but `PR-UR:`  without the `L` is not.

lib/links.js
```diff
@@ -73,6 +86,15 @@ class LinkParser {
     }
     return [];
   }
+
+  getAltPrUrl() {
+    const text = this.$.text();
+    const refs = text.match(PR_RE);
```
sam-github:
```suggestion
    const refs = text.match(PR_RE) || [];
```

would allow removing the conditional and the multireturn

lib/metadata_gen.js
```diff
@@ -31,15 +32,24 @@ class MetadataGenerator {
     const parser = new LinkParser(owner, repo, op);
     const fixes = parser.getFixes();
     const refs = parser.getRefs();
-
+    const altUrl = parser.getAltPrUrl();
```
sam-github: I find it confusing when the same thing has slightly different names, `altUrl` is not obviously the result of `getAltPrUrl()`, but it would be if it was named `altPrUrl`.

components/git/land.js
BethGriggs: We often have release line-specific PRs where there may not be an equivalent commit in master, so we cannot assume it from the branch. Also, only members of backporters/releasers should be landing backports - so I think it will be clearer to keep the option separate/explicit (at least for now).

sam-github: sounds reasonable.

components/git/land.js
sam-github: there are other cases

BethGriggs:
Mostly LGTM. I've pulled down the code and tried pulling in a few backports and it works as expected 😃. The only one I had issues with was https://github.com/nodejs/node/pull/31431 - it states that it cannot detect PR-URL (but the PR-URL is in each of the commits). Is the issue because the tool is looking for "PR-URL:" in the PR body rather than in the individual commit messages?

AshCripps:
> 
> 
> Mostly LGTM. I've pulled down the code and tried pulling in a few backports and it works as expected 😃. The only one I had issues with was [nodejs/node#31431](https://github.com/nodejs/node/pull/31431) - it states that it cannot detect PR-URL (but the PR-URL is in each of the commits). Is the issue because the tool is looking for "PR-URL:" in the PR body rather than in the individual commit messages?

Yeah the tool itself doesn't have that functionality that we could see at least, it only goes through the data inside the PR message itself

MylesBorins:
Some of the commits in that PR had corrupt meta data... specifically they had `PR-URL: #number-of-pr` instead of the actual URL

andrewhughes101:
I've updated the PR to always place the Backport-PR-URL below the PR-URL in the metadata. I've also created some asciinema of this feature in action

Single commit: https://asciinema.org/a/300758
Multiple commits: https://asciinema.org/a/300808

lib/links.js
```diff
@@ -42,6 +43,18 @@ class LinkParser {
     return result;
   }
 
+  getPRUrlsFromArray(arr) {
+    const result = [];
+    for (const item of arr) {
+      const m = item.match(PR_RE);
+      if (!m) continue;
+      const ref = m[1];
```
BethGriggs: Should this variable be named something along the lines of `prUrl` rather than `ref`?

lib/links.js
```diff
@@ -42,6 +43,18 @@ class LinkParser {
...
+      const url = this.getRefUrlFromOP(ref);
+      if (url) result.push(url);
+    }
+    return result;
+  }
+
   // Do this so we can reliably get the correct url.
   // Otherwise, the number could reference a PR or an issue.
   getRefUrlFromOP(ref) {
```
BethGriggs: Potentially rename to `getUrlFromOP` as it now returns more than just refs?

PR #388, opened 2020-02-29
git-node: add git-node-release

codebytere:
This PR constitutes the first portion of an automated release flow to abstract away the tedious aspects of preparing and promoting releases.

The preparation and promotion aspects of this new flow will be split into two separate PRs for easier review, so this is just the preparation step, with the promotion step marked as a TODO for implementation in a subsequent PR.

There are several things i'd like to refactor to make this a little less verbose in the future, but to avoid scope creep for now i've just marked some of them TODO and unless you all feel they are blocking i'll move to address them in follow-ups.

### DIFF
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -12,11 +12,11 @@ const PROMOTE = 'promote';
 
 const releaseOptions = {
   prepare: {
-    describe: 'Prepare a new release with the given version number',
+    describe: 'Prepare a new release of Node.js',
     type: 'boolean'
   },
   promote: {
-    describe: 'Promote new release with the given version number',
+    describe: 'Promote new release of Node.js',
     type: 'boolean'
   },
   security: {
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -36,7 +36,7 @@ function builder(yargs) {
 
 function handler(argv) {
   if (argv.newVersion) {
-    const newVersion = semver.coerce(argv.newVersion);
+    const newVersion = semver.clean(argv.newVersion);
     if (semver.valid(newVersion)) {
       if (argv.prepare) {
         return release(PREPARE, argv);
@@ -81,7 +81,7 @@ async function main(state, argv, cli, dir) {
     // Check the branch diff to determine if the releaser
     // wants to backport any more commits before proceeding.
     cli.startSpinner('Fetching branch-diff');
-    const raw = prep.getBranchDiff();
+    const raw = prep.getBranchDiff({ onlyNotableChanges: false });
     const diff = raw.split('*');
     cli.stopSpinner('Got branch diff');
 
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -35,15 +35,10 @@ function builder(yargs) {
 }
 
 function handler(argv) {
-  if (argv.newVersion) {
-    const newVersion = semver.clean(argv.newVersion);
-    if (semver.valid(newVersion)) {
-      if (argv.prepare) {
-        return release(PREPARE, argv);
-      } else if (argv.promote) {
-        return release(PROMOTE, argv);
-      }
-    }
+  if (argv.prepare) {
+    return release(PREPARE, argv);
+  } else if (argv.promote) {
+    return release(PROMOTE, argv);
   }
 
   // If more than one action is provided or no valid action
@@ -78,6 +73,18 @@ async function main(state, argv, cli, dir) {
 
     if (prep.warnForWrongBranch()) return;
 
+    // If the new version was automatically calculated, confirm it.
+    if (!argv.newVersion) {
+      const create = await cli.prompt(
+        `Create release with new version ${prep.newVersion}?`,
+        { defaultAnswer: true });
+
+      if (!create) {
+        cli.error('Aborting release preparation process');
+        return;
+      }
+    }
+
     // Check the branch diff to determine if the releaser
     // wants to backport any more commits before proceeding.
     cli.startSpinner('Fetching branch-diff');
@@ -85,17 +92,20 @@ async function main(state, argv, cli, dir) {
     const diff = raw.split('*');
     cli.stopSpinner('Got branch diff');
 
-    const staging = `v${semver.major(argv.newVersion)}.x-staging`;
-    const proceed = await cli.prompt(
-      `There are ${diff.length - 1} commits that may be ` +
-      `backported to ${staging} - do you still want to proceed?`,
-      { defaultAnswer: false });
+    const outstandingCommits = diff.length - 1;
+    if (outstandingCommits !== 0) {
+      const staging = `v${semver.major(prep.newVersion)}.x-staging`;
+      const proceed = await cli.prompt(
+        `There are ${outstandingCommits} commits that may be ` +
+        `backported to ${staging} - do you still want to proceed?`,
+        { defaultAnswer: false });
 
-    if (!proceed) {
-      const seeDiff = await cli.prompt(
-        'Do you want to see the branch diff?');
-      if (seeDiff) cli.log(raw);
-      return;
+      if (!proceed) {
+        const seeDiff = await cli.prompt(
+          'Do you want to see the branch diff?');
+        if (seeDiff) cli.log(raw);
+        return;
+      }
     }
 
     return prep.prepare();
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -1,14 +1,19 @@
 'use strict';
 
 const semver = require('semver');
+const yargs = require('yargs');
+
+const auth = require('../../lib/auth');
 const CLI = require('../../lib/cli');
 const Release = require('../../lib/release');
 const Request = require('../../lib/request');
+const TeamInfo = require('../../lib/team_info');
 const { runPromise } = require('../../lib/run');
-const yargs = require('yargs');
 
 const PREPARE = 'prepare';
 const PROMOTE = 'promote';
+
+const RELEASERS = 'releasers';
 
 const releaseOptions = {
   prepare: {
@@ -69,8 +74,16 @@ module.exports = {
 async function main(state, argv, cli, req, dir) {
   const release = new Release(state, argv, cli, req, dir);
 
-  // TODO(codebytere): check if this command is being run by
-  // someone on the Releasers team in GitHub before proceeding.
+  cli.startSpinner('Verifying Releaser status');
+  const credentials = await auth({ github: true });
+  const request = new Request(credentials);
+  const info = new TeamInfo(cli, request, 'nodejs', RELEASERS);
+  const releasers = await info.getMembers();
+  if (!releasers.some(r => r.login === release.username)) {
+    cli.stopSpinner(`${release.username} is not a Releaser; aborting release`);
+    return;
+  }
+  cli.stopSpinner('Verified Releaser status');
 
   if (state === PREPARE) {
     if (release.warnForWrongBranch()) return;
@@ -91,7 +104,7 @@ async function main(state, argv, cli, req, dir) {
     if (!proceed) {
       const seeDiff = await cli.prompt(
         'Do you want to see the branch diff?', true);
-      if (seeDiff) console.log(raw);
+      if (seeDiff) cli.log(raw);
       return;
     }
 
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -1,0 +1,102 @@
+'use strict';
+
+const semver = require('semver');
+const CLI = require('../../lib/cli');
+const Release = require('../../lib/release');
+const Request = require('../../lib/request');
+const { runPromise } = require('../../lib/run');
+const yargs = require('yargs');
+
+const PREPARE = 'prepare';
+const PROMOTE = 'promote';
+
+const releaseOptions = {
+  prepare: {
+    describe: 'Prepare a new release with the given version number',
+    type: 'boolean'
+  },
+  security: {
+    describe: 'Prepare a new security release',
+    type: 'boolean'
+  }
+};
+
+function builder(yargs) {
+  return yargs
+    .options(releaseOptions).positional('newVersion', {
+      describe: 'Version number of the release to be created'
+    })
+    .example('git node release 1.2.3',
+      'Prepare a new release of Node.js tagged v1.2.3');
+}
+
+function handler(argv) {
+  if (argv.newVersion) {
+    const newVersion = semver.coerce(argv.newVersion);
+    if (semver.valid(newVersion)) {
+      return release(PREPARE, argv);
+    }
+  }
+
+  // If more than one action is provided or no valid action
+  // is provided, show help.
+  yargs.showHelp();
+}
+
+function release(state, argv) {
+  const logStream = process.stdout.isTTY ? process.stdout : process.stderr;
+  const cli = new CLI(logStream);
+
+  const req = new Request();
+  const dir = process.cwd();
+
+  return runPromise(main(state, argv, cli, req, dir)).catch((err) => {
+    if (cli.spinner.enabled) {
+      cli.spinner.fail();
+    }
+    throw err;
+  });
+}
+
+module.exports = {
+  command: 'release [newVersion|options]',
+  describe:
+    'Manage an in-progress release or start a new one.',
+  builder,
+  handler
+};
+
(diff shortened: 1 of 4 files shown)


Trott:
@nodejs/releasers

lib/release.js
```diff
@@ -0,0 +1,427 @@
...
+  getCurrentBranch() {
+    return runSync('git', ['rev-parse', '--abbrev-ref', 'HEAD']).trim();
+  }
+
+  getLastRef() {
+    return runSync('git', ['describe', '--abbrev=0', '--tags']).trim();
+  }
+
+  getChangelog() {
+    return runSync('npx', [
```
codebytere: i'm considering just adding `branch-diff` and `changelog-maker` to deps - i'm impartial as to whether we do though if others feel more strongly.

targos: sgtm

lib/release.js
```diff
@@ -0,0 +1,427 @@
...
+      if (line.includes('#define NODE_MAJOR_VERSION')) {
+        arr[idx] = `#define NODE_MAJOR_VERSION ${versionComponents.major}`;
+      } else if (line.includes('#define NODE_MINOR_VERSION')) {
+        arr[idx] = `#define NODE_MINOR_VERSION ${versionComponents.minor}`;
+      } else if (line.includes('#define NODE_PATCH_VERSION')) {
+        arr[idx] = `#define NODE_PATCH_VERSION ${versionComponents.patch}`;
+      } else if (line.includes('#define NODE_VERSION_IS_RELEASE')) {
+        arr[idx] = '#define NODE_VERSION_IS_RELEASE 1';
+      } else if (line.includes('#define NODE_VERSION_IS_LTS')) {
+        this.isLTS = arr[idx].split(' ')[2] === 1;
```
targos:
```suggestion
        this.isLTS = arr[idx].split(' ')[2] === '1';
```

lib/release.js
```diff
@@ -0,0 +1,427 @@
...
+      ];
+
+      if (versionComponents.patch !== 0) {
+        excludeLabels.push('semver-minor');
+      }
+
+      branchDiffOptions = [
+        'branch-diff',
+        stagingBranch,
+        'master',
```
targos:
```suggestion
        // TODO: use Current release branch instead of master for LTS
        'master',
```

lib/release.js
```diff
@@ -0,0 +1,427 @@
...
+        `${upstream}/${releaseBranch}`,
+        proposalBranch,
+        '--require-label=notable-change',
+        '-format=simple'
+      ];
+    } else {
+      const excludeLabels = [
+        'semver-major',
+        `dont-land-on-v${versionComponents.major}.x`,
+        `backport-requested-v${versionComponents.major}.x`
```
targos:
```suggestion
        `backport-requested-v${versionComponents.major}.x`,
        `backported-to-v${versionComponents.major}.x`,
        `backport-blocked-v${versionComponents.major}.x`,
```

lib/release.js
```diff
@@ -0,0 +1,427 @@
...
+      ];
+
+      if (versionComponents.patch !== 0) {
```
targos:
```suggestion
      if (this.isLTS) {
        excludeLabels.push('baking-for-lts');
      }

      if (versionComponents.patch !== 0) {
```

targos:
Thanks for working on this!

lib/release.js
MylesBorins: I'd be much more comfortable with including them as dependencies than running via npx. At the very least you could use a child_process with `node_modules/.bin` in the path and just call the binary.

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+      '      * https://ci.nodejs.org/job/node-test-pull-request/\n' +
+      '      * https://ci.nodejs.org/job/citgm-smoker/');
+    cli.info(
+      'If this release has deps/v8 changes, you\'ll also need to run:\n' +
+      '      * https://ci.nodejs.org/job/node-test-commit-v8-linux/'
+    );
+  }
+
+  async promote() {
+    // TODO(codebytere): implement.
```
MylesBorins: Might be reasonable to keep promotion as a separate thing in the node repo... although I definitely see the value in having the workflow all in one place.

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+
+    cli.ok(`Aborted \`git node release\` for version ${newVersion}`);
+  }
+
+  getCurrentBranch() {
+    return runSync('git', ['rev-parse', '--abbrev-ref', 'HEAD']).trim();
+  }
+
+  getLastRef() {
+    return runSync('git', ['describe', '--abbrev=0', '--tags']).trim();
```
MylesBorins: OHHHHH gitfu I didn't know. This is super cool

components/git/release.js
```diff
@@ -0,0 +1,115 @@
...
+  handler
+};
+
+async function main(state, argv, cli, req, dir) {
+  const release = new Release(state, argv, cli, req, dir);
+
+  cli.startSpinner('Verifying Releaser status');
+  const credentials = await auth({ github: true });
+  const request = new Request(credentials);
+  const info = new TeamInfo(cli, request, 'nodejs', RELEASERS);
```
MylesBorins: if it is an LTS release we should likely check that they are also a member of the LTS team.

Alternatively, if this is just for creating the release neither of the checks might matter... we've had non releasers prep (in fact it is part of onboarding)

components/git/release.js
```diff
@@ -0,0 +1,115 @@
...
+
+    if (!proceed) {
+      const seeDiff = await cli.prompt(
+        'Do you want to see the branch diff?', true);
+      if (seeDiff) cli.log(raw);
+      return;
+    }
+
+    return release.prepare();
+  } else if (state === PROMOTE) {
```
MylesBorins: as preparation and promotion are always happening at a different time it might make sense to make this completely different subcommands... e.g. `git node release prepare` vs `git node release promote`

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+    await this.createProposalBranch();
+
+    // Update version and release info in src/node_version.h.
+    const shouldUpdateNodeVersion = await cli.prompt(
+      `Update 'src/node_version.h' for ${newVersion}?`, true);
+    if (!shouldUpdateNodeVersion) return this.abort();
+    await this.updateNodeVersion();
+
+    // Check whether to update NODE_MODULE_VERSION (default false).
+    const shouldUpdateNodeModuleVersion = await cli.prompt(
```
MylesBorins: this could likely happen automatically for a SemverMajor right?

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+        'Specify variant (ex. \'v8_7.9\') for new NODE_MODULE_VERSION:',
+        false);
+      const versions = await cli.promptInput(
+        'Specify versions (ex. \'14.0.0-pre\') for new NODE_MODULE_VERSION:',
+        false);
+      this.updateNodeModuleVersion(runtime, variant, versions);
+    }
+
+    // Update any REPLACEME tags in the docs.
+    const shouldUpdateREPLACEMEs = await cli.prompt(
```
MylesBorins: would we ever want to not do this? Also seems like something that could "just happen"™️

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+      'Update REPLACEME items in docs?', true);
+    if (!shouldUpdateREPLACEMEs) return this.abort();
+    await this.updateREPLACEMEs();
+
+    this.date = await cli.promptInput(
+      'Enter release date in YYYY-MM-DD format:');
+
+    // Update Changelogs
+    const shouldUpdateChangelogs = await cli.prompt(
```
MylesBorins: Another one that probs should "just happen"™️

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+    cli.stopSpinner('Updated CHANGELOG.md');
+
+    cli.startSpinner(`Updating CHANGELOG_V${versionComponents.major}.md`);
+    await this.updateMajorChangelog();
+    cli.stopSpinner(`Updated CHANGELOG_V${versionComponents.major}.md`);
+
+    await cli.prompt('Finished editing the changelogs?', false);
+
+    // Create release commit.
+    const shouldCreateReleaseCommit = await cli.prompt(
```
MylesBorins: This is an interesting one... because my first thought was "just happen"™️ but realistically it might make sense for this to be optional. The one case I'm specifically thinking about is "update a release against staging". But in retrospect that might be a different command like `git node release --update`... which would essentially reset to staging and run the scripts to generate release commit again.

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+      'Create release commit?', true);
+    if (!shouldCreateReleaseCommit) return this.abort();
+
+    // Proceed with release only after the releaser has amended
+    // it to their liking.
+    const created = await this.createReleaseCommit();
+    if (!created) {
+      const lastCommitSha = runSync('git', ['rev-parse', '--short', 'HEAD']);
```
MylesBorins: what is the case where we would hit this?

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+  openPullRequest() {
+    const { newVersion, upstream, cli, versionComponents } = this;
+    const proposalBranch = `v${newVersion}-proposal`;
+    const releaseBranch = `v${versionComponents.major}.x`;
+
+    const pushed = runSync('git', ['push', upstream, proposalBranch]).trim();
+
+    if (pushed) {
+      runSync('open',
+        [`https://github.com/nodejs/node/compare/${releaseBranch}...${proposalBranch}?expand=1`]);
```
MylesBorins: this is dope!

lib/release.js
```diff
@@ -0,0 +1,437 @@
...
+      const excludeLabels = [
+        'semver-major',
+        `dont-land-on-v${versionComponents.major}.x`,
+        `backport-requested-v${versionComponents.major}.x`,
+        `backported-to-v${versionComponents.major}.x`,
+        `backport-blocked-v${versionComponents.major}.x`
+      ];
+
+      if (isLTS) {
+        excludeLabels.push('semver-minor');
```
MylesBorins: there are times where LTS includes semver-minor fwiw so this check should likely be `isLTS && !semverMinor`

lib/release.js
codebytere: i thought so, but i don't know through what mechnism the scripts would know what the variant would be 🤔

components/git/release.js
codebytere: the more i think about this the more i agree - it's enough code that the file is gonna get extremely chonky once i add the promote step code. I'll split this out and just disregard the promote step entirely in this PR.

lib/release.js
codebytere: i went with a similar flow as git node land - sometimes you don't want the amended commit message and instead want to edit it yourself to add something that's notable but not tagged as such. In that case, i'd want to wait for the reviewer to give me the go-ahead (& allow them to make those edits) before continuing, i think

lib/release.js
codebytere: good point - i can just log that it happened without making it interactive

lib/release.js
MylesBorins: If you compare previous tag to new tag and major version is different?

lib/release.js
codebytere: i think for this one i do want to remove the prompt to _start_ editing changelogs, but for similar reasons to the release commit i still would potentially want to pause and confirm that the releaser is happy with the updated changelog before proceeding, so that they can shove in an extra notable commit or something 🤔

lib/release.js
codebytere: the variant is something like `v8_7.9` - where does that come from? I just figured it'd be easier to provide but if it's easy to autodetermine i'll swap it out!

components/git/release.js
codebytere: moved to promoter section only for now!

MylesBorins:
Just tried to use it and got an error

```
⠋ Fetching branch-diffTypeError: Cannot read property 'onlyNotableChanges' of undefined
    at ReleasePreparation.getBranchDiff (/Users/mborins/code/node-core-utils/lib/prepare_release.js:352:14)
    at main (/Users/mborins/code/node-core-utils/components/git/release.js:84:22)
    at release (/Users/mborins/code/node-core-utils/components/git/release.js:59:21)
    at Object.handler (/Users/mborins/code/node-core-utils/components/git/release.js:42:16)
    at Object.runCommand (/Users/mborins/code/node-core-utils/node_modules/yargs/lib/command.js:240:40)
    at Object.parseArgs [as _parseArgs] (/Users/mborins/code/node-core-utils/node_modules/yargs/yargs.js:1145:41)
    at Object.get [as argv] (/Users/mborins/code/node-core-utils/node_modules/yargs/yargs.js:1079:21)
    at Object.<anonymous> (/Users/mborins/code/node-core-utils/bin/git-node:20:3)
    at Module._compile (internal/modules/cjs/loader.js:1147:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:10)
```

MylesBorins:
Also... for `git node release --prepare` it would be nice to have a default that determined for you the type of release (patch / minor / major) rather than having to provide the explicit version number.

components/git/release.js
```diff
@@ -0,0 +1,105 @@
...
+async function main(state, argv, cli, dir) {
+  if (state === PREPARE) {
+    const prep = new ReleasePreparation(argv, cli, dir);
+
+    if (prep.warnForWrongBranch()) return;
+
+    // Check the branch diff to determine if the releaser
+    // wants to backport any more commits before proceeding.
+    cli.startSpinner('Fetching branch-diff');
+    const raw = prep.getBranchDiff();
```
MylesBorins:
```suggestions
const raw = prep.getBranchDiff({ onlyNotableChange: false });
```

MylesBorins:
Some bugs I found playing with this

* I provided the version as `v13.11.0` and the branch it crearted was `vv13.11.0` (and similar duplication across meta data. Should likely not do that 😅 (ignore v)
* I accidentally said yes to changing module version number which resulted in reformatting the entire module version data and thus a lot of churn. Might want to make sure that generated module version number information has a consistent style. Honestly the question came as a surprise... and we should maybe only ask it for Semver-Major release (imho)

MylesBorins:
I also just accidentally clicked through on the "create PR" before I had a chance to audit stuff... would prefer to not have that step 😅

MylesBorins:
One other bit... while it is nice having the notable commit information in the commit message we likely don't want it as markdown

MylesBorins:
Automation is missing one of the link references at the top of the version specific changelog 

https://github.com/nodejs/node/pull/32218

MylesBorins:
dogfooded again. Everything worked great until I attempted to push upstream and then it bailed with the following warning
```
Error: remote:
remote: Create a pull request for 'v13.12.0-proposal' on GitHub by visiting:
remote:      https://github.com/nodejs/node/pull/new/v13.12.0-proposal
remote:
To github.com:nodejs/node.git
 * [new branch]            v13.12.0-proposal -> v13.12.0-proposal

    at exports.runSync (/Users/mborins/code/node-core-utils/lib/run.js:60:11)
    at ReleasePreparation.pushBranch (/Users/mborins/code/node-core-utils/lib/prepare_release.js:247:5)
    at ReleasePreparation.prepare (/Users/mborins/code/node-core-utils/lib/prepare_release.js:160:10)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

rvagg:
wow, this is a great way to deal with release complexity, and it even edits abi_version_registry!

Minor suggestion is to add changelog-maker to the dependencies list, not strictly essential but since you're directly relying on the executable it it may be a good idea to lock the version to the code rather than whatever branch-diff pulls in.

MylesBorins:
Seems like the href for the prior release is getting replaced rather than appended in CHANGELOG_vX.x

https://github.com/nodejs/node/pull/32376#discussion_r395615448

MylesBorins:
trying again and the "auto detect" wasn't working for me... it kept trying to make a v14.x release and complaining that i was on v13.x (for a 13.x release)

targos:
Rubberstamp LGTM. I'll try it for the next release I prepare.

PR #392, opened 2020-03-11
Fixing a typo in CONTRIBUTING.md

mohouyizme:
"intergration" is a misspelling of "integration".


priyank-p:
Merged, thanks @mohouyizme for the typo fix.

Also, I don't know if we want to get rid of the mention of `test/integration` since a) we don't have test files there and b) fixtures could be used for a request made to an external server as we do for github?

mohouyizme:
I think so.

PR #397, opened 2020-03-21
feat: support metadata query by userid

srl295:
`git node metadata @srl295` will show up to 10 active PRs and give the current status.

I'm not a frequent contributor to node. But, sometimes I have a couple of different PRs in the works. This helps me find out what the current status is.

### DIFF
--- a/components/metadata.js
+++ b/components/metadata.js
@@ -3,18 +3,14 @@
 const Request = require('../lib/request');
 const auth = require('../lib/auth');
 const PRData = require('../lib/pr_data');
+const PRQuery = require('../lib/pr_query');
 const PRSummary = require('../lib/pr_summary');
 const PRChecker = require('../lib/pr_checker');
 const MetadataGenerator = require('../lib/metadata_gen');
 
 const fs = require('fs');
 
-module.exports = async function getMetadata(argv, cli) {
-  const credentials = await auth({
-    github: true
-  });
-  const request = new Request(credentials);
-
+async function getOneMetadata(argv, cli, request) {
   const data = new PRData(argv, cli, request);
   await data.getAll();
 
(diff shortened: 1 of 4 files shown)


components/metadata.js
```diff
@@ -47,4 +43,29 @@ module.exports = async function getMetadata(argv, cli) {
...
+    const prids = await myQuery.getPRs();
+    for (const prid of prids) {
+      cli.separator(`${prid}`);
+      // new argv, with the prid
+      const nargv = Object.assign({}, argv, { prid });
+      delete nargv.assignee;
+      await getOneMetadata(nargv, cli, request);
+      cli.separator();
+    }
+    return {}; // ?
```
srl295: not sure what the return should be here

joyeecheung: I think a better way to handle multiple requests is to add a different method for it, which calls the old `getMetadata`, and then call it from the command handler, instead of overwriting the method that's used to querying single requests. That also helps you avoid the `nargv.assignee` tinkering. In addition you may want to update the `fs.writeFileSync` call in `getMetadata` to `fs.appendFile` instead.

https://github.com/nodejs/node-core-utils/blob/1e9daeb15475efdfea4a418686395299ad892149/components/git/metadata.js#L83-L88

lib/pr_query.js
```diff
@@ -0,0 +1,36 @@
...
+const PR_QUERY = 'PRQuery';
+
+class PRQuery {
+  /**
+   * @param {Object} argv
+   * @param {Object} cli
+   * @param {Object} request
+   */
+  constructor(argv, cli, request) {
+    const { assignee, owner, repo } = argv;
```
joyeecheung: I think it's probably enough to just pass these directly here instead of through `argv` - this is very far away from the command handler now and `argv` isn't too meaningful here.

codebytere:
@srl295 do you think this is something you still want to work on?

srl295:
Can't now, job change
________________________________

Issue #393, opened 2020-03-11
`ncu-ci` flags will fail silently if no value is passed


mmarchini:
```console
$ ncu-ci --copy url https://ci.nodejs.org/job/node-test-pull-request/29714/

 # Command didn't run, nothing copied to clipboard, process exited with 0
$ ncu-ci --markdown url https://ci.nodejs.org/job/node-test-pull-request/29714/

 # Command didn't run, no error message explainint to the user they should use `--markdown <file>`
$ ncu-ci --markdown foo.md url https://ci.nodejs.org/job/node-test-pull-request/29714/
--------------------------------------------------------------------------------                                                                                                                                                                                                  
[1/1] Running PR: 29714                                                                                                                                                                                                                                                           
--------------------------------------------------------------------------------                                                                                                                                                                                                  
✔  Build data downloaded
✔  Data downloaded
----------------------------------- Summary ------------------------------------
Result     FAILURE
URL        https://ci.nodejs.org/job/node-test-pull-request/29714/
Source     https://github.com/nodejs/node/pull/32116/
Commit     [0f38b8d45bd1] deps: patch V8 to run on Xcode 8
Date       2020-03-10 17:04:14 -0700
Author     Matheus Marchini <mmarchini@netflix.com>
----------------------- ubuntu1804_sharedlibs_debug_x64 ------------------------
...
--------------------------------- Other builds ---------------------------------
Unstable   https://ci.nodejs.org/job/node-test-commit-linux/33387/
Unstable   https://ci.nodejs.org/job/node-test-commit-osx/32540/
Unstable   https://ci.nodejs.org/job/node-test-commit-arm/29944/
Unstable   https://ci.nodejs.org/job/node-test-commit-windows-fanned/34549/
--------------------------------------------------------------------------------
Written markdown to foo.md
```

PR #411, opened 2020-05-06
fix: validate ncu-ci args

codebytere:
Closes https://github.com/nodejs/node-core-utils/issues/393.

Not a perfect solution, but this validates that the user didn't type something like:
```
ncu-ci --markdown url https://ci.nodejs.org/job/node-test-pull-request/29714/
```
 and throws a helpful error if that's the case.

cc @mmarchini @joyeecheung

### DIFF
--- a/bin/ncu-ci
+++ b/bin/ncu-ci
@@ -23,6 +23,14 @@ const Request = require('../lib/request');
 const CLI = require('../lib/cli');
 const yargs = require('yargs');
 
+const commandKeys = [
+  'rate',
+  'walk',
+  'url',
+  'pr',
+  'commit'
+];
+
 // eslint-disable-next-line no-unused-vars
 const argv = yargs
   .command({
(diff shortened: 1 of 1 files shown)


bin/ncu-ci
```diff
@@ -115,13 +123,20 @@ const argv = yargs
...
-    describe: 'Write the results as markdown to the path'
+    describe: 'Write the results as markdown to <path>'
+  }).check(argv => {
+    if (argv.markdown && commandKeys.includes(argv.markdown)) {
+      throw new Error('--markdown <path> did not specify a valid path');
+    } else if (argv.json && commandKeys.includes(argv.json)) {
+      throw new Error('--json <path> did not specify a valid path');
+    }
+    return true;
   })
```
targos, line 140: I think we can use `requiresArg: true` in the option's definition instead of manual validation

codebytere:
@targos i tried that but it didn't seem to have the desired effect - it just silently fails as before:

```
node-core-utils on git:validate-ncu-ci-args ❯ bin/ncu-ci --markdown url https://ci.nodejs.org/job/node-test-pull-request/29714/
```

if i specify:
```
.option('json <path>', {
    type: 'string',
    requiresArg: true,
    describe: 'Write the results as json to <path>'
  })
  .option('markdown <path>', {
    type: 'string',
    requiresArg: true,
    describe: 'Write the results as markdown to <path>'
  })
```
with no manual validation

PR #415, opened 2020-05-09
fix: enforce default for non-confirmation prompts

codebytere:
Fixes an edge case whereby if the user prompted for a `number` or `input` type and did not provide a default answer, `prompt()` would return true and like created a weird state.

This fixes that by enforcing that if non-`confirmation` prompts are invoked, a default is enforced. Also added tests to confirm correct behavior.

cc @targos @joyeecheung

### DIFF
--- a/test/unit/cli.test.js
+++ b/test/unit/cli.test.js
@@ -173,6 +173,36 @@ describe('cli', () => {
       cli.setAssumeYes();
     });
 
+    it('rejects when no default answer is provided for \'input\' type', () => {
+      return assert.rejects(async() => {
+        await cli.prompt('What is your favorite color', {
+          questionType: cli.QUESTION_TYPE.INPUT
+        });
+      }, /defaultAnswer must be provided for non-confirmation prompts/);
+    });
+
+    it('rejects when no default answer is provided for \'number\' type', () => {
+      return assert.rejects(async() => {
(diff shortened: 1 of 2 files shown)


test/unit/cli.test.js
```diff
@@ -173,6 +173,36 @@ describe('cli', () => {
       cli.setAssumeYes();
     });
 
+    it('rejects when no default answer is provided for \'input\' type', () => {
+      assert.rejects(() => {
```
targos, line 177: `assert.rejects` returns a Promise so I think we need to either return it or convert the test to an async function and await (if mocha supports that)

test/unit/cli.test.js
```diff
@@ -173,6 +173,36 @@ describe('cli', () => {
...
+        cli.prompt('What is your favorite color', {
+          questionType: cli.QUESTION_TYPE.INPUT
+        });
+      }, /defaultAnswer must be provided for non-confirmation prompts/);
+    });
+
+    it('rejects when no default answer is provided for \'number\' type', () => {
+      assert.rejects(() => {
```
targos, line 185: ditto

test/unit/cli.test.js
codebytere: whoops, fixed!

PR #420, opened 2020-05-20
feat: handle unmarked DEP0XXX tags during landing and release

codebytere:
Adds an additional step in `git node land` to automatically update any `DEP0XXX`, `DEP0XX1` etc tags in the codebase to incremented new numbers. The release preparation script will also check for unmarked numbers and optionally update them as well.

Tested locally but please double check for any potential edge cases i might have missed!

cc @targos

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -1,14 +1,16 @@
 'use strict';
 
 const path = require('path');
+const {
+  getUnmarkedDeprecations,
+  updateDeprecations
+} = require('./deprecations');
 
 const {
   runAsync, runSync, forceRunAsync
 } = require('./run');
 const Session = require('./session');
-const {
-  shortSha
-} = require('./utils');
+const { shortSha } = require('./utils');
 
 const isWindows = process.platform === 'win32';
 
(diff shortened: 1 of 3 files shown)


richardlau:
It would be better if this was part of the `git node land <pr>` workflow to stop `DEP0XXX` landing in the first place. Typically we'd fix those up in master before backporting (because I think we want the numbers to be consistent across master/release lines). I guess we could fix these at release time as long as the change is merged upstream afterwards (hopefully without conflicts).

codebytere:
@richardlau good point - i added this since it came up yesterday during the v14 release with [this PR](https://github.com/nodejs/node/pull/33294), but a second pass of the releases guide shows:

> This assignment should
occur when the PR is landed, but a check will be made when the release build is
run.

I'll migrate this logic into the land logic!

richardlau:
> 
> 
> @richardlau good point - i added this since it came up yesterday during the v14 release with [this PR](https://github.com/nodejs/node/pull/33294), but a second pass of the releases guide shows:
> 
> > This assignment should occur when the PR is landed, but a check will be made when the release build is run.
> 
> I'll migrate this logic into the land logic!

If you want to also codify the check during release (in addition to modifying the land logic) I'd have no objection 😁.

codebytere:
@targos i'd love to be able to get this in before next week's release if you have some time to peek at it 🙇‍♀️

codebytere:
@targos in that case - should i just adapt this PR to check for unmarked DEP0XXX tags and not try to fix them?

codebytere:
@targos i've thought about this a bit more - i think it would still make sense to automate multiple in the landing stage? I can't think of a case where there will be two conflicting deprecations introduced in the same PR 🤔 

In the release stage it makes sense only to automate if there's one though, i agree!

lib/landing_session.js
```diff
@@ -84,6 +86,19 @@ class LandingSession extends Session {
...
+
+    // Check for and maybe assign any unmarked deprecations in the codebase.
+    const unmarkedDeprecations = await getUnmarkedDeprecations();
+    const unmarkedDepCount = unmarkedDeprecations.length;
+    if (unmarkedDepCount > 0) {
+      cli.startSpinner('Assigning deprecation numbers to DEPOXXX items');
+      await updateDeprecations(unmarkedDeprecations);
+
+      // Amend the last commit with the updated deprecation items.
+      await runAsync('git', ['commit', '--amend', '--no-edit']);
```
targos, line 98: do we need to `git add` the modified files?

codebytere: heh yes - fixed 🤦‍♀️

Issue #421, opened 2020-05-20
`git node backport v8` does not work on macOS


dominykas:
I got an error saying `Resolve merge conflicts and enter 'RESOLVED'` - however there were no conflicts. The actual underlying failure was `patch: unrecognized option '--merge'` - it seems the `patch` on macOS does not have that option.

I'm happy to research/PR whatever would be considered an acceptable fix?

Also not sure if there's a good way to detect conflicts vs other kinds of errors - the error message was rather misleading.

codebytere:
The line originates [here](https://github.com/nodejs/node-core-utils/blob/a9c0676d0f64df5c8e86d7287f69381dc64049f2/lib/update-v8/backport.js#L203-L210) and unfortunately it looks like you're right - the patch commands are similar on macOS and on Linux but there are some nontrivial option differences, one of which is the `--merge` command.

I think our best bet here is probably to move away from the `patch` command and use `git apply` to accomplish the same goal. The consequent difference unfortunately though is that `git apply` won't apply patch if you have other uncommitted changes in your code, & it also won't apply a patch that does not apply cleanly. The above linked code would therefore require some light restructuring to take those externalities into account.

cc @targos for other thoughts!

dominykas:
I don't know what `patch --merge` leaves on the filesystem when there's a conflict, but `git apply -3` (to do a three-way merge) worked for me 👍

targos:
@dominykas what did `git apply -3` leave?

dominykas:
@targos the `<<<<<<< ours` delimited stuff - not sure what it's called.

targos:
I vaguely remember that `git apply` didn't always leave conflict markers, but maybe I'm wrong. And it's probably better to have something that work sometimes on all platforms than never on macOS

PR #432, opened 2020-05-30
fix: use git apply not system apply

codebytere:
Closes https://github.com/nodejs/node-core-utils/issues/421.

Switches to `git apply` vs system `apply`, since the available options for system `apply` differ between systems.

cc @targos

### DIFF
--- a/components/git/v8.js
+++ b/components/git/v8.js
@@ -82,9 +82,14 @@ function main(argv) {
     options.v8Dir = path.resolve(options.v8Dir);
   }
 
-  options.execGitNode = function execGitNode(...args) {
-    return execa('git', args, { cwd: options.nodeDir });
+  options.execGitNode = function execGitNode(cmd, args, input) {
+    args.unshift(cmd);
+    return execa('git', args, {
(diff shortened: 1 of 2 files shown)


targos:
There is some regression. For example with this commit: https://github.com/v8/v8/commit/50af61718d49c4f04d0839476c78e3b305d03de4

`git node v8 backport 50af61718d49c4f04d0839476c78e3b305d03de4` succeeds with the current implementation and doesn't work with this PR (it asks to resolve merge conflicts but there are no changed files).

codebytere:
@targos should be fixed up now - working with the supplied commit.

targos:
It broke one of the V8 commands:

https://ci.nodejs.org/job/node-update-v8-canary/1366/display/redirect

```
05:56:25] Commit V8 update [started]
[05:56:25] Commit V8 update [failed]
[05:56:25] → args.unshift is not a function
✖ args.unshift is not a function
```

I'll have a look later if I have time.

codebytere:
@targos ah dang i know what it is - i missed two files when i was grepping. Fix up in a bit.

Issue #427, opened 2020-05-30
Run make lint during `git node land`


BridgeAR:
Running the linter does not take long and it's at least a minimal verification that everything works as expected besides linting the commit message.

PR #435, opened 2020-06-04
fix: lint during the landing process

codebytere:
Closes https://github.com/nodejs/node-core-utils/issues/427.

Runs `make lint` after the patch has been downloaded and applied, and optionally stage and amend updated files if lint fails.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -199,7 +199,13 @@ class LandingSession extends Session {
 
     const cleanLint = await this.validateLint(patch);
     if (!cleanLint) {
-      const correctedLint = await cli.prompt('Corrected lint errors?');
+      const tryFixLint = await cli.prompt(
+        'Try fixing JS lint with \'make lint-js-fix\'?');
+      if (tryFixLint) {
+        await runAsync('make', ['lint-js-fix']);
+      }
+
+      const correctedLint = await cli.prompt('Corrected all lint errors?');
       if (correctedLint) {
         await runAsync('git', ['add', '.']);
 
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -132,18 +132,12 @@ class LandingSession extends Session {
   }
 
   async validateLint() {
-    const { cli } = this;
-
     // The linter is currently only run on non-Windows platforms.
     if (os.platform() === 'win32') {
       return true;
     }
 
     const linted = await runAsync('make', ['lint']);
-    if (!linted) {
-      cli.warn('There are lint errors in your patch. ' +
-        'Please fix them before proceeding');
-    }
     return linted;
   }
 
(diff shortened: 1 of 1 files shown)


richardlau:
> Closes #427.
> 
> Runs `make lint` after the patch has been downloaded and applied, and optionally stage and amend updated files if lint fails.

Isn’t that going to fail on Windows?

codebytere:
@richardlau you might be right - my esoteric windows knowledge is admittedly weaker 😅

targos:
Does git node land currently work on windows ?

codebytere:
We could also only run `make lint` on linux/macOS potentially

lib/landing_session.js
```diff
@@ -103,8 +103,20 @@ class LandingSession extends Session {
     return command;
   }
 
-  async suggestAfterPatch(patch) {
+  async validateLint() {
     const { cli } = this;
+
```
BridgeAR, line 108:
```suggestion

    if (os.platform() === 'win32') {
      // The linter is currently only run on non-Windows platforms.
      return true;
    }
```

BridgeAR:
We might want add some tests for these changes.

codebytere:
I don't know we have tests for any of the actual patch downloading code right now - @joyeecheung what do you think?

lib/landing_session.js
```diff
@@ -120,8 +121,25 @@ class LandingSession extends Session {
...
+    if (os.platform() === 'win32') {
+      return true;
+    }
+
+    const linted = await runAsync('make', ['lint']);
+    if (!linted) {
+      cli.warn('There are lint errors in your patch. ' +
+        'Please fix them before proceeding');
+      return false;
+    }
```
lundibundi, line 137:
```suggestion
    }
    return true;
```
?
OR
```javascript
const linted = await runAsync('make', ['lint']);
if (!linted) {
  cli.warn('There are lint errors in your patch. ' +
    'Please fix them before proceeding');
}
return linted;
```

lib/landing_session.js
```diff
@@ -157,16 +176,36 @@ class LandingSession extends Session {
...
+        'to start a new one run `git node land --abort`');
       return;
     }
     await this.tryResetBranch();
 
     const patch = await this.downloadAndPatch();
+
+    const cleanLint = await this.validateLint(patch);
+    if (!cleanLint) {
+      const correctedLint = await cli.prompt('Corrected lint errors?');
```
lundibundi, line 192: Sorry if I missed something but, is a user supposed to fix the issues in another terminal and then come back to this?

Should we perhaps try to run `make lint-js-fix` for the user here?

codebytere: @lundibundi yes - the intent is that they fix it elsewhere and continue. We could potentially try to run `make lint-js-fix` though - i'm not opposed to doing that as well!

codebytere: added

lib/landing_session.js
```diff
@@ -167,16 +186,42 @@ class LandingSession extends Session {
...
+    const cleanLint = await this.validateLint(patch);
+    if (!cleanLint) {
+      const tryFixLint = await cli.prompt(
+        'Try fixing JS lint with \'make lint-js-fix\'?');
+      if (tryFixLint) {
+        await runAsync('make', ['lint-js-fix']);
+      }
+
+      const correctedLint = await cli.prompt('Corrected all lint errors?');
+      if (correctedLint) {
```
lundibundi, line 209: :thinking: shouldn't we run `this.validateLint(patch);` again to make sure before committing or it will be checked later down the line?

lib/landing_session.js
```diff
@@ -167,16 +186,42 @@ class LandingSession extends Session {
...
-        'run `git node land --abort`');
+      cli.warn('Landing session already in progress - ' +
+        'to start a new one run `git node land --abort`');
       return;
     }
     await this.tryResetBranch();
 
     const patch = await this.downloadAndPatch();
+
+    const cleanLint = await this.validateLint(patch);
```
lundibundi, line 200:
```suggestion
    const cleanLint = await this.validateLint();
```

lib/landing_session.js
codebytere: @lundibundi we can't really - it validates the pre-downloaded patch from upstream which can't be updated per se.

codebytere:
@lundibundi i cleaned up the flow a bit - i think this is a bit better for users now

Issue #442, opened 2020-06-27
Set exit code when commands fail


mmarchini:
If I run `git node land` on a new machine with nothing configured, I'll get the instructions to set required variables, and then the process will exit with status code 0. In order for us to be able to use ncu on a Commit Queue, it needs to return the appropriate status codes when something goes wrong, otherwise automation won't work.

(I didn't check which other failures will exit with 0, but all failures should set the exit code to a non-zero code)

codebytere:
Closed by https://github.com/nodejs/node-core-utils/pull/443

mmarchini:
I'd like to keep this open because there are still many situations where the exit code will return zero even if the command failed. Would be good for us to list those as a next step.

PR #443, opened 2020-06-27
feat: set error code in some failure cases

mmarchini:
Adds an method to `cli` to set the process exit code, and use that
method on some known failure situations (missing config and `git node land`
failed checks). There are still more places where we should set exit
code.

Ref: https://github.com/nodejs/node-core-utils/issues/442

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -164,6 +164,10 @@ class CLI {
       this.log(prefix + chalk.bold(`${error}  ${obj}`));
     }
   }
+
+  setExitCode(statusCode) {
+    process.exitCode = statusCode;
+  }
 };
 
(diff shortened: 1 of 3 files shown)


lib/cli.js
```diff
@@ -164,6 +164,10 @@ class CLI {
       this.log(prefix + chalk.bold(`${error}  ${obj}`));
     }
   }
+
+  setErrorCode(statusCode = 1) {
```
codebytere, line 168: it might be a tad clearer if this was `setExitCode(statusCode)` and then we passed `1` explicitly - i'd probably not be 100% sure what this did without checking the definition i don't think 🤔

PR #445, opened 2020-06-27
ncu-ci: command to start CI for PRs

mmarchini:


lib/ci/run_ci.js
```diff
@@ -0,0 +1,73 @@
...
+        method: 'POST',
+        headers: {
+          'Jenkins-Crumb': crumb
+        },
+        body: this.payload
+      });
+      this.cli.stopSpinner('Starting PR CI job');
+      this.cli.ok('CI job started');
+    } catch (err) {
+      this.cli.stopSpinner('');
```
codebytere, line 67: stopSpinner takes an optional status parameter - we can just say
```
this.cli.stopSpinner('Failed to start CI', cli.SPINNER_STATUS.FAILED);
```
here and remove the `cli.error`

lib/ci/run_ci.js
```diff
@@ -0,0 +1,73 @@
...
+    try {
+      this.cli.startSpinner('Starting PR CI job');
+      await this.request.text(CI_PR_URL, {
+        method: 'POST',
+        headers: {
+          'Jenkins-Crumb': crumb
+        },
+        body: this.payload
+      });
+      this.cli.stopSpinner('Starting PR CI job');
```
codebytere, line 64: same as below - we can remove the `cli.ok` and say:
```
this.cli.stopSpinner('PR CI job successfully started');
```
(it assumes success by default)

lib/ci/run_ci.js
```diff
@@ -0,0 +1,73 @@
...
+    this.repo = repo;
+    this.prid = prid;
+  }
+
+  async getCrumb() {
+    try {
+      const { crumb } = await this.request.json(CI_CRUMB_URL);
+      return crumb;
+    } catch (e) {
+      this.cli.error('Failed to get breadcrumbs');
```
codebytere, line 29: the cli error output nesting might be a little weird here  🤔 we might instead want to return `false` in the catch block and then down at the callsite do something like
```
this.cli.startSpinner('Validating Jenkins credentials');
const crumb = await this.getCrumb();

if (!crumb) {
  this.cli.stopSpinner('Jenkins credentials invalid');
  return;
}
this.cli.stopSpinner('Jenkins credentials valid');
```

mmarchini:
Done. I also made some changes so that if we fail to start the CI for any reason, the process exits with status code != 0 :)

mmarchini:
I think this deserves a @nodejs/build ping to double-check if we're not doing anything we shouldn't in the Jenkins calls (for example, using a user token instead of the token defined in the Job configuration).

bin/ncu-ci
```diff
@@ -147,6 +171,57 @@ const commandToType = {
...
+                'or pass it via the --repo option');
+    }
+    if (!owner) {
+      cli.error('GitHub owner is missing, please set it via ncu-config ' +
+                'or pass it via the --owner option');
+      validArgs = false;
+    }
+    if (!validArgs) {
+      // TODO(mmarchini): change to our API once #443 lands
+      // this.cli.setExitCode(1);
```
lundibundi, line 212: This can now be fixed.

mmarchini:
Added some tests. Would be great if we could include this in the next release, as it is required for the Request CI Queue on nodejs/node :)

codebytere:
I might not have all the context i might need as i'm not a member of @nodejs/build, but if no one else comments then i think we may be fine to go ahead and merge?

mmarchini:
That's what I'm thinking, yes. If something with this approach is not right/good, we can always improve on a future version (or remove if needed)

Trott:
I'm pretty sure @nodejs/jenkins-admins is a subset of @nodejs/build, but uh, pinging them anyway....

mmarchini:
If folks want to help review but don't want/have time to dabble into ncu code, the main question is: is this the right/best way to start a Jenkins job via API? Below is a simplified version of the flow implemented in the PR:

```javascript
// this code probably doesn't work, but it shows 
// which calls the PR makes to start a job via 
// Jenkins API

// Use the token generated on Jenkins interface to
// authenticate.
const username = '***';
const personalJenkinsToken = '***';
const auth = Buffer
  .from(`${username}:${personalJenkinsToken}`)
  .toString('base64')
const headers = { Authorization: `Basic ${auth}` };

// Get breadcrumb for CSRF validation
const { crumb } = await (await fetch('https://ci.nodejs.org/crumbIssuer/api/json', { headers }));

// `/job/node-pull-request/build` expects a multipart
// request, so use FormData to generate the payload
const payload = new FormData();
payload.append('json', JSON.stringify({
  parameter: [
    { name: 'CERTIFY_SAFE', value: 'on' },
    { name: 'TARGET_GITHUB_ORG', value: 'nodejs' },
    { name: 'TARGET_REPO_NAME', value: 'node' },
    { name: 'PR_ID', value: 123456 },
    { name: 'REBASE_ONTO', value: '<pr base branch>' },
    { name: 'DESCRIPTION_SETTER_DESCRIPTION', value: '' }
  ]
}));

// Send POST with breadcrump + auth on headers, and 
// FormData payload on body
await fetch('https://ci.nodejs.org/job/node-pull-request/build', {
  method: 'POST',
  headers: {
    'Jenkins-Crumb': crumb,
    ...headers
  },
  body: payload
});
```

codebytere:
Going to merge this just so we can move forward with it :) I think if there are any changes to the flow start we can update as needed!

mmarchini:
Can we get a new release with this included? :)

@nodejs/node-core-utils

codebytere:
@mmarchini https://github.com/nodejs/node-core-utils/releases/tag/v1.23.0 🌟

mmarchini:
found a bug which doesn't prevent from using the tool, but if the user doens't have permission to start CI it will show as if everything worked fine. I already have a fix locally and will send a PR soon.

PR #455, opened 2020-07-10
feat: allow citgm job comparison

codebytere:
Follow-up to https://github.com/nodejs/node-core-utils/pull/454.
Closes https://github.com/nodejs/node-core-utils/issues/352.

Extends `ncu-ci cigtm <jobid>` to `ncu-cu cigtm <jobid> [jobid2]`, allowing for direct comparisons between two CITGM runs to see if new failures were introduced.

I also did a bit of a refactor here, since there were a million classes in the result parser file and it was getting wildly unwieldy. I tried to do the minimum necessary to make this work but I plan to follow it up with the remainder of the the class split-out work.
 
Display:
```sh
Developer ❯ ncu-ci citgm 2392 2390                                             11:06PM
--------------------------------------------------------------------------------
[1/1] Running CITGM: 2392
--------------------------------------------------------------------------------
✔  Summary data downloaded
✔  Results data downloaded
✔  Summary data downloaded
✔  Results data downloaded
----------------------------------- Summary ------------------------------------
Result     FAILURE
URL        https://ci.nodejs.org/job/citgm-smoker/2392/
Source     https://api.github.com/repos/nodejs/node/git/refs/heads/v12.x
Commit     [feed95cd4c2c] Working on v12.18.1
Date       2020-06-02 20:27:47 +0200
Author     Michaël Zasso <targos@protonmail.com>
----------------------------------- Summary ------------------------------------
Result     FAILURE
URL        https://ci.nodejs.org/job/citgm-smoker/2390/
Source     https://github.com/nodejs/node/pull/33811/
Commit     [9a60117875dd] 2020-06-16, Version 12.18.1 'Erbium' (LTS)
Date       2020-06-09 20:23:09 -0700
Author     Shelley Vohr <shelley.vohr@gmail.com>
----------------------------------- Results ------------------------------------
┌────────────────────────┬───────────────────────────┬────────────────────────────┬─────────────────────────┐
│        (index)         │             0             │             1              │            2            │
├────────────────────────┼───────────────────────────┼────────────────────────────┼─────────────────────────┤
│       debian9-64       │ 'express-session-v1.17.1' │ 'yeoman-generator-v4.10.1' │                         │
│     centos7-ppcle      │    'ember-cli-v3.18.0'    │      'multer-v1.4.2'       │ 'torrent-stream-v1.2.0' │
│      rhel7-s390x       │  'torrent-stream-v1.2.0'  │                            │                         │
[...]

PR #464, opened 2020-08-04
chore: split CI builds to new files

codebytere:
Follow-up to https://github.com/nodejs/node-core-utils/pull/455:

> I also did a bit of a refactor here, since there were a million classes in the result parser file and it was getting wildly unwieldy. I tried to do the minimum necessary to make this work but I plan to follow it up with the remainder of the the class split-out work.

This is the remainder of that work - the file was inscrutably long and so i felt this would be more maintainable in this form. Specs are passing, but please take a good look if you can to make sure i didn't whiff on anything!

cc @nodejs/node-core-utils


priyank-p:
Not saying this needs to be done this way, but I think doing this refactor one or couple of class per commit would have made reviewing this way easier. Less changes to look at per commit means it is easier to spot a mistake :)

On the contrary, if you were missing an import of a class then eslint would have complained but doing it that way is still better.

Issue #466, opened 2020-08-07
Querying Lite-CI API will stuck `git node land`


mmarchini:
![stuck-lite-ci](https://user-images.githubusercontent.com/4048656/89604493-3bad7d00-d820-11ea-8942-110d2ac25556.gif)

We should probably remove checks for Lite-CI since we don't use it anymore.

codebytere:
@mmarchini happy to take it unless you're already on it!

PR #468, opened 2020-08-08
chore(land): remove lite-ci checks

mmarchini:
Lite CI is not used anymore, but some PRs got Lite-CI runs after some
changes to the bot. node-core-utils Lite-CI parsing is not working
though, therefore ncu will fail to land those PRs. Removing Lite-CI
checks fix the issue and also reduces complexity.

Fix: https://github.com/nodejs/node-core-utils/issues/466

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -21,7 +21,6 @@ const {
   JobParser,
   CI_TYPES,
   CI_PROVIDERS,
-  isLiteCI,
   isFullCI
 } = require('./ci/ci_type_parser');
 const { PRBuild } = require('./ci/ci_result_parser');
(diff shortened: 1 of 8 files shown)


mmarchini:
cc @nodejs/node-core-utils 

codecov is complaining that we reduced coverage on the changes, which is n/a since this PR just removes code, so we should be good to land with some approvals :)

lib/pr_checker.js
```diff
@@ -199,9 +198,9 @@ class PRChecker {
     }
   }
 
-  hasFullOrLiteCI(ciMap) {
+  hasFullCI(ciMap) {
     const cis = [...ciMap.keys()];
-    return cis.find(ci => isFullCI(ci) || isLiteCI(ci));
+    return cis.find(isFullCI);
```
lundibundi, line 203: Nit: could this be `cis.some()`?

mmarchini: Both work and both make sense, I don't think there's an upside to changing it

Issue #324, opened 2019-02-07
git-node: detect CI status also from the GitHub API


joyeecheung:
Currently it only parses the thread and looks for CI links pasted manually, but it can also just query the GitHub API for the status of the last commit in the PR in case someone forgets to paste the link.

We cannot replace the link parsing entirely with this though because the github bot does not always update the CI status reliably. If the node-test-commit status of the last commit is not green we could also print a warning (again we cannot just error because the bot or Jenkins is not always reliable)

sam-github:
I think @richardlau made a good point in https://github.com/nodejs/node/pull/25861#issuecomment-461256551 that having `ci: ` is good for after the fact observation.

The suggestion I would make for `git node land` is that when it says no CI detected, it mention specifically that it is looking for a `ci:` comment. You could print a message saying "ci passed, but not comment, please add `ci: ...` as a comment.... but honestly, just clarity about what the tool is looking for would be enough, IMO.

joyeecheung:
@sam-github That's a good point. Although it would still be useful to query the CI status and display a hint if it's not green or yellow, so that the user is aware that they are pushing a commit to master that did not fully pass the CI (even when it's just flaked).

PR #469, opened 2020-08-09
feat: check Actions and handle doc-only changes

mmarchini:
Doc-only changes don't need a full Jenkins CI, instead we can check
if the last Actions run was successful. Therefore this commit also adds
check for Action runs. Jenkins CI messages were improved as well.

Fix: https://github.com/nodejs/node-core-utils/issues/324
Ref: https://github.com/nodejs/node/issues/32335

### DIFF
--- a/lib/ci/ci_type_parser.js
+++ b/lib/ci/ci_type_parser.js
@@ -25,8 +25,8 @@ const CI_TYPE_ENUM = {
 };
 
 const CI_PROVIDERS = {
-  JENKINS: 'jenkins',
-  GITHUB: 'github-check'
+  GITHUB: 'github-check',
+  NODEJS: 'nodejs'
 };
 
 const { JOB_CI, FULL_CI } = CI_TYPE_ENUM;
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -204,7 +204,7 @@ class PRChecker {
   }
 
   async checkCI() {
-    const ciType = this.argv.ciType || CI_PROVIDERS.JENKINS;
+    const ciType = this.argv.ciType || CI_PROVIDERS.NODEJS;
     const providers = Object.values(CI_PROVIDERS);
 
     if (!providers.includes(ciType)) {
@@ -214,8 +214,8 @@ class PRChecker {
     }
 
     let status = false;
-    if (ciType === CI_PROVIDERS.JENKINS) {
-      status = await this.checkJenkinsCI();
+    if (ciType === CI_PROVIDERS.NODEJS) {
+      status = await this.checkNodejsCI();
     } else if (ciType === CI_PROVIDERS.GITHUB) {
       status = this.checkGitHubCI();
     }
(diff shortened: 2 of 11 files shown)


mmarchini:
Waiting for https://github.com/nodejs/node-core-utils/pull/468 before landing this. Also need to add testing for the new scenarions introduced and change name of some existing tests to reflect their current results.

mmarchini:
cc @nodejs/node-core-utils

lib/pr_checker.js
```diff
@@ -292,17 +292,61 @@ class PRChecker {
         cli.error(
           `${failures.length} failure(s) on the last Jenkins CI run`);
         status = false;
+      // NOTE(mmarchini): not sure why PEDING returns null
```
targos, line 295:
```suggestion
      // NOTE(mmarchini): not sure why PENDING returns null
```

lib/ci/ci_type_parser.js
```diff
@@ -25,8 +25,8 @@ const CI_TYPE_ENUM = {
 };
 
 const CI_PROVIDERS = {
-  JENKINS: 'jenkins',
```
codebytere, line 28: what motivated the name change? i'm not sure it'd be clear to me what "nodejs" CI meant 🤔

mmarchini: It doesn't test only `jenkins` anymore, so a new type made more sense. And since the only repo we use ncu and Jenkins is `nodejs/node`, I thought a provider specific for that repo made sense. Happy to hear other suggestions for names though.

An alternative I thought was accepting a list of providers, so for `nodejs/node` we would have `['jenkins', 'actions']`. But that increases complexity.

Although now I realized that maybe libuv also uses ncu? They definitely generate metadata and have a similar landing process.

mmarchini:
cc @nodejs/libuv do you use `node-core-utils` to land Pull Requests in `libuv/libuv`? Trying to determine if these changes will impact other projects.

cjihrig:
> cc @nodejs/libuv do you use `node-core-utils` to land Pull Requests in `libuv/libuv`? Trying to determine if these changes will impact other projects.

I don't. I'd like to, but never got around to trying to make it work. I don't believe others use it either, but I don't want to speak for them.

richardlau:
I don't think `node-core-utils` can land anything outside of `nodejs/node` without workarounds? For example it assumes certain things are in the project's README such as a TSC section and email addresses.

In the past I have used `git node metadata` pointed at the Node.js README to generate the "Reviewed-by:" metadata for libuv commits, but landed manually.

mmarchini:
> I don't think node-core-utils can land anything outside of nodejs/node without workarounds

I made it _mostly_ work for llnode (had to send a few PRs here and use Node.js README, which is inaccurate for llnode), but lately I've been thinking it's better to keep this tool focused on `nodejs/node` instead of making it configurable for any project.

lib/ci/ci_type_parser.js
mmarchini: Also FWIW, we stopped using ncu on llnode, which is what motivated me to have more than one provider here. We can probably get rid of this.

mmarchini:
cc @nodejs/node-core-utils

PR #473, opened 2020-08-14
feat: implement autorebase for PRs with multiple commits

lundibundi:
This basically adds `--autorebase` that will run interactive rebase with `--autosquash` option to handle `!fixup` commits (but refuse to land `squash!` as those require msg editing) and `git-node land --amend` for each commit.

Hopefully this will allow us to land PRs with multiple commits in `commit-queue`.

Couldn't find a way to write tests for landing, did I miss it somewhere?
Also, anyone knows a better way to run `git-land` without the need for `shell: true` option?

/cc @nodejs/node-core-utils

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -16,15 +16,17 @@ const { shortSha } = require('./utils');
 const isWindows = process.platform === 'win32';
 
 class LandingSession extends Session {
-  constructor(cli, req, dir, prid, backport) {
+  constructor(cli, req, dir, prid, backport, autorebase) {
     super(cli, dir, prid);
     this.req = req;
     this.backport = backport;
+    this.autorebase = autorebase;
   }
 
   get argv() {
     const args = super.argv;
     args.backport = this.backport;
+    args.autorebase = this.autorebase;
     return args;
   }
 
@@ -131,6 +133,17 @@ class LandingSession extends Session {
     return command;
   }
 
+  makeRebaseSuggestion(subjects) {
+    const suggestion = this.getRebaseSuggestion(subjects);
+    this.cli.log('Please run the following commands to complete landing\n\n' +
+      `$ ${suggestion}\n` +
+      '$ git node land --continue');
+  }
+
+  canAutomaticallyRebase(subjects) {
+    return subjects.every(line => !line.startsWith('squash!'));
+  }
+
   async validateLint() {
     // The linter is currently only run on non-Windows platforms.
     if (os.platform() === 'win32') {
@@ -168,14 +181,34 @@ class LandingSession extends Session {
       }
 
       return this.final();
+    } else if (this.autorebase && this.canAutomaticallyRebase(subjects)) {
+      // Run git rebase in interactive mode with autosquash but without editor
+      // so that it will perform everything automatically.
+      cli.log(`There are ${subjects.length} commits in the PR. ` +
+        'Attempting autorebase.');
+      const { upstream, branch } = this;
+      const assumeYes = this.cli.assumeYes ? '--yes' : '';
+      const msgAmend = `-x "git node land --amend ${assumeYes}"`;
+      try {
+        await forceRunAsync('git',
+          ['rebase', `${upstream}/${branch}`, '-i', '--autosquash', msgAmend],
+          {
+            ignoreFailure: false,
+            spawnArgs: {
+              shell: true,
+              env: { ...process.env, GIT_SEQUENCE_EDITOR: ':' }
+            }
+          });
+        return this.final();
+      } catch (e) {
+        await runAsync('git', ['rebase', '--abort']);
+        const count = subjects.length;
+        cli.log(`Couldn't rebase ${count} commits in the PR automatically`);
+        this.makeRebaseSuggestion(subjects);
+      }
+    } else {
+      this.makeRebaseSuggestion(subjects);
     }
-
-    const suggestion = this.getRebaseSuggestion(subjects);
-
-    cli.log(`There are ${subjects.length} commits in the PR`);
-    cli.log('Please run the following commands to complete landing\n\n' +
-            `$ ${suggestion}\n` +
-            '$ git node land --continue');
   }
 
   async apply() {
--- a/components/git/land.js
+++ b/components/git/land.js
@@ -47,6 +47,11 @@ const landOptions = {
     describe: 'Prevent adding Fixes and Refs information to commit metadata',
     default: false,
     type: 'boolean'
+  },
+  autorebase: {
+    describe: 'Automatically rebase branches with multiple commits',
+    default: false,
+    type: 'boolean'
   }
 };
 
@@ -165,7 +170,8 @@ async function main(state, argv, cli, req, dir) {
       cli.log('run `git node land --abort` before starting a new session');
       return;
     }
-    session = new LandingSession(cli, req, dir, argv.prid, argv.backport);
+    session = new LandingSession(cli, req, dir, argv.prid, argv.backport,
+      argv.autorebase);
     const metadata = await getMetadata(session.argv, argv.skipRefs, cli);
     if (argv.backport) {
       const split = metadata.metadata.split('\n')[0];


lib/landing_session.js
```diff
@@ -130,6 +132,10 @@ class LandingSession extends Session {
     return command;
   }
 
+  canAutomaticallyRebase(subjects) {
+    return subjects.every(line => !line.startsWith('squash!'));
```
mmarchini, line 136: I suggest we don't check the commits, `fixup!` as a prefix is also valid, and if all commits are valid we still want to autorebase. The last ncu check will catch any commit not following the guidelines, so it should be safe to skip this check here

lib/landing_session.js
```diff
@@ -155,14 +161,28 @@ class LandingSession extends Session {
         return;
       }
       return this.final();
+    } else if (this.autorebase && this.canAutomaticallyRebase(subjects)) {
+      // Run git rebase in interactive mode with autosquash but without editor
+      // so that it will perform everything automatically.
+      cli.log(`There are ${subjects.length} commits in the PR. ` +
+        'Attempring autorebase.');
```
mmarchini, line 168: s/attemprimg/attempting/

lib/landing_session.js
```diff
@@ -118,7 +120,7 @@ class LandingSession extends Session {
   getRebaseSuggestion(subjects) {
     const { upstream, branch } = this;
     let command = `git rebase ${upstream}/${branch} -i`;
-    command += ' -x "git node land --amend"';
+    command += ' -x "git-node land --amend"';
```
mmarchini, line 123: Keep `git node`

lib/landing_session.js
mmarchini: Ahh, I see, it's skipping squash. Should be fine for a first version, but we want to handle squash as well

lundibundi: > fixup! as a prefix is also valid,

Yep, that's why it is not here.

My main problem with allowing `squash!` here is that it will just throw every commit-message and commit-title together and the result is unlikely to be what the user wants. This is fine for manual landing (we can always change the message ourselves) but is not compatible with `commit-queue` as it will open an editor for commit-message editing.
Should we perhaps add another option for auto-landing where this check will be performed?

lib/landing_session.js
lundibundi: Sure, but since the command is now different why do we keep suggesting the old one? Or am I missing something?

lib/landing_session.js
lundibundi: Should we change the option to 
`--rebase=auto` -- skip squash!
`--rebase` || `--rebase=all` - allow squash, the default

?

lib/landing_session.js
mmarchini: The command shouldn't be different, git will turn any executable named `git-<command>` into a git command (`git <command>`). Did you get any errors trying to run `git node` (when installing with npm, because when developing `git node` won't work because the executable won't be in the path)?

lib/landing_session.js
mmarchini: I think we should programmatically change squash with fixup before (or while) running rebase, squash is never what users want on nodejs/node. But let's land this as is and we can improve in a follow up, this is already way better than what we have today. Thank you for working on it!

richardlau: FWIW I don't think we should assume squash==fixup. I would be okay treating squash! commits as not landable via commit queue.

mmarchini: Can you elaborate on why we shouldn't (for nodejs/node)? `squash!` will always be invalid there.

richardlau: I agree `squash!` would be invalid there. If it is used in a PR I would want commit queue to treat that as invalid and not second guess that it is to be treated as fixup.

mmarchini: I suggested replacing squash with fixup because I want to minimize the cases where we leave the burden to collaborators (having to go to the terminal to land a PR is a very heavyweight process). If `commit-queue` doesn't handle squash it might increase the burden on collaborators. We could disallow squash as part of https://github.com/nodejs/node/pull/32417, but then we're shifting the burden to the author of the PR (which again, could be avoided with automation). IMO this is a case where absolute perfect correctness is less desirable than improved user experience. As a compromise, the linter could _warn_ the user that squashes will be converter into fixups (instead of asking users to open the terminal, checkout the code, rebase changing the commits, and force push).

richardlau: We should probably take this conversation outside of this PR 😄.

lundibundi: I completely agree with @richardlau that assuming `squash! == fixup!` is not the right choice in general since they have a different meaning for both git and a user and the person writing `squash!` may want to combine/change the messages later.

@mmarchini I understand your point and looking at the PRs into Node.js people usually use `squash!` as a `fixup!` so it makes sense to just replace them but in a rare case where someone wants to change the message it's better to just ask them to do so manually and either change to `fixup!` or squash everything the way they want and force-push to the branch instead of second-guessing in the automation. (should we perhaps add a separate label for `fixup-everrything-to-first-commit`?)
IMO changing something in your branch is not that much of a problem and is done easily. Landing PRs to master, on the other hand, is what introduces the most burden.

mmarchini: > IMO changing something in your branch is not that much of a problem and is done easily. Landing PRs to master, on the other hand, is what introduces the most burden.

That's a fair assumption once you get used to it, but not all new contributors are familiar enough with git for this task to be considered easy. Even existing collaborators might find it hard or a burden (note that burden !== easy in this case, something can be easy and still impose a burden barrier).

Maybe we should add a `manual-land` label that prevents the commit queue from attempting to land, in cases where the author or collaborator landing wants to change something before landing.

I still disagree on not assuming `squash == fixup` but that can be discussed outside this PR.

mmarchini: fyi I didn't land this yet because I want to test against a PR or two first, but none of the discussion above is blocking.

lundibundi: Sure, I still need to fix the nits =)

> Even existing collaborators might find it hard or a burden (note that burden !== easy in this case, something can be easy and still impose a burden barrier).

Yeah, that's what I thought when suggested the `fixup-everything-to-first-commit`. It would be even better to be able to just specify a commit message in a PR-comment, though I think it's currently not favored due to notifications.

> Maybe we should add a manual-land label that prevents the commit queue from attempting to land, in cases where the author or collaborator landing wants to change something before landing.

+1

mmarchini:
Humm, just tried to use this and it ended up rather confusing...

```console
$ git node land --autorebase 34671
✔  Done loading data for nodejs/node/pull/34671
----------------------------------- PR info ------------------------------------
Title      errors: improve ERR_INVALID_OPT_VALUE message (#34671)
   ⚠  Could not retrieve the email or name of the PR author's from user's GitHub profile!
Branch     lundibundi:improve-errors-opt-value -> nodejs:master
Labels     author ready, child_process, errors, net
Commits    2
 - errors: improve ERR_INVALID_OPT_VALUE error
 - fixup! errors: improve ERR_INVALID_OPT_VALUE error
Committers 1
 - Denys Otrishko <shishugi@gmail.com>
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/34671
Reviewed-By: James M Snell <jasnell@gmail.com>
Reviewed-By: Rich Trott <rtrott@gmail.com>
Reviewed-By: Mary Marchini <oss@mmarchini.me>
   ℹ  Last Full PR CI on 2020-08-14T11:29:16Z: https://ci.nodejs.org/job/node-test-pull-request/32788/
✔  Build data downloaded
   ℹ  This PR was created on Fri, 07 Aug 2020 19:50:06 GMT
   ✔  Approvals: 3
   ✔  - James M Snell (@jasnell) (TSC): https://github.com/nodejs/node/pull/34671#pullrequestreview-463577321
   ✔  - Rich Trott (@Trott) (TSC): https://github.com/nodejs/node/pull/34671#pullrequestreview-463906392
   ✔  - Mary Marchini (@mmarchini) (TSC): https://github.com/nodejs/node/pull/34671#pullrequestreview-467939143
? This PR should be ready to land, do you want to continue? Yes
   ✔  No git am in progress
   ✔  No git rebase in progress
? Do you want to try reset the local master branch to upstream/master? Yes
⠴ Bringing upstream/master up to date...From github.com:nodejs/node
 * branch                  master     -> FETCH_HEAD
✔  upstream/master is now up-to-date
✔  Downloaded patch to /home/mmarchini/workspace/nodejs/node-master/.ncu/34671/patch
Applying: errors: improve ERR_INVALID_OPT_VALUE error
error: patch failed: lib/internal/errors.js:1114
error: lib/internal/errors.js: patch does not apply
Patch failed at 0001 errors: improve ERR_INVALID_OPT_VALUE error
hint: Use 'git am --show-current-patch' to see the failed patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
? The normal `git am` failed. Do you want to retry with 3-way merge? Yes
Applying: errors: improve ERR_INVALID_OPT_VALUE error
Using index info to reconstruct a base tree...
M       lib/internal/errors.js
M       lib/net.js
Falling back to patching base and 3-way merge...
Auto-merging lib/net.js
Auto-merging lib/internal/errors.js
Applying: fixup! errors: improve ERR_INVALID_OPT_VALUE error
   ✔  Patches applied
There are 2 commits in the PR. Attempring autorebase.
Executing: git-node land --amend
--------------------------------- New Message ----------------------------------
errors: improve ERR_INVALID_OPT_VALUE error

* use util.inspect for value presentation
* allow to optionally specify error reason

PR-URL: https://github.com/nodejs/node/pull/34671
Reviewed-By: James M Snell <jasnell@gmail.com>
Reviewed-By: Rich Trott <rtrott@gmail.com>
Reviewed-By: Mary Marchini <oss@mmarchini.me>
? Use this message? Yes
[detached HEAD d6c25a1a19] errors: improve ERR_INVALID_OPT_VALUE error
 Author: Denys Otrishko <shishugi@gmail.com>
 Date: Fri Aug 7 20:31:20 2020 +0300
 9 files changed, 45 insertions(+), 26 deletions(-)
Successfully rebased and updated refs/heads/master.
```

After the rebase, it doesn't give any instructions on how to advance (`git node land --continue`). Also, it doesn't respect `--yes` after the rebase, which prevents us from using it on `commit-queue`. I'll add some inline comments with suggestions.

lib/landing_session.js
```diff
@@ -155,14 +161,28 @@ class LandingSession extends Session {
         return;
       }
       return this.final();
+    } else if (this.autorebase && this.canAutomaticallyRebase(subjects)) {
+      // Run git rebase in interactive mode with autosquash but without editor
+      // so that it will perform everything automatically.
+      cli.log(`There are ${subjects.length} commits in the PR. ` +
+        'Attempting autorebase.');
+      const { upstream, branch } = this;
+      const msgAmend = '-x "git-node land --amend"';
```
mmarchini, line 170: This fixes the interaction with --yes:

```suggestion
      const msgAmend = 
          `-x "git-node land --amend ${cli.assumeYes ? '--yes' : ''}"`;
```

mmarchini:
Ok, this should do it:

```diff
diff --git a/lib/landing_session.js b/lib/landing_session.js
index 0f239cb..ec06e55 100644
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -7,7 +7,7 @@ const {
 } = require('./deprecations');
 
 const {
-  runAsync, runSync, forceRunAsync
+  runAsync, runSync, forceRunAsync, runAsyncBase
 } = require('./run');
 const Session = require('./session');
 const { shortSha } = require('./utils');
@@ -167,15 +167,33 @@ class LandingSession extends Session {
       cli.log(`There are ${subjects.length} commits in the PR. ` +
         'Attempring autorebase.');
       const { upstream, branch } = this;
-      const msgAmend = '-x "git-node land --amend"';
-      await runAsync('git',
-        ['rebase', `${upstream}/${branch}`, '-i', '--autosquash', msgAmend],
-        {
-          spawnArgs: {
-            shell: true,
-            env: { ...process.env, GIT_SEQUENCE_EDITOR: ':' }
-          }
-        });
+      const msgAmend =
+          `-x "git-node land --amend ${cli.assumeYes ? '--yes' : ''}"`;
+      try {
+        await runAsyncBase('git',
+          ['rebase', `${upstream}/${branch}`, '-i', '--autosquash', msgAmend],
+          {
+            spawnArgs: {
+              shell: true,
+              env: { ...process.env, GIT_SEQUENCE_EDITOR: ':' }
+            }
+          });
+        return this.final();
+      } catch {
+        await runAsync('git',
+          ['rebase', '--abort'],
+          {
+            spawnArgs: {
+              shell: true,
+              env: { ...process.env, GIT_SEQUENCE_EDITOR: ':' }
+            }
+          });
+        const suggestion = this.getRebaseSuggestion(subjects);
+        cli.log(`Couldn't rebase ${subjects.length} commits in the PR automatically`);
+        cli.log('Please run the following commands to complete landing\n\n' +
+                `$ ${suggestion}\n` +
+                '$ git-node land --continue');
+      }
     } else {
       const suggestion = this.getRebaseSuggestion(subjects);
       cli.log(`There are ${subjects.length} commits in the PR`);
diff --git a/lib/run.js b/lib/run.js
index 63ae10d..8bebc16 100644
--- a/lib/run.js
+++ b/lib/run.js
@@ -26,6 +26,8 @@ function runAsyncBase(cmd, args, options = {}) {
   });
 }
 
+exports.runAsyncBase = runAsyncBase;
+
 exports.forceRunAsync = function(cmd, args, options) {
   return runAsyncBase(cmd, args, options).catch((error) => {
     if (error.message !== IGNORE) {
```

It fixes interaction with --yes, calls `final()` if rebase works and aborts the rebase + gives instructions if rebase fails (couldn't test this last one).

lib/landing_session.js
lundibundi: Oh, that's where it was, thanks.

lundibundi:
Didn't get to check the `git rebase --abort` yet (git am is kinda confusing =), still trying to figure out how to check landing on random branches and not real PRs in node). 
Removed `shell: true`, thanks @mmarchini.

After implementing it this way, I'm thinking whether we should just ask (via cli.prompt) whether to make a rebase instead of making it an option?

mmarchini:
> still trying to figure out how to check landing on random branches and not real PRs in node).

You can open test PRs to nodejs/node-auto-test :)

> After implementing it this way, I'm thinking whether we should just ask (via cli.prompt) whether to make a rebase instead of making it an option

I think for this version we can keep as a flag and in the future when we're more confident it works as expected we can change to a question

lundibundi:
~~Removed shell: true, thanks @mmarchini.~~
Ok, had a stale ncu installed, without shell: true the git still cannot find git node
"git node land --amend ": git node land --amend : command not found

On the other hand got to check the `git rebase --abort`, IIUC we have to use `forceRunAsync` because a normal `runAsync` will call `process.exit(1)` which not what we want.

mmarchini:
Yes, in the diff I suggested exporting runAsyncBase and using it, but if force works I'm happy with it too.

> Ok, had a stale ncu installed, without shell: true the git still cannot find git node
"git node land --amend ": git node land --amend : command not found

To run the command it's fine to use `git-node` (that might be how we do it elsewhere), just when printing the suggestion it is preferable to print the same command we use on the documentation and help messages (which is `git node`)

lundibundi:
Rebased to fix conflicts. This is ready for review.

mmarchini:
Code LGTM. Started testing but couldn't find any PR on nodejs/node to test this, and when I tried on nodejs/node-auto-test I hit some issues. Will try again later or tomorrow.

mmarchini:
Ok, it's hard to find a PR that passes all checks _and_ has a properly formatted `fixup!` commit. I'm in favor of landing this and tweak it as issues arise.

richardlau:
> 
> 
> Ok, it's hard to find a PR that passes all checks _and_ has a properly formatted `fixup!` commit. I'm in favor of landing this and tweak it as issues arise.

https://github.com/nodejs/node/pull/34798?

mmarchini:
Landing so I can rebase and land #487 as well to test with https://github.com/nodejs/node/pull/34798.

mmarchini:
It works, nice!

PR #475, opened 2020-08-15
feat(land): add skipChecks option

mmarchini:
skipChecks will allow skipping certain checks. Useful to skip "commit
after last approval" on commit queue.

Ref: https://github.com/nodejs/node/issues/34770

### DIFF
--- a/components/git/land.js
+++ b/components/git/land.js
@@ -4,10 +4,13 @@ const { parsePRFromURL } = require('../../lib/links');
 const getMetadata = require('../metadata');
 const CLI = require('../../lib/cli');
 const Request = require('../../lib/request');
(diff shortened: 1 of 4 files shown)


mmarchini:
I'm not 100% sure we're going to use this, but I guess it doesn't hurt to have it either.

components/git/land.js
```diff
@@ -42,6 +45,21 @@ const landOptions = {
     describe: 'Land a backport PR onto a staging branch',
     default: false,
     type: 'boolean'
+  },
+  skipChecks: {
```
codebytere, line 49: We should add this option to documentation as well, perhaps with an accompanying example :)

codebytere:
@mmarchini could you rebase this?

Issue #268, opened 2018-08-06
$XDG_CONFIG_HOME should default to ~/.config


ChALkeR:
Ref: https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

> `$XDG_CONFIG_HOME` defines the base directory relative to which user specific configuration files should be stored. If `$XDG_CONFIG_HOME` is either not set or empty, a default equal to `$HOME/.config` should be used.

Atm, that does not work — my `XDG_CONFIG_HOME` is not overriden (i.e. is unset), I have a config at `~/.config/ncurc`, and `git-node` still complains.

ChALkeR:
`require('os').homedir()` aready follows `$HOME`, so the config location should set to `${require('os').homedir()}/.config` if `process.env.XDG_CONFIG_HOME` is not set or is empty.

PR #479, opened 2020-08-16
fix: check $HOME/config on linux

codebytere:
Closes https://github.com/nodejs/node-core-utils/issues/268.

Ensures that on Linux, `.ncurc` is placed in either `XDG_CONFIG_HOME` or `$HOME/config` and cleans up exports a bit to match style in other parts of the project.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -8,33 +8,27 @@ const GLOBAL_CONFIG = Symbol('globalConfig');
 const PROJECT_CONFIG = Symbol('projectConfig');
 const LOCAL_CONFIG = Symbol('localConfig');
 
-exports.GLOBAL_CONFIG = GLOBAL_CONFIG;
-exports.PROJECT_CONFIG = PROJECT_CONFIG;
-exports.LOCAL_CONFIG = LOCAL_CONFIG;
-
 function getNcurcPath() {
-  if (process.env.XDG_CONFIG_HOME !== 'undefined' &&
-      process.env.XDG_CONFIG_HOME !== undefined) {
-    return path.join(process.env.XDG_CONFIG_HOME, 'ncurc');
-  } else {
-    return path.join(os.homedir(), '.ncurc');
+  let configHome = os.homedir();
+  if (process.platform === 'linux') {
+    configHome = process.env.XDG_CONFIG_HOME || `${os.homedir()}/config`;
   }
+  return path.join(configHome, '.ncurc');
 }
-exports.getNcurcPath = getNcurcPath;
 
-exports.getMergedConfig = function(dir, home) {
-  const globalConfig = exports.getConfig(GLOBAL_CONFIG, home);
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -8,33 +8,27 @@ const GLOBAL_CONFIG = Symbol('globalConfig');
...
-
 function getNcurcPath() {
-  if (process.env.XDG_CONFIG_HOME !== 'undefined' &&
-      process.env.XDG_CONFIG_HOME !== undefined) {
-    return path.join(process.env.XDG_CONFIG_HOME, 'ncurc');
-  } else {
-    return path.join(os.homedir(), '.ncurc');
+  let configHome = os.homedir();
+  if (process.platform === 'linux') {
+    configHome = process.env.XDG_CONFIG_HOME || `${os.homedir()}/config`;
```
mmarchini, line 14: I think it should be `${os.homedir()}/.config`

codebytere: whoops yes you're right! also need to figure out how to update the one failing test accordingly, since it's not really testing for the right signal anymore 🤔

codebytere: think that should do it

lib/config.js
```diff
@@ -8,33 +8,27 @@ const GLOBAL_CONFIG = Symbol('globalConfig');
 const PROJECT_CONFIG = Symbol('projectConfig');
 const LOCAL_CONFIG = Symbol('localConfig');
 
-exports.GLOBAL_CONFIG = GLOBAL_CONFIG;
-exports.PROJECT_CONFIG = PROJECT_CONFIG;
-exports.LOCAL_CONFIG = LOCAL_CONFIG;
-
 function getNcurcPath() {
-  if (process.env.XDG_CONFIG_HOME !== 'undefined' &&
```
joyeecheung, line 16: I don't quite remember why we have this condition now...but I guess it doesn't hurt to keep it?

lib/config.js
```diff
@@ -8,33 +8,27 @@ const GLOBAL_CONFIG = Symbol('globalConfig');
...
-      process.env.XDG_CONFIG_HOME !== undefined) {
-    return path.join(process.env.XDG_CONFIG_HOME, 'ncurc');
-  } else {
-    return path.join(os.homedir(), '.ncurc');
+  let configHome = os.homedir();
+  if (process.platform === 'linux') {
+    configHome = process.env.XDG_CONFIG_HOME || `${os.homedir()}/.config`;
```
joyeecheung, line 14: I guess this would be a breaking change..maybe it would be safer to check whether `$HOME/.ncurc` exists and emit a warning if it does (also falling back to it) before completely moving to `$HOME/.config/.ncurc` otherwise people might be surprised why their config suddenly stopped working

codebytere: good point - will do!

Issue #233, opened 2018-04-08
get-metadata: false-positive 'Fixes:'


vsemozhetbyt:
PR: https://github.com/nodejs/node/pull/19832
```
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/19832
Fixes: https://github.com/Fix
Reviewed-By: James M Snell <jasnell@gmail.com>
Reviewed-By: Trivikram Kamat <trivikr.dev@gmail.com>
Reviewed-By: Gus Caplan <me@gus.host>
--------------------------------------------------------------------------------
```
Maybe we should check if these lines do have an URL.

PR #485, opened 2020-08-19
fix: potential non-url Fixes being added to metadata

codebytere:
Closes https://github.com/nodejs/node-core-utils/issues/233.

Filters out anything from being added to `Fixes` metadata that does not look e.g like:

Fixes #16
Fixes nodejs/build#1961
Fixes https://github.com/nodejs/node/issue/12345


richardlau:
Would it still allow URLs?

joyeecheung:
From what I can tell this would make links like `Fixes: https://github.com/nodejs/node/issue/12345` invalid?

codebytere:
Missed a spot in the regex - should be fixed up now!

mmarchini:
The goal lgtm, would need more time to review the regex (if others think the regex is fine I'm good with it too!)

richardlau:
Tests would be good. Particularly one showing an example of a non-url Fixes being ignored and one showing a Fixes with a full URL is accepted.

PR #486, opened 2020-08-19
chore(git-node): avoid dealing with patch files for landing

addaleax:
Working with patch files is, on the one hand, unclean because it
requires manual parsing to figure out e.g. the commit message subjects,
and on the other hand, poses actual problems when dealing with binary
files.

Clean the code up by fetching the original commits themselves and
applying those commits directly.

This also resolves some TODO comments, e.g. we can now check whether
the commits that are to be applied match the ones reported through
the GitHub API.

---

I’m not sure what kind of testing this patch needs? I’ve tried it out manually on one PR so far, but there doesn’t appear to be anything automated in place here.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -30,6 +30,7 @@ class LandingSession extends Session {
     this.lint = lint;
     this.autorebase = autorebase;
     this.fixupAll = fixupAll;
+    this.expectedCommitShas = [];
   }
 
   get argv() {
@@ -44,6 +45,8 @@ class LandingSession extends Session {
   async start(metadata) {
     const { cli } = this;
     this.startLanding();
+    this.expectedCommitShas =
+      metadata.data.commits.map(({ commit }) => commit.oid);
     const status = metadata.status ? 'should be ready' : 'is not ready';
     // NOTE(mmarchini): default answer is yes. If --yes is given, we need to be
     // more careful though, and we change the default to the result of our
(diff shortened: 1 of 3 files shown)


mmarchini:
Thank you! I always struggled to understand why we used patches instead of fetching the PR commits.

> I’m not sure what kind of testing this patch needs? I’ve tried it out manually on one PR so far, but there doesn’t appear to be anything automated in place here.

I think most of our landing process is untested and doesn't have the testing infrastructure it needs in place, so relying on manual tests is the way to go for now.

lib/landing_session.js
```diff
@@ -65,34 +68,44 @@ class LandingSession extends Session {
...
 
-    // TODO(joyeecheung): restore previously downloaded patches
     cli.startSpinner(`Downloading patch for ${prid}`);
-    const patch = await req.text(
-      `https://github.com/${owner}/${repo}/pull/${prid}.patch`);
-    this.savePatch(patch);
-    cli.stopSpinner(`Downloaded patch to ${this.patchPath}`);
+    await runAsync('git', [
+      'fetch', `https://github.com/${owner}/${repo}.git`,
+      `refs/pull/${prid}/merge`]);
```
mmarchini, line 76: `/merge`?

lib/landing_session.js
```diff
@@ -65,34 +68,44 @@ class LandingSession extends Session {
...
+    const [base, head] = (await runAsync('git', [
+      'rev-parse', 'FETCH_HEAD^1', 'FETCH_HEAD^2'], { captureStdout: true }))
```
mmarchini, line 78: ^1, ^2, etc. in git means "prevous commit", "two commits before this commit", etc. Does this have some special meaning in this case due to us fetching the `/merge` ref (like ^1 gets the head of the merge commit, ^2 gets the base of the merge commit)? If that's the case it might be worth adding a comment explaining what's happening.

lib/landing_session.js
```diff
@@ -65,34 +68,44 @@ class LandingSession extends Session {
...
+      cli.error(`Mismatched commits:\n${mismatchedCommits}`);
+      process.exit(1);
+    }
+
+    const commitInfo = { base, head, shas: commitShas };
+    this.saveCommitInfo(commitInfo);
+
     try {
-      await forceRunAsync('git', ['am', this.patchPath], {
+      await forceRunAsync('git', ['cherry-pick', `${base}..${head}`], {
```
mmarchini, line 101: Can't we `git checkout head && git rebase` instead? What's the advantage of using cherry-pick? With rebase we could autosquash and amend by default, which would considerably simplify the workflow.

(we can do this in a future PR too, not blocking)

mmarchini:
This will conflict with #473

lib/landing_session.js
addaleax: > like ^1 gets the head of the merge commit, ^2 gets the base of the merge commit

Yes, except it’s the other way around :) I’ve added a comment.

lib/landing_session.js
addaleax: I think so – this was just closer to the original concept of using `git am` to apply the patch

If we do want to autosquash here, that would change the structure a lot more, right? We’d essentially be merging two steps into one?

addaleax:
> This will conflict with #473

Since I’ll not be involved in Node.js as much in the future, just consider this something that bugged me for quite a while and that I wanted to provide a fix for, but that I’ll probably not be involved in merging long-term. If you want to make changes, feel free to also just push to this branch :)

lib/landing_session.js
mmarchini: > If we do want to autosquash here, that would change the structure a lot more, right? We’d essentially be merging two steps into one?

Right, which is why I think it would simplify the workflow. But definitely doesn't need to be part of this PR :)

lib/run.js
```diff
@@ -21,7 +30,11 @@ function runAsyncBase(cmd, args, options = {}) {
         err.messageOnly = true;
         return reject(err);
       }
-      return resolve();
+      if (captureStdout === 'lines') {
+        stdout = stdout.split(/\r?\n/g);
```
lundibundi, line 34: Could we possibly instead run `stdout.trim().split(...)`? (to avoid the check below)

lundibundi:
Also, I'm fine with landing this first or helping merge if my PR lands first since my PR is much simpler anyway and I always find git-am very confusing when there are conflicts (like what do I change to fix them? :smile: ).

lib/landing_session.js
```diff
@@ -170,9 +174,12 @@ class LandingSession extends Session {
       return this.final();
     }
 
+    const subjects = await runAsync('git',
+      ['log', '--pretty=format:%s', `${commitInfo.base}..${commitInfo.head}`],
+      { captureStdout: 'lines' });
     const suggestion = this.getRebaseSuggestion(subjects);
 
-    cli.log(`There are ${subjects.length} commits in the PR`);
+    cli.log(`There are ${commitInfo.shas.length.length} commits in the PR`);
```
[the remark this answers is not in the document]
addaleax: Thank you :)

lib/run.js
addaleax: I actually had written that first, but there might be cases in which whitespace occurs intentionally at the beginning of the first line/end of the last line? :woman_shrugging: With `git` it’s not an issue,  I think, but this would also be used to spawn other programs

lundibundi: :thinking: that's true but I can't come up with an example where one would use `'lines'` but would require leading/trailing spaces to be preserved.
Anyway, just a suggestion/Nit.

mmarchini:
Let's land the autorebase first as it should allow the commit-queue to land some PRs it can't today. I'm happy to rebase this and fix the conflicts after we land the autorebase.

mmarchini:
@addaleax I rebased and fixed the conflicts locally, let me know if I can force-push.

addaleax:
@mmarchini You can generally do whatever you think is best on any of my branches :)

mmarchini:
Force pushed to fix conflicts :)

addaleax:
This can be merged, right? :)

mmarchini:
Yeah, I just wanted to make a release first (but I don't have npm publish rights, so I'm waiting for someone else to do it).

mmarchini:
Will try to rebase and land this weekend :)

addaleax:
Resolved another (trivial) conflict :)

mmarchini (approved):

Issue #446, opened 2020-06-27
Suggestion: auto-land and auto-release NCU Pull Requests


mmarchini:
Looking at our Pull Requests list, seems like we have six pull requests ready to land (with approvals), but I don't think it's clear who is allowed to land PRs in this repository. Similarly, sometimes commits will sit on master for a while until someone has time to publish a new version on npm. My suggestions:

1. A scheduled GitHub Action that will merge all PRs with at least one approval and green CI. This could run every 5 minutes (quickest the scheduler can run), or maybe every hour (since we don't have a big flow of PRs here).
2. A scheduled GitHub Action that will make a new release to npm every day (if we want quick releases) or every week (if we want slower releases)

We would need to identify semverness of PRs somehow, either via labels or using https://www.conventionalcommits.org/en/v1.0.0/. Looking at our logs, we're already not strictly following Node.js core commit guidelines and we have mixed Node.js and conventional commit messages.

As for npm releases, it might be tricky because we require 2FA. If anyone worked with 2FA and automated npm releases before, I'd be happy to see solutions :)

codebytere:
@mmarchini I definitely agree that there's a bit of a 911 problem going on with who has authority to merge things, but i'm not 100% sure that automatic merges are the right solution here for a few reasons:

1) Fairly few people (i think it's basically ~4-5 of us atm?) maintain this repo, so it's often the case that a 1-2 specific people's approval is needed for a given PR as they're the one who'll have the necessary context
2) CI on this repo is a bit of a struggle bus at present especially on macOS - I see it fail/flake all the time on PRs like [this one](https://github.com/nodejs/node-core-utils/pull/445/checks?check_run_id=814514129)

What do you think of perhaps adding CODEOWNERS? I think that might make ownership areas a bit more straightforward 🤔 

For the latter suggestion, I think weekly releases would be my personal preference but i'd also be ok with daily if that's the consensus from others!

https://github.com/GoogleCloudPlatform/wombat-dressing-room is the best solution to the 2FA problem i can think of off the top of my head - Electron also built a similar thing with https://docs.continuousauth.dev/ but it integrates with Slack so that's probably not ideal for our situation here.

mmarchini:
[...]

PR #488, opened 2020-08-20
chore(ci): use Actions to generate changelog

mmarchini:
`release-please` will bump the package version and generate changelogs
based on conventionalcommit. It does so by opening a PR every time a
commit lands, and it updates that PR for each subsequent commit.
Publishing to npm is still done manually.

Ref: https://github.com/nodejs/node-core-utils/issues/446


mmarchini:
cc @nodejs/node-core-utils this should offload some of the release process to Actions (we still need to publish it manually though).

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,19 @@
...
+  name: Lint
+  runs-on: ubuntu-latest
+  steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js v12
+      uses: actions/setup-node@v1
+      with:
+        node-version: v12
+    - run: npm install
+    - run: npx commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }}
```
targos, line 19: What happens if the event is a push to master?

mmarchini: Good catch, seems like there's a syntax error here as well 🤔. Lemme tweak it on my fork will update soon.

mmarchini: why reinvent the wheel right 😄

Issue #467, opened 2020-08-08
Add a verbose option


joyeecheung:
A few things that come to mind

1. Show commands spawned by ncu
2. Show HTTP request response, in particular responses that could not be parsed as JSON (e.g. when Jenkins authorization fails, it returns XML)
3. Pipe stdout and stderr, also do not use spinners if verbosity is on in case the output is overwritten

joyeecheung:
Reminder: this is not yet fixed because of 3 in the OP

PR #503, opened 2020-09-22
feat: support NCU_VERBOSITY=debug environment variable

joyeecheung:
Support NCU_VERBOSITY=debug which tells ncu to log CLI commands
and HTTP requests to the command line for debugging.

Refs: https://github.com/nodejs/node-core-utils/issues/467

I switched to using environment variables instead of options since it's easier to turn the setting on/off that way. Also I left out `Pipe stdout and stderr, also do not use spinners if verbosity is on in case the output is overwritten` mentioned in https://github.com/nodejs/node-core-utils/issues/467 in this PR since that requires a bit more refactoring.

cc @nodejs/node-core-utils 

Screen Shot 2020-09-22 at 11 41 58 AM

### DIFF
--- a/lib/verbosity.js
+++ b/lib/verbosity.js
@@ -1,0 +1,29 @@
+'use strict';
+
+const chalk = require('chalk');
+const util = require('util');
+
+const VERBOSITY = {
+  NONE: 0,
+  DEBUG: 2
+};
+
+let verbosity = VERBOSITY.NONE;
+
+exports.isDebugVerbosity = function() {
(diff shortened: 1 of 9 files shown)


lib/verbosity.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const chalk = require('chalk');
+const util = require('util');
+
+const VERBOSITTY = {
```
codebytere, line 6: nit: s/VERBOSITTY/VERBOSITY

codebytere (approved):

PR #505, opened 2020-10-03
docs: add a tip about using git-node-land behind firewall

joyeecheung:
Maybe it's worth mentioning when onboarding collaborators living in mainland China :)


joyeecheung:
cc @nodejs/node-core-utils

codebytere (approved):

Issue #496, opened 2020-09-06
ncu-ci broken


mmarchini:
```
TypeError: Cannot read property 'JS_TEST_FAILURE' of undefined
    at FailureAggregator.formatAsMarkdown (/opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/node-core-utils/lib/ci/failure_aggregator.js:85:44)
    at WalkCommand.aggregate (/opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/node-core-utils/bin/ncu-ci:437:34)
    at main (/opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/node-core-utils/bin/ncu-ci:567:24)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

https://github.com/nodejs/reliability/runs/1076696317?check_suite_focus=true

Probably broke in the last release

PR #507, opened 2020-10-07
fix: undefined failures & JSON error

codebytere:
Closes https://github.com/nodejs/node-core-utils/issues/496.

Fixes two things: 

- `FAILURE_TYPES_NAME` should be imported from `CIFailureParser`
- `this.json` should be set regardless of whether the user has chosen to pass `--json`

cc @mmarchini

### DIFF
--- a/bin/ncu-ci
+++ b/bin/ncu-ci
@@ -351,11 +351,11 @@ class CICommand {
       await build.getResults();
       build.display();
 
-      if (argv.json) {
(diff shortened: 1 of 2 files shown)


codebytere:
@nodejs/node-core-utils if someone wouldn't mind taking a quick peek before the next release that'd be great 🙇🏻‍♀️

joyeecheung (approved):

PR #518, opened 2020-11-02
chore: fix tests on Windows and enable it in actions

targos:
Use a gitattributes file to force text files to be checked out with LF
line endings. This resolves issues on Windows with tests that compare
values with the contents of files on disk.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Force checking out with LF on Windows to avoid issues in tests that read files.
+* text=auto eol=lf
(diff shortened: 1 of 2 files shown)


targos:
macOS seems flaky : 

```
  1 failing

  1) auth
       asks for auth data if no ncurc is found:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/node-core-utils/node-core-utils/test/unit/auth.test.js)
  
```

codebytere (approved):

richardlau (approved):

PR #546, opened 2021-05-10
feat: add `--lite` option for `git-node land`

DeeDeeG:
Makes it possible to use `git-node land` in repositories other than
`nodejs/node`. This is helpful for producing Node-style commit
footers, and enforcing some Node PR/commit conventions.

---

Context: Looking into whether `git-node land` can be used for landing stuff at the `node-gyp` repo. With this change, the answer would be "yes."

See: https://github.com/nodejs/node-gyp/issues/2286

cc @rvagg thoughts? (I like that `git-node land` takes some of the work out of using Node-style commit message prefixes, and maintaining commit message footers such as `Reviewed-By:` and `PR-URL:`, but dropping all that and just using `release-please` and "conventional commit"-style commit messages would be even easier.)

Note to `node-core-utils` maintainers: If `node-gyp` maintainers aren't interested in this `--lite` flag feature, then I don't think there is a real-world user for it. In that case, with no real-world users, I wouldn't recommend merging this. Basically, I can only see this being useful if `node-gyp` maintainers want it (or someone pops up in the comments here to say they want it).

### DIFF
--- a/components/git/land.js
+++ b/components/git/land.js
@@ -71,6 +71,13 @@ const landOptions = {
     describe: 'Query Jenkins CI results when checking the PR',
     default: true,
     type: 'boolean'
+  },
+  lite: {
+    type: 'boolean',
+    default: false,
+    describe: 'Eliminates some requirements, such as configuring a Jenkins ' +
+    'API token, and parsing TSC members and Collaborators from README.md. ' +
+    'Suitable for using "git-node land" in repositories other than nodejs/node.'
   }
 };
 
--- a/components/metadata.js
+++ b/components/metadata.js
@@ -12,7 +12,7 @@ const fs = require('fs');
 async function getMetadata(argv, skipRefs, cli) {
   const credentials = await auth({
     github: true,
-    jenkins: true
+    jenkins: !argv.lite
   });
   const request = new Request(credentials);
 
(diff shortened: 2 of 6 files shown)


rvagg:
yeah, could be a reasonable way to go, I still think moving away from this is probably the future for node-gyp, I tend to avoid pushing commits to nodejs/node these days because this process is getting more complex and onerous and the tooling probably only makes it more pleasant for frequent committers who have a good workflow and muscle memory for it.

I'll try and find some time to try this on node-gyp and report back.

DeeDeeG:
Yeah, the more I use this, the more of a barrier I see it would be to contribution. I have shifted my attention mostly to understanding `release-please`, at least for `node-gyp`.

For the record:I tried this on my machine, and it worked for me. I didn't end up pushing the "landed" commit anywhere, but here is a screen recording of it:

https://user-images.githubusercontent.com/20157115/117824482-868b7380-b23c-11eb-9262-dc833b9ad70f.mov

Note to `node-core-utils` maintainers: I still think this PR might be of interest if anyone wants to use `git-node land` elsewhere besides `nodejs/node`. There are a few other repos using the Node-style commit message prefixes and footers. But until someone actually wants to use `git-node land` with those repos, it seems reasonable not to merge this. If interest in this is renewed some time in the future, feel free to ping me about it. I will probably be able to respond to reviews/get this ready to merge if/when that happens.

PR #549, opened 2021-06-30
fix(update-v8): force-add all files after cloning V8

targos:
Add all V8 files with --force before updating DEPS. We have to do this
because some files are checked in by V8 despite .gitignore rules.


targos:
See https://github.com/nodejs/node/pull/38990#issuecomment-863110122 and discussion in https://chromium-review.googlesource.com/c/v8/v8/+/2972915

richardlau (approved):

PR #565, opened 2021-09-24
update-v8: add abseil-cpp as a V8 dependency

richardlau:
Add new V8 dependency introduced in V8 9.6.

Refs: https://github.com/nodejs/node-v8/issues/211

cc @targos @nodejs/v8-update

### DIFF
--- a/lib/update-v8/constants.js
+++ b/lib/update-v8/constants.js
@@ -42,6 +42,12 @@ exports.v8Deps = [
     since: 55
   },
   {
+    name: 'abseil-cpp',
+    repo: 'third_party/abseil-cpp',
+    gitignore: '!third_party/abseil-cpp',
+    since: 96
+  },
+  {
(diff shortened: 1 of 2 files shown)


targos (approved):

targos:
https://github.com/nodejs/node-v8/runs/3740771276?check_suite_focus=true

```
[06:04:01] → Cannot read property 'split' of undefined
[06:04:01] Major V8 update [failed]
✖ Cannot read property 'split' of undefined
[06:04:01] → Cannot read property 'split' of undefined
```

Looks like they're having issues with this dependency and removed it for now: https://github.com/v8/v8/commit/67601ba9c2e3564338363ee3cba24ff97bc11020

richardlau:
@targos Should we revert this then to fix the canary workflow?

targos:
Let's wait a couple of days to see if they reland it

PR #569, opened 2021-10-10
feat: spawn the user's editor to edit commit messages

targos:


targos:
Ping. I'd be really happy to have this feature :)

Trott (approved):
Code looks fine to me, but I didn't test it. Speaking of which, would it make sense to add tests for these options?

joyeecheung:
> Speaking of which, would it make sense to add tests for these options?

I think we'd have to create some kind of infra for testing git node in general first - testing silent runs with --yes would be easier but testing interactive runs like this one requires would be difficult. 

Speaking of which..have you tried running this with --yes and --autorebase? @targos

targos:
> Speaking of which..have you tried running this with --yes and --autorebase?

Yes, no problem with those options.

aduh95 (approved):

PR #573, opened 2021-10-20
fix: update detection of changelog links

richardlau:
The Node.js `CHANGELOG.md` file was changed so that `_` characters in
links are now escaped.

Refs: https://github.com/nodejs/node/pull/40322

---

Note that we'll need to update again after https://github.com/nodejs/node/pull/40475 but for now this needs to work against the current `v16.x-staging` branch.

### DIFF
--- a/lib/prepare_release.js
+++ b/lib/prepare_release.js
@@ -331,7 +331,8 @@ class ReleasePreparation {
     const data = await fs.readFile(mainChangelogPath, 'utf8');
     const arr = data.split('\n');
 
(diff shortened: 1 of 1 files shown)


richardlau:
> I put HTML comments in the README to remind people (including myself) of tooling that needs to be updated if the format of certain parts change. Maybe we can/should do that for CHANGELOG too?
> 
> https://github.com/nodejs/node/blame/master/README.md#L159-L160 https://github.com/nodejs/node/blame/master/README.md#L262-L264

Sounds reasonable.

PR #583, opened 2021-10-31
feat: suggest `gh pr` commands to finish landing

targos:
Also make the output easier to copy and paste.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -11,7 +11,7 @@ const {
   runAsync, runSync, forceRunAsync
 } = require('./run');
 const Session = require('./session');
-const { shortSha } = require('./utils');
+const { shortSha, isGhAvailable } = require('./utils');
 
 const isWindows = process.platform === 'win32';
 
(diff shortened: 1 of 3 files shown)


targos:
Example output:

```
The following commits are ready to be pushed to upstream/master
- 61a73f80c7 typings: add more bindings typings
Temporary files removed.
To finish landing:
1. Run:
   git push upstream master
2. Post "Landed in 61a73f80c7f9" in https://github.com/nodejs/node/pull/40415
   gh pr comment 40415 --body "Landed in 61a73f80c7f9"
   gh pr close 40415
```

Trott (approved):
Works for me if we don't think it will confuse folks.

aduh95:
Maybe we can test if `gh` is on the $PATH before printing its commands? It might be confusing for someone who doesn’t know what `gh` is.

targos:
> Maybe we can test if `gh` is on the $PATH before printing its commands? It might be confusing for someone who doesn’t know what `gh` is.

Ok, I'm thinking about using [`which`](https://github.com/npm/node-which#usage) to check whether a `gh` command exists. Do you have a better idea?

Trott:
> > Maybe we can test if `gh` is on the $PATH before printing its commands? It might be confusing for someone who doesn’t know what `gh` is.
> 
> Ok, I'm thinking about using [`which`](https://github.com/npm/node-which#usage) to check whether a `gh` command exists. Do you have a better idea?

For UNIX-like operating systems, maybe [use `command -v` over `which`](https://stackoverflow.com/a/677212/436641), although `which` will certainly work.

For Windows, I guess we'd want to use [`where`](https://stackoverflow.com/a/304447/436641).

targos:
I linked to the `which` npm package, not the Unix command

targos:
Done.

aduh95 (approved):

PR #586, opened 2021-10-31
feat(land): avoid landing on the wrong default branch

targos:
Fail if the user is trying to land a commit on "master" or "main", and
the default branch name doesn't match.


targos:
Example:

```
$ /Users/targos/git/nodejs/node-core-utils/bin/git-node land 40415
   ⚠  You are running git-node-land on `main`, but the default branch is master.
```

codebytere (approved):

PR #588, opened 2021-11-02
feat: check fast-track approvals

Mesteery:


lib/pr_checker.js
```diff
@@ -168,6 +168,25 @@ class PRChecker {
...
+        return false;
+      }
+      const [, requestor] = comment.bodyText.match(/^Fast-track has been requested by @(.+?)\. Please 👍 to approve\.$/);
+      const collaborators = [...this.data.collaborators.values()]
+        .map((c) => c.login);
+      if (!comment.reactions.nodes.some((r) =>
+        r.user.login !== requestor &&
+        collaborators.includes(r.user.login))) {
+        cli.error('The fast-track request requires' +
+                  " at least one collaborator's 👍.");
```
Trott, line 185: Fast tracking requires two collaborator approvals.

> The pull request can be fast-tracked if two collaborators approve the fast-tracking request. To land, the pull request itself still needs two collaborator approvals and a passing CI.



Ref: https://github.com/nodejs/node/blob/86099a375af95f519ae699b68ce101aa20236c5f/doc/guides/collaborator-guide.md#waiting-for-approvals

Would it be possible to implement logic such that it checks to see if the person that requested fast-tracking is also the author of the PR and, if so, checks for two collaborator fast-track 👍 reactions, and otherwise accepts one collaborator fast-track 👍 reaction as long as it's not a 👍 from either the PR author or the collaborator proposing fast-tracking?

In other words:

```
If fast-track requester is PR author:
    Require 2 👍 reactions from collaborators who are not the PR author
else:
    Require 1 👍 reaction from a collaborator who is not the PR author or the collaborator who requested fast-tracking
```

Trott:
Thanks for tackling this problem. The absence of fast-track confirmation is definitely a good loophole to fix!

targos (approved):

lib/pr_checker.js
```diff
@@ -9,6 +9,8 @@ const WAIT_TIME_SINGLE_APPROVAL = 24 * 7;
 
 const GITHUB_SUCCESS_CONCLUSIONS = ['SUCCESS', 'NEUTRAL', 'SKIPPED'];
 
+const FAST_TRACK_RE = /^Fast-track has been requested by @(.+?)\. Please 👍 to approve\.$/;
```
aduh95, line 12: Would it be possible to loosen this regex so if the wording of the comment changes in the future, it doesn't require to update ncu?

lib/pr_checker.js
```diff
@@ -168,6 +170,32 @@ class PRChecker {
...
+    if (isFastTracked) {
+      const comment = this.comments.find((c) =>
+        FAST_TRACK_RE.test(c.bodyText));
+      if (!comment) {
+        cli.error('Unable to find the fast-track request comment.');
+        return false;
+      }
+      const [, requester] = comment.bodyText.match(FAST_TRACK_RE);
+      const collaborators = [...this.data.collaborators.values()]
+        .map((c) => c.login);
```
aduh95, line 182: nit: we can skip the intermediary array creation.
```suggestion
      const collaborators = Array.from(this.data.collaborators.values(),
                                       (c) => c.login);
```

lib/pr_checker.js
Mesteery: I thought about it, but I'm afraid it would match a comment that is not a fast-track request (eg. if someone quotes the fast-track request comment).

Mesteery:
ping @Trott

lib/pr_checker.js
```diff
@@ -168,6 +170,32 @@ class PRChecker {
       }
     }
 
+    if (isFastTracked) {
+      const comment = this.comments.find((c) =>
+        FAST_TRACK_RE.test(c.bodyText));
+      if (!comment) {
+        cli.error('Unable to find the fast-track request comment.');
+        return false;
+      }
```
Trott, line 179: This is probably OK but just in case: There is a window, often of several minutes, where an issue is labeled with `fast-track` but the comment has not yet appeared. So this error may happen legitimately. Maybe that's fine? Or maybe we just need to mention it in the error message (perhaps by adding "Please try again in a few minutes." or something like that?).

Trott (approved):

lib/pr_checker.js
Mesteery: If there is no comment it means that there is no approval and therefore the PR can not be landed. So I think it's useless.

PR #590, opened 2021-11-06
chore: update many dependencies

targos:
- do major updates that do not affect us
- replace fs-extra with fs.promises
- replace node-fetch with undici


targos:
Ping :)

tniessen (approved):

Issue #594, opened 2021-11-17
TypeError: fetch is not a function


codebytere:
Seen after pulling latest `master` - I suspect https://github.com/nodejs/node-core-utils/commit/a571938d6210e1b3138a89037e3791f518bfc7d5.

<details>
<summary>Stacktrace</summary>

```console
node on git:master ❯ g node land 40757                                12:11PM
TypeError: fetch is not a function
    at wrappedFetch (/Users/codebytere/Developer/node-core-utils/lib/request.js:17:10)
    at Request.fetch (/Users/codebytere/Developer/node-core-utils/lib/request.js:37:12)
    at Request.text (/Users/codebytere/Developer/node-core-utils/lib/request.js:45:17)
    at Request.json (/Users/codebytere/Developer/node-core-utils/lib/request.js:50:29)
    at Request.query (/Users/codebytere/Developer/node-core-utils/lib/request.js:108:31)
    at Request.gql (/Users/codebytere/Developer/node-core-utils/lib/request.js:68:33)
    at LandingSession.warnForWrongBranch (/Users/codebytere/Developer/node-core-utils/lib/landing_session.js:486:65)
    at main (/Users/codebytere/Developer/node-core-utils/components/git/land.js:175:21)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

</details>

cc @targos

targos:
Did you run `npm install` ?

codebytere:
@targos yes, and removed `node_modules` to be doubly sure!

targos:
Oh wow, it looks like `undici` doesn't always export a `fetch` function:

```
PS C:\git\TEST> volta run --node 16.2.0 node -p "require('undici').fetch"
undefined
PS C:\git\TEST> volta run --node 16.13.0 node -p "require('undici').fetch"
[AsyncFunction: fetch]
```

I thought it would be a good idea to use the module to give it some real world test coverage...

@nodejs/undici

targos:
We should probably go back to `node-fetch`. `undici.fetch` doesn't work on the latest version of Node.js 14

ronag:
We only export fetch for node >= 16.5 since we depend on web streams

Ethan-Arrowood:
That is strange, but I believe Undici fetch is still considered experimental even on node 16. Best to stick to node-fetch for now. I can take a closer look later on as well

codebytere:
[...]

PR #595, opened 2021-11-17
fix(deps): revert to node-fetch

codebytere:
Closes https://github.com/nodejs/node-core-utils/issues/594.


targos (approved):
Use "fix" in the commit message otherwise it won't generate a release PR

PR #597, opened 2021-12-03
lib, test: prepare for npm audit fix

joyeecheung:
I tried to run `npm audit fix` today and here are the changes that are necessary for the tests to pass with the upgrade, so opening a PR first to fix them. The first commit makes the test data more deterministic (probably an issue that's exposed by my particular setup as this fails on the main branch too)  by making the dates different so that the items can be sorted properly by date. The second fixes a breakage caused by a breaking change in yargs by explicitly making the checkComments default to false (instead of counting on it being coerced from undefined to that). The last one are just format changes required by eslint upgrades.

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -40,7 +40,8 @@ function buildYargs(args = null) {
     .option('check-comments', {
       demandOption: false,
       describe: 'Check for \'LGTM\' in comments',
-      type: 'boolean'
+      type: 'boolean',
+      default: false
     })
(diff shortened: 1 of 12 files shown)


targos:
It looks like you started to work from a very old version of the code.

joyeecheung:
Oops, just noticed that my branch is out of date, I was running on the master branch. They should be fixed on the main branch now. Closing.

Issue #117, opened 2017-11-18
get-metadata: add "Refs:" in metadata


vsemozhetbyt:
Example: https://github.com/nodejs/node/pull/17107

IIRC, nodejs/node-review [adds this data](https://github.com/nodejs/node-review/search?utf8=%E2%9C%93&q=Refs&type=).

Tiriel:
Strange, this feature is already implemented and is supposed to work, but I can't get Refs in metadata either. 
Unit tests are even working and properly returning Refs. I'll investigate, if someone doesn't beat me to it.

joyeecheung:
node-review doesn't add that refs, actually.

<img width="405" alt="screen shot 2017-11-19 at 12 36 44 am" src="https://user-images.githubusercontent.com/4299420/32982590-c0de4904-ccc1-11e7-9e06-511b1cc77e2b.png">

The link parser code is largely borrowed from node-review. It currently only take urls that are `a.issue-link` i.e. github issues/PRs

joyeecheung:
@Tiriel Did you have a chance to look into this?

Tiriel:
@joyeecheung Not yet, had a pretty busy week. Putting it on my schedule for tonight!

cPhost:
The reason `Refs` is not being inserted to metadata because of
https://github.com/nodejs/node-core-utils/blob/edd107a706571bb69937f6ee02b74ff6b12c13f1/lib/links.js#L48

The query selector at line 48 `a.issue-link` returns undefined because github no longer uses `.issue-link` class and thats the reason the tests are passing but the feature is not working.

Way to fix this issue is just to remove `.issue-link` . and remove the unnecessary attributes form test except `href` just to make the test more readable.

\cc @joyeecheung @Tiriel

Tiriel:
Still looking for a propoer way to parse the links recursively, I'm on it ;)

joyeecheung:
@cPhost SGTM

Tiriel:
Don't know what I did last night but I was way too far... Opening a PR. Removing the class was all that was needed it seems. Tests OK and example given OK too.

PR #612, opened 2022-03-19
fix: make `fixes` and `refs` detection stricter

richardlau:
Only parse `fixes` and `refs` links at the beginning of lines.

Previously the parser was picking up the "fix" type Conventional
Commits titles from the changelog in body of the npm update pull
requests.

Refs: https://github.com/nodejs/node/pull/42382#issuecomment-1072920361

cc @Trott 

---

I have updated one of the existing test cases as otherwise the test was failing with this change. This test was from https://github.com/nodejs/node-core-utils/issues/117, for https://github.com/nodejs/node/pull/17107. I ran `node ../node-core-utils/bin/get-metadata.js 17107` with this diff:
```diff
diff --git a/lib/links.js b/lib/links.js
index 4d313a1..639ced5 100644
--- a/lib/links.js
+++ b/lib/links.js
@@ -14,6 +14,7 @@ export class LinkParser {
     this.owner = owner;
     this.repo = repo;
     this.$ = cheerio.load(html);
+    console.log(JSON.stringify(html));
   }
 
   getFixesUrlsFromArray(arr) {
```
and found the HTML differs from what is in the test fixture, so I updated it. The test passes was the updated HTML. I used the same diff to capture the test data for https://github.com/nodejs/node/pull/42382.


Trott (approved):

lib/links.js
```diff
@@ -1,8 +1,8 @@
 import cheerio from 'cheerio';
 
-const FIXES_RE = /(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/mgi;
+const FIXES_RE = /^(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/mgi;
```
tniessen, line 3:
```suggestion
const FIXES_RE = /^\s*(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/mgi;
```

lib/links.js
```diff
@@ -1,8 +1,8 @@
...
 const FIX_RE = /(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/i;
-const REFS_RE = /Refs?\s*:\s*(\S+)/mgi;
+const REFS_RE = /^Refs?\s*:\s*(\S+)/mgi;
```
tniessen, line 5:
```suggestion
const REFS_RE = /^\s*Refs?\s*:\s*(\S+)/mgi;
```

tniessen (approved):

lib/links.js
```diff
@@ -1,8 +1,8 @@
 import cheerio from 'cheerio';
 
-const FIXES_RE = /(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/mgi;
+const FIXES_RE = /^\s*(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/mgi;
```
Trott, line 3: By tolerating spaces at the start of the line, might this then match the "Fixes:" etc. in patches we float from V8 etc.? (Maybe that's not a problem?)

Trott: Maybe for a separate PR, but I think we probably don't need to tolerate spaces before the colon:

```suggestion
const FIXES_RE = /^\s*(Close[ds]?|Fix(e[ds])?|Resolve[sd]?):\s*(\S+)/mgi;
```

lib/links.js
```diff
@@ -1,8 +1,8 @@
...
 const FIX_RE = /(Close[ds]?|Fix(e[ds])?|Resolve[sd]?)\s*:\s*(\S+)/i;
-const REFS_RE = /Refs?\s*:\s*(\S+)/mgi;
+const REFS_RE = /^\s*Refs?\s*:\s*(\S+)/mgi;
```
Trott, line 5: Maybe for a separate PR, but I think we probably don't need to tolerate spaces before the colon:

```suggestion
const REFS_RE = /^\s*Refs?:\s*(\S+)/mgi;
```

lib/links.js
richardlau: > By tolerating spaces at the start of the line, might this then match the "Fixes:" etc. in patches we float from V8 etc.? (Maybe that's not a problem?)

Not sure. Maybe we need to find example PRs and run `git node metadata` on them to see what they generate? We can do that on both open and closed PRs -- there will be an error about the PR not being landable but the output will still include the generated metadata.

It's entirely possible we may need stricter checking of the right hand side of the colon, but it's hard to speculate on that without more concrete examples.

richardlau: Found one. https://github.com/nodejs/node/pull/37276 results in `Fixes: https://github.com/v8:11389` (with and without this PR):
```
$ git node metadata 37276
✔  Done loading data for nodejs/node/pull/37276
----------------------------------- PR info ------------------------------------
Title      deps: V8: workaround for darwin arm64 code page decommit failures (#37276)
   ⚠  Could not retrieve the email or name of the PR author's from user's GitHub profile!
Branch     matinzd:patch-1 -> nodejs:master
Labels     v8 engine, backported-to-v14.x
Commits    2
 - deps: V8: Workaround MacOS 11.2 code page decommit failures
 - bump: v8: v8_embedder_string to 25
Committers 2
 - GitHub <noreply@github.com>
 - matinzd <matin.zadehdolatabad@rambody.me>
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/37276
Fixes: https://github.com/v8:11389
Refs: https://bugs.chromium.org/p/v8/issues/detail?id=11389#c18
Reviewed-By: Michaël Zasso <targos@protonmail.com>
Reviewed-By: Ash Cripps <acripps@redhat.com>
```

richardlau: I've just noticed that for PR-URL we have a different parsing scheme that pulls the links out of the corresponding anchor tag (the parser is parsing HTML). Maybe we should be reusing that logic?

richardlau: Reusing the same logic that is used for `refs` and `PR-URL` seems a better solution. Or at least a more consistent one. PR: https://github.com/nodejs/node-core-utils/pull/614

richardlau:
Closing in favour of https://github.com/nodejs/node-core-utils/pull/614

PR #613, opened 2022-03-19
fix: correct username and token validation

Trott:
The current username and token validation regular expressions will match
any string. This adds tests and fixes the regular expressions.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -1,4 +1,3 @@
-import assert from 'node:assert';
 import fs from 'node:fs';
 import { ClientRequest } from 'node:http';
 import util from 'node:util';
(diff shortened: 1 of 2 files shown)


lib/auth.js
```diff
@@ -11,9 +10,24 @@ const ghauth = util.promisify(ghauthBase);
...
+  process.exit(1);
+}
+
 function check(username, token) {
-  assert(typeof username === 'string' && /^[a-zA-Z0-9]*/.test(username));
-  assert(typeof token === 'string' && /^[0-9a-f]*/.test(token));
+  if (typeof username !== 'string') {
+    errorExit(`username must be a string, received ${typeof username}`);
+  }
+  if (!/^[a-zA-Z0-9]+$/.test(username)) {
```
richardlau, line 22: These are GitHub usernames, aren't they? They're allowed to contain hyphens, at least according to https://github.com/join:
image

Trott, line 22:
```suggestion
  if (!/^[a-zA-Z0-9\-]+$/.test(username)) {
```

Trott: I added hyphens as a permissible character.

lib/auth.js
```diff
@@ -11,9 +10,24 @@ const ghauth = util.promisify(ghauthBase);
...
+  process.exit(1);
+}
+
 function check(username, token) {
-  assert(typeof username === 'string' && /^[a-zA-Z0-9]*/.test(username));
-  assert(typeof token === 'string' && /^[0-9a-f]*/.test(token));
+  if (typeof username !== 'string') {
+    errorExit(`username must be a string, received ${typeof username}`);
+  }
+  if (!/^[a-zA-Z0-9\-]+$/.test(username)) {
```
Trott, line 22:
```suggestion
  if (!/^[a-zA-Z0-9-]+$/.test(username)) {
```

lib/auth.js
```diff
@@ -11,9 +10,24 @@ const ghauth = util.promisify(ghauthBase);
...
+}
+
 function check(username, token) {
-  assert(typeof username === 'string' && /^[a-zA-Z0-9]*/.test(username));
-  assert(typeof token === 'string' && /^[0-9a-f]*/.test(token));
+  if (typeof username !== 'string') {
+    errorExit(`username must be a string, received ${typeof username}`);
+  }
+  if (!/^[a-zA-Z0-9-]+$/.test(username)) {
+    errorExit(`username must be alphanumeric, received ${username}`);
```
richardlau, line 23: Something more in line with GitHub error text (without the bits we're not actually checking for)?
```suggestion
    errorExit(`username may only contain alphanumeric characters or hyphens, received ${username}`);
```

richardlau (approved):

PR #627, opened 2022-04-26
docs: clarify global gitignore and fix typos

LiviaMedeiros:
The [Make sure your credentials won't be committed](https://github.com/nodejs/node-core-utils#make-sure-your-credentials-wont-be-committed) part of guide implies that `~/.gitignore_global` is a default path for global gitignore.

If anyone followed this part too literally, I'd recommend to doublecheck their setup with something like:
```bash
$ cat "$(git config --global core.excludesFile || echo ${XDG_CONFIG_HOME:-${HOME}/.config}/git/ignore)"
# should show the contents of global gitignore
```

This is applicable to Linux; I'm unsure about other platforms.
`git config --global core.excludesfile` should work everywhere, but note that setting it overrides file in default location (if any).

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ To obtain the Jenkins API token
   token in the inbox that appears, and click `GENERATE`.
 4. Copy the generated token.
 5. Add it into your `ncurc` file (`~/.ncurc` or `$XDG_CONFIG_HOME/ncurc`)
-  with `jenkins_token` as key, like this
+  with `jenkins_token` as key, like this:
 
   ```json
   {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -113,7 +113,8 @@ To obtain the Jenkins API token
 
 ### Make sure your credentials won't be committed
 
-Put the following entries into `~/.gitignore_global`
+Put the following entries into your global gitignore file
```
aduh95, line 116:
```suggestion
Put the following entries into your
[global gitignore file](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreexcludesFile)
```

Issue #602, opened 2022-01-14
"git node land" hides co-authored-by trailers


tniessen:
The original commit message in https://github.com/nodejs/node/pull/41490 was:

```
commit f042c61749a34565ab57b5ed5f7c89ab9e894494
Author: Tobias Nießen <tniessen@tnie.de>
Date:   Tue Jan 11 20:38:00 2022 +0000

    src: gracefully handle errors in GetX509NameObject

    Co-authored-by: Anna Henningsen <anna@addaleax.net>
```

However, `git node land` added metadata at the end without incorporating the `Co-authored-by` line:

```
commit 721cc6b5e2d6f4112b56ee5ead19ddaeb99ac7a6
Author: Tobias Nießen <tniessen@tnie.de>
Date:   Tue Jan 11 20:38:00 2022 +0000

    src: gracefully handle errors in GetX509NameObject

    Co-authored-by: Anna Henningsen <anna@addaleax.net>

    PR-URL: https://github.com/nodejs/node/pull/41490
    Reviewed-By: Anna Henningsen <anna@addaleax.net>
    Reviewed-By: Colin Ihrig <cjihrig@gmail.com>
```

GitHub does not recognize the `Co-authored-by` line since there is an empty line between that and the metadata lines at the end. I didn't know this and I only coincidentally noticed that GitHub did not display the correct commit authorship after pushed.

I have force-pushed the correct the commit message:

```
commit 4dd1f42df6ee34c1a5456862647092cb9686ec9a (HEAD -> master, upstream/master)
Author: Tobias Nießen <tniessen@tnie.de>
Date:   Tue Jan 11 20:38:00 2022 +0000

    src: gracefully handle errors in GetX509NameObject

    PR-URL: https://github.com/nodejs/node/pull/41490
    Co-authored-by: Anna Henningsen <anna@addaleax.net>
    Reviewed-By: Anna Henningsen <anna@addaleax.net>
    Reviewed-By: Colin Ihrig <cjihrig@gmail.com>
```

Both messages pass `core-validate-commit`.

---

Other examples:
- https://github.com/nodejs/node/commit/500241d50df3049603b342afdc0270ecb1faccff
- https://github.com/nodejs/node/commit/f6e5050579d8d0ec7c68ed499dcc8701bd49c18a
- https://github.com/nodejs/node/commit/09c25bb224095b688b395acd2d7d58e4ed847e31
[...]

PR #632, opened 2022-05-14
fix: respect existing trailers in commit messages

tniessen:
Use `git interpret-trailers` (probably requires a fairly recent git version) to see if the original commit message already contains trailers and do not add an empty line before adding metadata if it does (as suggested in https://github.com/nodejs/node-core-utils/issues/602#issuecomment-1015172790). This should fix how `git node land` treats `Co-authored-by` lines.

No reordering happens. The generated metadata is appended to the existing trailer. My understanding is that the order of trailers is irrelevant, but if that's not true, we might want to revisit that part.

This particular part of the code base does not seem to be covered by any tests, so please review carefully.

Fixes: https://github.com/nodejs/node-core-utils/issues/602

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -306,9 +306,24 @@ export default class LandingSession extends Session {
     const original = runSync('git', [
       'show', 'HEAD', '-s', '--format=%B'
     ]).trim();
+
+    // git has very specific rules about what is a trailer and what is not.
+    // Instead of trying to implement those ourselves, let git parse the
+    // original commit message and see if it outputs any trailers.
+    const originalHasTrailers = runSync('git', [
+      'interpret-trailers', '--parse', '--no-divider'
+    ], {
+      input: `${original}\n`
+    }).trim().length !== 0;
+
     const metadata = this.metadata.trim().split('\n');
     const amended = original.split('\n');
-    if (amended[amended.length - 1] !== '') {
+
+    // If the original commit message already contains trailers (such as
+    // "Co-authored-by"), we simply add our own metadata after those. Otherwise,
+    // we have to add an empty line so that git recognizes our own metadata as
+    // trailers in the amended commit message.
+    if (!originalHasTrailers) {
       amended.push('');
     }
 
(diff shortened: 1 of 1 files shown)


lib/landing_session.js
```diff
@@ -306,9 +306,24 @@ export default class LandingSession extends Session {
...
+    // original commit message and see if it outputs any trailers.
+    const originalHasTrailers = runSync('git', [
+      'interpret-trailers', '--parse', '--no-divider'
+    ], {
+      input: `${original}\n`
+    }).trim().length !== 0;
+
     const metadata = this.metadata.trim().split('\n');
     const amended = original.split('\n');
-    if (amended[amended.length - 1] !== '') {
```
tniessen, line 311: This condition appears to have always been `false` because the code above calls `trim()` on the commit message before splitting it at `\n`.

lib/landing_session.js
```diff
@@ -317,9 +332,13 @@ export default class LandingSession extends Session {
...
     for (const line of metadata) {
-      if (original.includes(line)) {
-        if (line) {
+      if (line.length !== 0 && original.includes(line)) {
+        if (originalHasTrailers) {
           cli.warn(`Found ${line}, skipping..`);
+        } else {
+          cli.error('Git found no trailers in the original commit message, ' +
+                    `but '${line}' is present and should be a trailer.`);
+          process.exit(1);  // make it work with git rebase -x
```
tniessen, line 341: It seems reasonable to abort in this case, but if anyone thinks it could be problematic, I'd be happy to revert this particular check.

targos:
I'm going to try it with some backports and V8 update

tniessen:
@targos Have you had a chance to test this yet? Looks like it'll be difficult to write tests for this part.

targos:
> Have you had a chance to test this yet?

It's still on my todo list.

aduh95 (approved):
Just tried that with https://github.com/nodejs/node/pull/44048 and I confirm this PR fixes the issue.

targos (approved):

PR #634, opened 2022-06-02
chore: update GitHub actions

aduh95:
And don't share `GITHUB_TOKEN` when not necessary.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -14,7 +14,6 @@ jobs:
       - uses: google-github-actions/release-please-action@v3
         id: release
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
           release-type: node
           package-name: node-core-utils
   npm-publish:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-please.yml
```diff
@@ -14,7 +14,6 @@ jobs:
       - uses: google-github-actions/release-please-action@v3
         id: release
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
```
targos, line 17: Without the token, CI is not run on the release pull requests.

aduh95: Not according to their docs: https://github.com/google-github-actions/release-please-action#setting-up-this-action

targos: It's not specific to this action, see https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow.
But you can disregard my comment because there's no point passing `secrets.GITHUB_TOKEN` (we need a PAT to fix the issue). We can see in https://github.com/nodejs/node-core-utils/pull/630 that no CI was run.

aduh95: Ah gotcha, yes passing the token won't help with that.

Issue #639, opened 2022-07-04
ncu-ci commit TypeError: Cannot read properties of null (reading '1')


F3n67u:
## Reproduce

```
$ ncu-ci commit 54550                      
--------------------------------------------------------------------------------
[1/1] Running COMMIT: 54550
--------------------------------------------------------------------------------
✔  Build data downloaded
⠏ Querying API for job/node-test-binary-windows-js-suites/RUN_SUBSET=2,nodes=win10-arm64-COMPILED_BY-vs2019-arm64/13778/TypeError: Cannot read properties of null (reading '1')
    at new JSTestFailure (file:///usr/local/lib/node_modules/node-core-utils/lib/ci/ci_failure_parser.js:78:62)
    at file:///usr/local/lib/node_modules/node-core-utils/lib/ci/ci_failure_parser.js:168:16
    at Array.map (<anonymous>)
    at filter (file:///usr/local/lib/node_modules/node-core-utils/lib/ci/ci_failure_parser.js:167:29)
    at CIFailureParser.parse (file:///usr/local/lib/node_modules/node-core-utils/lib/ci/ci_failure_parser.js:296:22)
    at TestRun.parseConsoleText (file:///usr/local/lib/node_modules/node-core-utils/lib/ci/build-types/job.js:93:26)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Promise.all (index 13)
    at async CommitBuild.getResults (file:///usr/local/lib/node_modules/node-core-utils/lib/ci/build-types/commit_build.js:106:25)
    at async JobCommand.drain (file:///usr/local/lib/node_modules/node-core-utils/bin/ncu-ci.js:352:7)
```

## Expected behavior
No Error happened, could process `job/node-test-commit/54550/` normally

## Actual behavior 
TypeError thrown

## Additional Information

Refs: https://github.com/nodejs/reliability/issues/316

Jenkins job URL: 
https://ci.nodejs.org/job/node-test-commit-osx-arm/6298/
https://ci.nodejs.org/job/node-test-commit/54550/
https://ci.nodejs.org/job/node-test-pull-request/44957/

aduh95:
The CI has been locked in preparation for the security release, I think the error is the expected behavior.

richardlau:
This was occuring at the end of last week before the CI was locked down, see https://github.com/nodejs/reliability/issues/316.

F3n67u:
[...]

PR #640, opened 2022-07-11
fix: parse ci failure error

F3n67u:
fix https://github.com/nodejs/node-core-utils/issues/639

### DIFF
--- a/lib/ci/ci_failure_parser.js
+++ b/lib/ci/ci_failure_parser.js
@@ -155,7 +155,7 @@ const FAILURE_FILTERS = [{
 }, {
   // TODO: match indentation to avoid skipping context with '...'
   filter(ctx, text) {
-    const pattern = /not ok \d+[\s\S]+? {2}\.\.\.\r?\n/mg;
+    const pattern = /^not ok \d+[\s\S]+? {2}\.\.\.\r?\n/mg;
     const matches = text.match(pattern);
     if (!matches) {
       return null;
--- a/test/fixtures/jenkins/test-runner-test-failure/expected.json
+++ b/test/fixtures/jenkins/test-runner-test-failure/expected.json
@@ -1,0 +1,9 @@
+[{
+  "builtOn": "test-nearform-macos11.0-arm64-1",
+  "file": "message/test_runner_desctibe_it",
+  "highlight": 0,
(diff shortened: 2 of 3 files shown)


MoLow (approved):
LGTM,
probably best to add a test here? https://github.com/nodejs/node-core-utils/blob/main/test/unit/ci_failure_parser.test.js

F3n67u:
I will add a test for this fix.

F3n67u:
> LGTM, probably best to add a test here? https://github.com/nodejs/node-core-utils/blob/main/test/unit/ci_failure_parser.test.js

of course. I am working on the test. this is why this pr is a draft.😀

codebytere (approved):

F3n67u:
I added a test for the fix, please take another look.

lib/ci/ci_failure_parser.js
```diff
@@ -155,7 +155,7 @@ const FAILURE_FILTERS = [{
 }, {
   // TODO: match indentation to avoid skipping context with '...'
   filter(ctx, text) {
-    const pattern = /not ok \d+[\s\S]+? {2}\.\.\.\r?\n/mg;
+    const pattern = /^not ok \d+[\s\S]+? {2}\.\.\.\r?\n/mg;
```
aduh95, line 158: Following the same logic
```suggestion
    const pattern = /^not ok \d+[\s\S]+? {2}\.\.\.$/mg;
```

F3n67u: thanks

F3n67u: Changing `\r?\n` to `$` will make this regex not match `\r?\n` after `...`, this is why the test fails. May I revert this change?

F3n67u: Reason to revert:

1. after the change, the `reason` of the parse result has `\n` at the end of each line except the last line `...`. I thinks we would better keep consistent: add `\n` to the last line also.
2. avoid potential breaking change, if anywhere consume the reason property.

aduh95 (approved):

F3n67u:
Update CI reliability job stop working again, please help to merge this pr, Thank you! https://github.com/nodejs/reliability/runs/7587812485?check_suite_focus=true

PR #641, opened 2022-07-12
feat: ignore .md files when do `requiresJenkinsRun` check

F3n67u:
This pr ignores .md files when do `requiresJenkinsRun` check

Refs: https://github.com/nodejs/node/pull/43483#issuecomment-1160210710

cc @aduh95

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -404,6 +404,13 @@ export default class PRChecker {
       return false;
     }
 
+    const files = pr.files.nodes;
+
+    // Don't require Jenkins run for doc-only change.
+    if (files.every(({ path }) => path.endsWith('.md'))) {
+      return false;
+    }
+
     const ciNeededFolderRx = /^(deps|lib|src|test)\//;
     const ciNeededToolFolderRx =
       /^tools\/(code_cache|gyp|icu|inspector|msvs|snapshot|v8_gypfiles)/;
@@ -416,7 +423,7 @@ export default class PRChecker {
     ];
     const ciNeededExtensionList = ['.gyp', '.gypi', '.bat'];
 
-    return pr.files.nodes.some(
+    return files.some(
       ({ path }) =>
         ciNeededFolderRx.test(path) ||
         ciNeededToolFolderRx.test(path) ||
--- a/test/fixtures/pull_requests/code-change.json
+++ b/test/fixtures/pull_requests/code-change.json
@@ -1,0 +1,23 @@
+{
+  "createdAt": "2022-07-12T11:55:42Z",
+  "authorAssociation": "MEMBER",
+  "author": {
+    "login": "F3n67u",
+    "email": "F3n67u@outlook.com",
+    "name": "Feng Yu"
+  },
+  "url": "https://github.com/nodejs/node/pull/43796",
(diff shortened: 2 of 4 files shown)


F3n67u:
covert to draft because I still add more tests for this feature to keep coverage increasing.

lib/pr_checker.js
```diff
@@ -404,6 +404,14 @@ export default class PRChecker {
       return false;
     }
 
+    const files = pr.files.nodes;
+
+    // Don't require Jenkins run if pr is doc-only changes
+    const isDocOnlyChange = files.every(({ path }) => path.endsWith('.md'));
+    if (isDocOnlyChange) {
```
aduh95, line 411: I don't think we need a variable that we don't reuse later, and the comment above already explains what it does.
```suggestion
    if (files.every(({ path }) => path.endsWith('.md'))) {
```

aduh95 (approved):

tniessen (commented):
How often does this help, i.e., how often do PRs edit `*.md` files outside of `doc`?

If it does not happen a lot, I am concerned about being too permissive.
1. There is no guarantee that dependencies will use `.md` for documentation only.
2. Even node itself does not use `.md` for documentation only. For example, `test/fixtures` contains `.md` files that affect tests (e.g., because they are linter test cases).

aduh95:
> 2\. Even node itself does not use `.md` for documentation only. For example, `test/fixtures` contains `.md` files that affect tests (e.g., because they are linter test cases).

@tniessen those tests are covered by the GHA CI, I think it's safe to skip Jenkins CI for those.

tniessen:
> > 2. Even node itself does not use `.md` for documentation only. For example, `test/fixtures` contains `.md` files that affect tests (e.g., because they are linter test cases).
> 
> @tniessen those tests are covered by the GHA CI, I think it's safe to skip Jenkins CI for those.

I don't think it's a problem with the current state of the repository, but there is no guarantee that this won't change in the future.

> > 1. There is no guarantee that dependencies will use `.md` for documentation only.
> 
> Sure, but that seems even more rare than what this PR is trying to fix. For those case, it's still possible to add the `needs-ci` label manually.

Agreed, both false positives without this PR and false negatives with this PR are unlikely.

---

Anyway, I am not explicitly blocking this PR; erroring on the side of caution is merely my personal preference.

F3n67u:
> How often does this help, i.e., how often do PRs edit `*.md` files outside of `doc`?
> 
> If it does not happen a lot, I am concerned about being too permissive.
> 
> 1. There is no guarantee that dependencies will use `.md` for documentation only.
> 2. Even node itself does not use `.md` for documentation only. For example, `test/fixtures` contains `.md` files that affect tests (e.g., because they are linter test cases).

@tniessen  Node core will add `needs-ci` label to all pr that modify files in `test` dir. If a pr modifies files in `test/fixtures` contains `.md` files that affect tests, the pr will have a `needs-ci` label and still require a Jenkins run.

This patch is useful in the below scenario:
1. A pr makes a `.md` only change and this change doesn't affect tests or other functional behavior. Even though, because the rule of pr label, this pr is added a `needs-ci` label and requires a Jenkins CI run to land in the commit queue.
2. A collaborator found this pr is just a doc change and remove the `needs-ci` label and then add the pr to the commit queue.
3.  As a surprise, commit queue fail because node-core-utils still think a Jenkins CI run is required. This is a surprise.

This patch's motivation is to remove this surprise: If you do think this `.md` change only pr is not needed ci run, then remove `needs-ci` will lead to a commit queue success but not commit queue to fail.

PR #645, opened 2022-07-30
fix: add a specific error message for the commit queue

aduh95:
We can detect when we're running from a GitHub Actions runner, and output an error message specific to the CQ (as `Use --fixupAll` is not a very actionable workaround in the context of the CQ).


targos:
Message sgtm. Can you fix the ESLint errors?

targos (approved):

PR #668, opened 2023-02-24
docs: update Jenkins profile access

VoltrexKeyva:
The Node.js Jenkins server now only allows you to access your Jenkins profile with your GitHub username rather than your GitHub ID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,8 @@ these commands.
 To obtain the Jenkins API token
 
 1. Open
-`https://ci.nodejs.org/user/<your-github-id>/configure` (replace
-\<your-github-id\> with your own GitHub ID)
+   `https://ci.nodejs.org/user/<your-github-username>/configure` (replace
+   \<your-github-username\> with your own GitHub username).
 2. Click on the `ADD NEW TOKEN` button in the `API Token` section.
 3. Enter an identifiable name (for example, `node-core-utils`) for this
   token in the inbox that appears, and click `GENERATE`.


README.md
```diff
@@ -93,8 +93,8 @@ these commands.
 To obtain the Jenkins API token
 
 1. Open
-`https://ci.nodejs.org/user/<your-github-id>/configure` (replace
-\<your-github-id\> with your own GitHub ID)
+`https://ci.nodejs.org/user/<your-github-username>/configure` (replace
+\<your-github-username\> with your own GitHub username)
```
aduh95, line 97: While we're at it, let's fix the indentation and the punctuation.
```suggestion
  `https://ci.nodejs.org/user/<your-github-username>/configure` (replace
  \<your-github-username\> with your own GitHub username).
```

VoltrexKeyva: Done, though indented with 2 spaces instead of 3 to stay consistent with the other lines.

aduh95: It should be 3 though, according to https://spec.commonmark.org/0.30/#list-items

> the continuation blocks must be indented at least to the column of the first character other than a space or tab after the list marker

(as the spec says, the above sentence is a simplification and is not always correct, but it is correct in this case)

I don't have a strong opinion wether we should fix the other occurrences before or after this PR, but I don't think it's correct to land this with only 2 spaces.

VoltrexKeyva: I'll indent it with 3 spaces and open a new PR to fix the other occurrences.

aduh95 (approved):

VoltrexKeyva:
@aduh95 can you merge this before I open the new PR?

PR #669, opened 2023-02-25
chore: update deps and dev deps

VoltrexKeyva:
- Update the dependencies and developer dependencies to their latest versions.
- Remove the `rimraf` dependency in favor of the built-in `fs.rmSync()` method.
- Remove deprecated `eslint-plugin-standard` developer dependency as it's no longer needed.


lib/session.js
```diff
@@ -1,7 +1,7 @@
 import path from 'node:path';
 import fs from 'node:fs';
 
-import rimraf from 'rimraf';
+import { rimrafSync } from 'rimraf';
```
aduh95, line 4: Can we get rid of it in favor of `fs.rmSync`?

VoltrexKeyva: Good idea, done.

package.json
```diff
@@ -34,38 +34,36 @@
...
-    "inquirer": "^8.2.4",
-    "listr2": "^4.0.5",
+    "ghauth": "^5.0.1",
+    "inquirer": "^9.1.4",
+    "listr2": "^5.0.7",
     "lodash": "^4.17.21",
     "log-symbols": "^5.1.0",
-    "ora": "^6.1.0",
+    "ora": "^6.1.2",
     "proxy-agent": "^5.0.0",
```
aduh95, line 53: note for a future PR: Undici has its own `ProxyAgent`, maybe we can migrate to that.

package.json
```diff
@@ -34,38 +34,36 @@
...
     "cheerio": "^1.0.0-rc.10",
     "clipboardy": "^3.0.0",
-    "core-validate-commit": "^3.16.0",
+    "core-validate-commit": "^3.18.0",
     "enquirer": "^2.3.6",
-    "execa": "^6.1.0",
-    "figures": "^4.0.1",
+    "execa": "^7.0.0",
+    "figures": "^5.0.0",
     "form-data": "^4.0.0",
```
aduh95, line 46: note for a future PR: Undici has its own `FormData` implementation, we should probably migrate to that.

package.json
```diff
@@ -34,38 +34,36 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "branch-diff": "^1.10.5",
-    "chalk": "^5.0.1",
-    "changelog-maker": "^2.8.0",
+    "branch-diff": "^2.1.0",
+    "chalk": "^5.2.0",
+    "changelog-maker": "^3.2.1",
     "cheerio": "^1.0.0-rc.10",
```
aduh95, line 40: note for a future PR: we should probably get rid of cheerio, currently we use it to scrap the links from the html page, but probably the GH API would do a better job at that.

package.json
```diff
@@ -34,38 +34,36 @@
...
     "enquirer": "^2.3.6",
-    "execa": "^6.1.0",
-    "figures": "^4.0.1",
+    "execa": "^7.0.0",
+    "figures": "^5.0.0",
     "form-data": "^4.0.0",
-    "ghauth": "^4.0.0",
-    "inquirer": "^8.2.4",
-    "listr2": "^4.0.5",
+    "ghauth": "^5.0.1",
```
aduh95, line 47: `ghauth` has switched to promises (with support for callbacks), we should remove the `promisify`. (We could also send upstream changes to switch from `node-fetch` to `undici`, but that's not something blocking this PR).

package.json
```diff
@@ -34,38 +34,36 @@
...
+    "inquirer": "^9.1.4",
+    "listr2": "^5.0.7",
     "lodash": "^4.17.21",
```
aduh95, line 50: note for a future PR: removing `lodash` would be nice

package.json
```diff
@@ -34,38 +34,36 @@
...
+    "ora": "^6.1.2",
     "proxy-agent": "^5.0.0",
-    "replace-in-file": "^6.3.2",
-    "rimraf": "^3.0.2",
+    "replace-in-file": "^6.3.5",
     "undici": "^5.20.0",
-    "which": "^2.0.2",
-    "yargs": "^17.5.0"
+    "which": "^3.0.0",
+    "yargs": "^17.7.1"
```
aduh95, line 57: note for a future PR: it'd be nice to switch to `util.parseArgs` (but that's likely not a lightway task, it's also fine to keep yargs)

aduh95 (commented):
Did a review of the deps, only `ghauth` needs to be addressed in this PR, rest of the comments are notes for later.

package.json
VoltrexKeyva: Done.

aduh95 (approved):

PR #670, opened 2023-02-26
chore(test): migrate to `node:test`

aduh95:


targos:
`Segmentation fault (core dumped)`

Interesting.

MoLow:
> Probably related to https://github.com/nodejs/node/pull/46674
I am pretty sure it is since node 16 is lacking many features of the test runner

MoLow:
I was able to run on all three node versions in my branch:
https://github.com/aduh95/node-core-utils/compare/node--test...MoLow:node-core-utils:node--test
the main 2 issues that are not supported on node 16 are 
https://github.com/nodejs/node/pull/45055
https://github.com/nodejs/node/pull/45161

MoLow:
@aduh95 the tests pass but coverage can only be collected on node >= 18

MoLow:
I think it might be a fair mitigation to run tests without coverage on node 16

aduh95:
It looks like there are a few test failures though

MoLow:
yeah since `node --test` now runs files inside `test/fixtures`, I think running `test-unit` would have the same outcome as running `test-all` had before, unless I am missing something

MoLow:
@aduh95 this should work: https://github.com/aduh95/node-core-utils/commit/d7ea450b4f5612d5508920aee953960602481977

aduh95:
For some reason, CI picks up v18.15.0 only on Windows 🤔

MoLow:
@aduh95 seems like `actions/setup-node@v3` should have `check-latest: true`

MoLow (approved):
🎉

MoLow:
@aduh95 this requires amending the commit message(s)

aduh95:
No it doesn't, the title of the PR (which is what GitHub uses as the default commit message) is complying with Conventional Commit spec.

MoLow:
so this check should be fixed to only look at the PR name? https://github.com/nodejs/node-core-utils/actions/runs/4430791079/jobs/7772963001?pr=670

aduh95:
The issue is that some PR are being "Rebase and merge"d, in which case it makes sense to check every commits.

targos (approved):

PR #671, opened 2023-02-26
fix: do not run `git cherry-pick --abort` on failure

aduh95:
I personally find it useful to keep the cherry-pick session as is to help with debugging what's going wrong.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -121,8 +121,6 @@ export default class LandingSession extends Session {
         ignoreFailure: false
       });
     } catch (ex) {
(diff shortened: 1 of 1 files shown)


aduh95:
> This is to help manual landing? 

Correct, e.g. when a PR has merge commits.

> Is it going to change something for CQ landing?

It should not, because the CQ runs `git node land --abort` after each failed attempt, and that would reset any git cherry-pick session: https://github.com/nodejs/node-core-utils/blob/6926407c8ba3e3e036d4328d8fff9a53c50aec65/lib/session.js#L278-L289

anonrig (approved):

targos (approved):

PR #675, opened 2023-02-27
fix: use correct V8 tag for major updates

targos:
Sometimes there might be additional tags with suffix such as `-pgo`


targos:
Fixes: https://github.com/nodejs/node-v8/actions/runs/4279399575/jobs/7450060217

aduh95 (approved):

codebytere (approved):

PR #676, opened 2023-03-01
fix: treat `fast-track` with not enough approvals as non-fatal

LiviaMedeiros:
If a PR has `fast-track` with not enough approvals, instead of exiting with error we can proceed to see if it meets regular time requirements.
This also makes error more informative, showing these requirements.

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -138,7 +138,7 @@ export default class PRChecker {
     const { requestedChanges, approved } = reviewers;
     const labels = pr.labels.nodes.map((l) => l.name);
 
-    const isFastTracked = labels.includes('fast-track');
+    let isFastTracked = labels.includes('fast-track');
     const isCodeAndLearn = labels.includes('code-and-learn');
     const isSemverMajor = labels.includes('semver-major');
 
@@ -184,13 +184,13 @@ export default class PRChecker {
         collaborators.includes(r.user.login.toLowerCase())).length;
 
       if (requester === pr.author.login && approvals < 2) {
-        cli.error('The fast-track request requires' +
+        cli.warn('The fast-track request requires' +
                   " at least two collaborators' approvals (👍).");
-        return false;
+        isFastTracked = false;
       } else if (approvals === 0) {
-        cli.error('The fast-track request requires' +
+        cli.warn('The fast-track request requires' +
                   " at least one collaborator's approval (👍).");
-        return false;
+        isFastTracked = false;
       }
     }
 
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -27,6 +27,7 @@ const WAIT_TIME_SINGLE_APPROVAL = 24 * 7;
 const GITHUB_SUCCESS_CONCLUSIONS = ['SUCCESS', 'NEUTRAL', 'SKIPPED'];
 
 const FAST_TRACK_RE = /^Fast-track has been requested by @(.+?)\. Please 👍 to approve\.$/;
+const FAST_TRACK_MIN_APPROVALS = 2;
 const GIT_CONFIG_GUIDE_URL = 'https://github.com/nodejs/node/blob/99b1ada/doc/guides/contributing/pull-requests.md#step-1-fork';
 
 export default class PRChecker {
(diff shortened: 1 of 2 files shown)


anonrig (approved):

anonrig:
Can you update the tests?

aduh95:
Maybe we could have the CQ remove the https://github.com/nodejs/node/labels/fast-track label on PRs that have been opened for more than 48h, so we can make statistics on how many PRs were actually fast-tracked (but AFAIK no one is doing any statistics, so it wouldn't be a big deal if we would simply ignore it).

targos:
> Maybe we could have the CQ remove the https://github.com/nodejs/node/labels/fast-track label on PRs that have been opened for more than 48h, so we can make statistics on how many PRs were actually fast-tracked (but AFAIK no one is doing any statistics, so it wouldn't be a big deal if we would simply ignore it).

Why not, if it's not too difficult. I'd go even further with this fix and not warn at all. After 48 hours, the fast-track request isn't relevant anymore.

lib/pr_checker.js
```diff
@@ -183,14 +185,13 @@ export default class PRChecker {
...
-        isFastTracked = false;
-      } else if (approvals === 0) {
-        cli.warn('The fast-track request requires' +
-                  " at least one collaborator's approval (👍).");
+      const missingFastTrackApprovals = FAST_TRACK_MIN_APPROVALS - approvals -
+        (requester === pr.author.login ? 0 : 1);
+      if (missingFastTrackApprovals > 0) {
         isFastTracked = false;
+        fastTrackAppendix = ' (or 0 minutes if there is ' +
+          `${missingFastTrackApprovals} more approval (👍) of ` +
```
aduh95, line 193:
```suggestion
          `${missingFastTrackApprovals} more approval(s) (👍) of ` +
```

lib/pr_checker.js
```diff
@@ -183,14 +185,13 @@ export default class PRChecker {
...
-        return false;
-      } else if (approvals === 0) {
-        cli.error('The fast-track request requires' +
-                  " at least one collaborator's approval (👍).");
-        return false;
+      const missingFastTrackApprovals = FAST_TRACK_MIN_APPROVALS - approvals -
+        (requester === pr.author.login ? 0 : 1);
+      if (missingFastTrackApprovals > 0) {
+        isFastTracked = false;
+        fastTrackAppendix = ' (or 0 minutes if there is ' +
```
aduh95, line 192: Sorry for my previous suggestion, it should probably be hours rather than minutes, to align with the other messages.

```suggestion
        fastTrackAppendix = ' (or 0 hours if there is ' +
```

LiviaMedeiros:
> Can we add a more precise error message when there's not enough approval, similar to what was done in

The combination of warning for fast-track and error for time was for the same thing. Does updated message look better?

> After 48 hours, the fast-track request isn't relevant anymore.

AFAICT it's not relevant for PRs with 2 or more approvals, but still might make sense for 7 days if there's only one?

I think, ideally CQ should "wait" for fast-tracked PRs to have enough approves rather than exiting with error.

aduh95:
> I think, ideally CQ should "wait" for fast-tracked PRs to have enough approves rather than exiting with error.

That's not really practical, as the CQ can't discriminate with a simple request a ready fast-tracked PR from one who lacks approvals. So if we don't remove the `commit-queue` label, it forces the CQ to go to the trouble of cloning the repo etc. every 5 minutes. I think the current way of doing things is a good compromise.

> > After 48 hours, the fast-track request isn't relevant anymore.
> 
> AFAICT it's not relevant for PRs with 2 or more approvals, but still might make sense for 7 days if there's only one?

No, as fast-tracked PR needs at least two approval to land, and if it has two approvals, it doesn't need to be fast-tracked.

aduh95 (commented):
I think we need to fix the grammar here, even if it makes the code uglier.

aduh95 (approved):
Could we add a test where there is only one approving review, and the fast-track request has only one or 0 :+1:? Non-blocking question, this PR can also land without it.

anonrig (approved):

Issue #677, opened 2023-03-02
Force push does not invalidate existing approved reviews


anonrig:
If the author force pushes to their branch after a reviewer approved it, I assumed that it would invalidate existing approved reviews, since the code that was reviewed is changed. But right now approved reviews count as valid ones, and might lead to unwanted code to be merged to main.

Example https://github.com/nodejs/node/pull/46904

```
➜  node git:(main) ✗ git node metadata 46904
✔  Done loading data for nodejs/node/pull/46904
----------------------------------- PR info ------------------------------------
Title      url: use private properties for brand check (#46904)
Author     Yagiz Nizipli <yagiz@nizipli.com> (@anonrig)
Branch     anonrig:url-context -> nodejs:main
Labels     semver-major, performance, whatwg-url, esm, author ready, worker, needs-ci
Commits    1
 - url: use private properties for brand check
Committers 1
 - Yagiz Nizipli <yagiz@nizipli.com>
------------------------------ Generated metadata ------------------------------
PR-URL: https://github.com/nodejs/node/pull/46904
Reviewed-By: Antoine du Hamel <duhamelantoine1995@gmail.com>
Reviewed-By: Ruben Bridgewater <ruben@bridgewater.de>
Reviewed-By: Benjamin Gruenbaum <benjamingr@gmail.com>
--------------------------------------------------------------------------------
   ℹ  This PR was created on Wed, 01 Mar 2023 21:28:43 GMT
   ✔  Approvals: 3
   ✔  - Antoine du Hamel (@aduh95) (TSC): https://github.com/nodejs/node/pull/46904#pullrequestreview-1320667498
   ✔  - Ruben Bridgewater (@BridgeAR) (TSC): https://github.com/nodejs/node/pull/46904#pullrequestreview-1320668469
   ✔  - Benjamin Gruenbaum (@benjamingr): https://github.com/nodejs/node/pull/46904#pullrequestreview-1321281126
   ✖  This PR needs to wait 21 more hours to land
   ✖  GitHub CI is still running
   ℹ  Last Full PR CI on 2023-03-02T23:13:10Z: https://ci.nodejs.org/job/node-test-pull-request/50175/
✔  Build data downloaded
   ✖  Last Jenkins CI still running
```

aduh95:
It’s always been the case, and I think it aligns with our rules defined in `docs/contributing`. The CQ will refuse to land a PR that had commits pushed to it since the last review though.

anonrig:
[...]

PR #680, opened 2023-03-08
fix(pr_checker): do not count non-approving reviews

aduh95:
Before this commit, any review after the last commit would disable the warning and let the CQ land the PR. This commit makes sure that only approving reviews from collaborators are taken into account to decide if a PR is ready to land.

Refs: https://github.com/nodejs/node-core-utils/issues/677#issuecomment-1454263877
Fixes: #643


aduh95:
ping @nodejs/node-core-utils

anonrig (approved):

PR #682, opened 2023-03-08
chore(deps): replace `eslint-plugin-node` with `eslint-plugin-n`

aduh95:
Without it, I'm unable to run the linter locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "eslint": "^8.35.0",
     "eslint-config-standard": "^17.0.0",
     "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-n": "^15.6.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^6.1.1",
     "mocha": "^10.2.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -59,6 +59,7 @@
     "eslint": "^8.35.0",
     "eslint-config-standard": "^17.0.0",
     "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-n": "^15.6.1",
     "eslint-plugin-node": "^11.1.0",
```
VoltrexKeyva, line 63: We should remove `eslint-plugin-node` in favor of `eslint-plugin-n` as the former is no maintained.

aduh95: Do you have a link that references that?

VoltrexKeyva: I don't have a direct link referencing that but `eslint-plugin-node` hasn't been updated in 3 years, and `eslint-plugin-n` also mentions it at the very top of their README, which references this issue: https://github.com/mysticatea/eslint-plugin-node/issues/300

aduh95: I agree it doesn't make sense to keep both.
```suggestion
    "eslint-plugin-n": "^15.6.1",
```

VoltrexKeyva (approved):

PR #684, opened 2023-03-08
feat(git-node): add support for the `--gpg-sign` git flag

aduh95:
I've also extended the max line length from 80 to 100 as it was making less readable to keep it at 80 IMO.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,7 +5,7 @@
     "space-before-function-paren": ["error", "never"],
     "no-multi-spaces": ["error", { "ignoreEOLComments": true }],
     "camelcase": "off",
-    "max-len": [2, 80, 4, {"ignoreRegExpLiterals": true, "ignoreUrls": true}],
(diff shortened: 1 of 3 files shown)


targos:
I personally set this in my gut config. What's the advantage of this flag compared to it ?

aduh95:
The flag let's you select a specific keyid to sign the commit with.

targos:
In the git config you can set ˋuser.signingkey` with the key I'd

aduh95:
If you have set a specific key id in your git config, the flag lets you specify a different one.

targos:
I still don't understand the use case. Don't you always use the same key for the node repo?

aduh95:
No, I usually don't sign my commits for the node repo (why would I, the CQ would sign them for me), but in the rare cases when I need to push manually to `main`, I have to do the `git commit --amend --no-edit -S` dance, and it seems easy to add the flag to ncu instead.

anonrig (approved):

PR #685, opened 2023-03-08
chore!: remove unmarked DEP0XXX handler

aduh95:
The deprecation numbers are linted by the `test-doc` test suite, and using `DEP0XXX` would be flagged as invalid before the PR can land.


richardlau:
> The deprecation numbers are linted by the `test-doc` test suite, and using `DEP0XXX` would be flagged as invalid before the PR can land.

Is that true on all supported release lines?

aduh95:
> > The deprecation numbers are linted by the `test-doc` test suite, and using `DEP0XXX` would be flagged as invalid before the PR can land.
> 
> Is that true on all supported release lines?

It seems that https://github.com/nodejs/node/pull/41992 hasn't been backported to v14.x, so I guess this should wait for v14.x to be officially EOL.

richardlau:
> > > The deprecation numbers are linted by the `test-doc` test suite, and using `DEP0XXX` would be flagged as invalid before the PR can land.
> > 
> > 
> > Is that true on all supported release lines?
> 
> It seems that [nodejs/node#41992](https://github.com/nodejs/node/pull/41992) hasn't been backported to v14.x, so I guess this should wait for v14.x to be officially EOL.

Thank you for checking. I suspect it's highly unlikely we'll deprecate something new in Node.js 14 before it reaches End-of-Life. I'd be okay landing if other @nodejs/releasers agree, otherwise we can wait.

targos:
SGTM

RafaelGSS (approved):

juanarbol (approved):
LGTM

BethGriggs (approved):

richardlau (approved):

targos:
@aduh95 there's a conflict now.

PR #688, opened 2023-03-22
fix(git-node): update README parser to account the recent TSC changes

aduh95:
Refs: https://github.com/nodejs/node/pull/47126
Refs: https://github.com/nodejs/TSC/pull/1350

BREAKING CHANGE: Old README format is no longer supported

### DIFF
--- a/lib/collaborators.js
+++ b/lib/collaborators.js
@@ -1,9 +1,10 @@
 import fs from 'node:fs';
 
-const TSC_TITLE = '### TSC (Technical Steering Committee)';
-const TSCE_TITLE = '### TSC Emeriti';
+const TSC_TITLE = '#### TSC voting members';
+const TSC_REGULAR_TITLE = '#### TSC regular members';
+const TSCE_TITLE = '#### TSC emeriti members';
 const CL_TITLE = '### Collaborators';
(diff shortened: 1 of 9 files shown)


richardlau:
> Refs: [nodejs/node#46985](https://github.com/nodejs/node/pull/46985)
> 
> BREAKING CHANGE: Old README format is no longer supported

Is that going to break using node-core-utils to land PR's in LTS lines? cc @nodejs/releasers

aduh95:
> Is that going to break using node-core-utils to land PR's in LTS lines?

I was about to comment regarding that. We need to either backport the README recent updates first, or backporters would need to use the `main` version of the README (which might be the default?).

Trott:
Not sure if it will help here, but here's what I did to make find-inactive-tsc.mjs backwards compatible: https://github.com/nodejs/node/blob/a4808b6c3b111608e0b9945c7f62534b285b0c96/tools/find-inactive-tsc.mjs#L56-L86

aduh95:
I think we should land this, the README change have landed on all supported release lines. wdyt @nodejs/node-core-utils?

RafaelGSS (approved):
LGTM

PR #697, opened 2023-05-03
fix: accurately define ncu-ci report condition

tniessen:
Problems are ignored if they occurred in less than two PRs. Thus, they are reported if they failed in two or more PRs.

https://github.com/nodejs/node-core-utils/blob/ecb1b7d0b86ab2afb581046220e3cfcc6254255a/lib/ci/failure_aggregator.js#L88-L89

### DIFF
--- a/lib/ci/failure_aggregator.js
+++ b/lib/ci/failure_aggregator.js
@@ -72,7 +72,7 @@ export class FailureAggregator {
     let output = 'Failures in ';
     output += `[${jobName}/${first.jobid}](${first.link}) to `;
     output += `[${jobName}/${last.jobid}](${last.link}) `;
-    output += 'that failed more than 2 PRs\n';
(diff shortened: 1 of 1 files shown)


lib/ci/failure_aggregator.js
```diff
@@ -72,7 +72,7 @@ export class FailureAggregator {
     let output = 'Failures in ';
     output += `[${jobName}/${first.jobid}](${first.link}) to `;
     output += `[${jobName}/${last.jobid}](${last.link}) `;
-    output += 'that failed more than 2 PRs\n';
+    output += 'that failed 2 or more PRs\n';
```
MoLow, line 75:
```suggestion
    output += 'that failed multiple PRs\n';
```

MoLow (approved):
whatever wording we go with, LGTM

PR #698, opened 2023-05-30
chore: update dependencies

VoltrexKeyva:
- Update the `replace-in-file` dependency to its new major version.
- Update the `eslint-plugin-n` developer dependency to its new major version.
- Update all other dependencies and developer dependencies to their latest versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "dependencies": {
     "branch-diff": "^2.1.1",
     "chalk": "^5.2.0",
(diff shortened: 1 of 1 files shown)


targos:
Have you checked the changelog of `replace-in-file` and assessed if the breaking change may affect us?

VoltrexKeyva:
> Have you checked the changelog of `replace-in-file` and assessed if the breaking change may affect us?

I did, it doesn't affect us.

PR #700, opened 2023-06-01
chore!: switch to `@node-core` scope

aduh95:
Not sure this is ready yet, we may need a new NPM_TOKEN on the repo secret.

Refs: https://github.com/nodejs/TSC/issues/1178

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -15,7 +15,7 @@ jobs:
         id: release
         with:
           release-type: node
-          package-name: node-core-utils
+          package-name: '@node-core/utils'
   npm-publish:
     needs: release-please
     if: ${{ needs.release-please.outputs.release_created }}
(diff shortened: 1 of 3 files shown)


MoLow:
we should probably also publish one last version to https://www.npmjs.com/package/node-core-utils with a output/warning saying it moved

aduh95:
> we should probably also publish one last version to https://www.npmjs.com/package/node-core-utils with a output/warning saying it moved

I think we can `npm deprecate` it without publishing a new version.

targos:
The current NPM_TOKEN should work. I can replace it with a granular one when the package exists on npm.

MoLow (dismissed):

targos:
release-please workflow needs an update as well.

targos (approved):

MoLow (approved):

VoltrexKeyva (approved):

PR #704, opened 2023-06-14
feat(git-node): add `git node vote`

aduh95:
This PR adds `git node vote` command, which can be used to participate to a vote using Caritat. The idea of integrating that in ncu is that TSC folks are very likely to already have ncu installed and configured on their local machine, reducing the possible friction for a user to cast a vote, as they don't need to provide their GitHub credentials to make the API calls.

Code reviews are welcome, note that we should wait for the transfer of Caritat npm package to the `@node-core` scope to be completed before landing.


components/git/vote.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+};
+
+let yargsInstance;
+
+export function builder(yargs) {
+  yargsInstance = yargs;
+  return yargs
+    .options(voteOptions)
+    .positional('prid', {
```
MoLow, line 44: should this be a number, like in `git node land`?

lib/voting_session.js
```diff
@@ -0,0 +1,125 @@
...
+
+import {
+  runAsync
+} from './run.js';
+import Session from './session.js';
+import {
+  getEditor
+} from './utils.js';
+
+import voteUsingGit from '@aduh95/caritat/voteUsingGit';
```
aduh95, line 13:
```suggestion
import voteUsingGit from '@node-core/caritat/voteUsingGit';
```

package.json
```diff
@@ -34,6 +34,7 @@
   ],
   "license": "MIT",
   "dependencies": {
+    "@aduh95/caritat": "^0.6.0",
```
aduh95, line 37:
```suggestion
    "@node-core/caritat": "^0.6.0",
```

components/git/vote.js
aduh95: It can be a number, however I don't expect much folks to use it as such (because there won't be any vote PR targeted to nodejs/node)

MoLow: Ok, that makes sense

MoLow (approved):

lib/voting_session.js
```diff
@@ -0,0 +1,125 @@
...
+    if (repository.pullRequest.merged) {
+      console.warn('The pull request appears to have been merged already.');
+    } else if (repository.pullRequest.closed) {
+      console.warn('The pull request appears to have been closed already.');
+    }
+    if (this.closeVote) return this.decryptKeyPart(repository.pullRequest);
+    const username = (await runAsync('git', ['config', '--get', 'user.name'],
+      { captureStdout: true })).trim();
+    const emailAddress = (await runAsync('git', ['config', '--get', 'user.email'],
+      { captureStdout: true })).trim();
```
aduh95, line 52:
```suggestion
    // @see https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_committing
    const username = process.env.GIT_AUTHOR_NAME || (await runAsync(
      'git', ['config', '--get', 'user.name'], { captureStdout: true })).trim();
    const emailAddress = process.env.GIT_AUTHOR_EMAIL || (await runAsync(
      'git', ['config', '--get', 'user.email'], { captureStdout: true })).trim();
```

aduh95:
@nodejs/node-core-utils given this PR is relatively large, it would help to have some extra pair(s) of eyes to review it :)

anonrig:
Since this is specific to tsc, why not make it “git node tsc-vote”?

components/git/vote.js
```diff
@@ -0,0 +1,84 @@
...
+      describe: 'URL of the vote Pull Request'
+    })
+    .example('git node vote https://github.com/nodejs/TSC/pull/12344',
+      'Start an interactive session to cast ballot for https://github.com/nodejs/TSC/pull/12344. ')
+    .example('git node vote https://github.com/nodejs/TSC/pull/12344 --abstain',
+      'Cast an empty ballot for https://github.com/nodejs/TSC/pull/12344')
+    .example('git node vote https://github.com/nodejs/TSC/pull/12344 --decrypt-key-part',
+      'Uses gpg to decrypt a key part to close the vote happening on https://github.com/nodejs/TSC/pull/12344');
```
VoltrexKeyva, line 52:
```suggestion
      'Start an interactive session to cast ballot for https://github.com/nodejs/TSC/pull/12344.')
    .example('git node vote https://github.com/nodejs/TSC/pull/12344 --abstain',
      'Cast an empty ballot for https://github.com/nodejs/TSC/pull/12344.')
    .example('git node vote https://github.com/nodejs/TSC/pull/12344 --decrypt-key-part',
      'Uses gpg to decrypt a key part to close the vote happening on https://github.com/nodejs/TSC/pull/12344.');
```

components/git/vote.js
```diff
@@ -0,0 +1,84 @@
+import auth from '../../lib/auth.js';
+import { parsePRFromURL } from '../../lib/links.js';
+import CLI from '../../lib/cli.js';
+import Request from '../../lib/request.js';
+import { runPromise } from '../../lib/run.js';
+import VotingSession from '../../lib/voting_session.js';
+
+export const command = 'vote [prid|options]';
+export const describe =
+  'Manage the current landing session or start a new one for a pull request';
```
joyeecheung, line 10: This doesn't seem quite right ;)

components/git/vote.js
```diff
@@ -0,0 +1,84 @@
...
+      'Uses gpg to decrypt a key part to close the vote happening on https://github.com/nodejs/TSC/pull/12344');
+}
+
+export function handler(argv) {
+  if (argv.prid) {
+    const parsed = parsePRFromURL(argv.prid);
+    if (parsed) {
+      Object.assign(argv, parsed);
+      return vote(argv);
+    }
```
anonrig, line 61: Should we give a proper error message here?

lib/voting_session.js
```diff
@@ -0,0 +1,126 @@
...
+      })
+    );
+    ac.abort();
+
+    const keyPart = out.toString('base64');
+    console.log('Your key part is', keyPart);
+    if (this.postComment) {
+      const { html_url } = await this.req.json(`https://api.github.com/repos/${this.owner}/${this.repo}/issues/${this.prid}/comments`, {
+        agent: this.req.proxyAgent,
+        method: 'POST',
```
joyeecheung, line 110: I think this should require additional permissions from the token? So the docs should probably be updated and/or there should be a hint if this returns a permission error code (403?)?

lib/voting_session.js
```diff
@@ -0,0 +1,126 @@
...
+          once(cp, 'error').then((er) => Promise.reject(er))
+        ]);
+        if (code !== 0) throw new Error('failed', { cause: code });
+        return Buffer.concat(await stdout);
+      })
+    );
+    ac.abort();
+
+    const keyPart = out.toString('base64');
+    console.log('Your key part is', keyPart);
```
anonrig, line 106: Nit: Is this intended, and/or useful?

joyeecheung (commented):
Can you add some docs to https://github.com/nodejs/node-core-utils/blob/main/docs/git-node.md?

anonrig (approved):

components/git/vote.js
aduh95: It's going to fall through the the `showHelp`, which should be enough to help folks understand what's wrong. Or do you have something else in mind?

lib/voting_session.js
aduh95: Yes, if one doesn't use `--post-comment`, they can copy/paste it themself in a comment.

components/git/vote.js
aduh95: Good catch!

lib/voting_session.js
```diff
@@ -0,0 +1,136 @@
...
+    ac.abort();
+    this.cli.stopSpinner('Found one key part.');
+
+    const keyPart = '-----BEGIN SHAMIR KEY PART-----\n' +
+                    out.toString('base64') +
+                    '\n-----END SHAMIR KEY PART-----';
+    this.cli.log('Your key part is:');
+    this.cli.log(keyPart);
+    const body = 'I would like to close this vote, and for this effect, I\'m revealing my ' +
+                 `key part:\n\n${'```'}\n${keyPart}\n${'```'}\n`
```
aduh95, line 115:
```suggestion
                 `key part:\n\n${'```'}\n${keyPart}\n${'```'}\n`;
```

Issue #686, opened 2023-03-14
Broken `git node release --prepare`?


targos:
Not sure when it started, but it seems that the command now generates an invalid changelog file:
- Wrong format for the list of commits
- Doesn't update the versions table in `CHANGELOG_Vx.md`

Maybe related to https://github.com/nodejs/node-core-utils/pull/669
Refs: https://github.com/nodejs/node/pull/47087

targos:
I fixed it. You can see the diff here: https://github.com/nodejs/node/compare/fd4b40e8edded567e941a12b92db78aadae5cec7..4221dce8a5b47087a331762cb0dfefb7b708ebf1

richardlau:
> Not sure when it started, but it seems that the command now generates an invalid changelog file:
> 
> * Wrong format for the list of commits
> 
> Maybe related to #669 Refs: [nodejs/node#47087](https://github.com/nodejs/node/pull/47087)

Most likely. That PR updated changelog-maker -- changelog-maker@3 changed the default output format: https://github.com/nodejs/changelog-maker/pull/121.

RafaelGSS:
Just opened two PRs that address this issue.

PR #711, opened 2023-07-03
fix: enforce markdown when generating changelog

RafaelGSS:
Refs: https://github.com/nodejs/node-core-utils/issues/686

### DIFF
--- a/lib/prepare_release.js
+++ b/lib/prepare_release.js
@@ -295,6 +295,7 @@ export default class ReleasePreparation {
 
     return runSync(changelogMaker, [
       '--group',
+      '--markdown',
       '--filter-release',
       '--start-ref',
       this.getLastRef()


lib/prepare_release.js
```diff
@@ -297,6 +297,7 @@ export default class ReleasePreparation {
       '--group',
       '--filter-release',
       '--start-ref',
+      '--markdown',
       this.getLastRef()
```
targos, line 301: This seems wrong. `this.getLastRef()` is the argument to the `--start-ref` flag.

RafaelGSS: Well, that worked when I tested it. I think since `--markdown` is a bool flag it won't affect the `--start-ref`. 
But, anyway, it's better to move one argument left and make it clear.

targos (approved):

richardlau (approved):

PR #712, opened 2023-07-03
fix: update changelog list when preparing release

RafaelGSS:
Refs: https://github.com/nodejs/node-core-utils/issues/686

### DIFF
--- a/lib/prepare_release.js
+++ b/lib/prepare_release.js
@@ -386,7 +386,7 @@ export default class ReleasePreparation {
                           ` ${releaseInfo}, @${username}\n`;
 
     const topHeader =
-      `<a href="#${lastRef.substring(1)}">${lastRef.substring(1)}</a><br/>`;
+      `<a href="#${lastRef.substring(1)}">${lastRef.substring(1)}</a>`;
     const newHeader =
       `<a href="#${newVersion}">${newVersion}</a><br/>`;
     for (let idx = 0; idx < arr.length; idx++) {


lib/prepare_release.js
```diff
@@ -386,9 +386,10 @@ export default class ReleasePreparation {
                           ` ${releaseInfo}, @${username}\n`;
 
     const topHeader =
-      `<a href="#${lastRef.substring(1)}">${lastRef.substring(1)}</a><br/>`;
+      `<b><a href="#${lastRef.substring(1)}">${lastRef.substring(1)}</a></b><br/>`;
```
richardlau, line 389: We may need to make this optional for it to work with all release lines.
Compare
- https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V16.md?plain=1
- https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md?plain=1
- https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md?plain=1

RafaelGSS: what if we just drop the <b> on v20?

richardlau: > what if we just drop the `<b>` on v20?

SGTM as well.

lib/prepare_release.js
```diff
@@ -386,7 +386,7 @@ export default class ReleasePreparation {
                           ` ${releaseInfo}, @${username}\n`;
 
     const topHeader =
-      `<a href="#${lastRef.substring(1)}">${lastRef.substring(1)}</a><br/>`;
+      `<a href="#${lastRef.substring(1)}">${lastRef.substring(1)}</a>`;
```
RafaelGSS, line 389: now it should match with or without <b>

richardlau (approved):

PR #721, opened 2023-08-22
chore: update dependencies

VoltrexKeyva:
- Update the `execa` dependency to its new major version.
- Update the other dependencies and developer dependencies to their latest versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,14 +34,14 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "branch-diff": "^2.1.3",
+    "branch-diff": "^2.1.4",
     "chalk": "^5.3.0",
     "changelog-maker": "^3.2.4",
     "cheerio": "^1.0.0-rc.12",
     "clipboardy": "^3.0.0",
     "core-validate-commit": "^4.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -34,14 +34,14 @@
...
-    "branch-diff": "^2.1.3",
+    "branch-diff": "^2.1.4",
     "chalk": "^5.3.0",
     "changelog-maker": "^3.2.4",
     "cheerio": "^1.0.0-rc.12",
     "clipboardy": "^3.0.0",
     "core-validate-commit": "^4.0.0",
     "enquirer": "^2.4.1",
-    "execa": "^7.2.0",
+    "execa": "^8.0.1",
```
aduh95, line 44: We should probably consider removing execa in favor of `run.js`

aduh95 (approved):

package.json
VoltrexKeyva: Oh nevermind, I thought you meant a library called `run.js` but you actually meant the `run.js` file in the repository 😅.

aduh95: Yes, that's what I meant. Sorry I missed your comment.

PR #724, opened 2023-09-07
feat(release): add skipBranchDiff option

ruyadorno:
Adds an option that allows skipping the initial `branch-diff` check step in the release command. This should be very convenient for releasers that are updating a current work in progress proposal, particularly helping avoid hitting GitHub API rate limits.

### DIFF
--- a/docs/git-node.md
+++ b/docs/git-node.md
@@ -247,6 +247,11 @@ git node release --prepare --startLTS
 git node release --prepare --security --filterLabel 18.x 18.20.1
 ```
 
+```
+# Skip the branch-diff initial check (useful when updating ongoing proposals)
+git node release --prepare 1.2.3 --skipBranchDiff
+```
+
 ## `git node sync`
 
 Demo: https://asciinema.org/a/221230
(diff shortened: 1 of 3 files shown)


docs/git-node.md
```diff
@@ -330,7 +335,7 @@ ncu-config set waitTimeMultiApproval 48
 
 ## `git node v8`
 
-Update or patch the V8 engine.  
+Update or patch the V8 engine.
```
RafaelGSS, line 338: revert it? I think I got the same behaviour when using npm run lint-fix.

RafaelGSS (approved):
LGTM

docs/git-node.md
targos: Refs: https://www.markdownguide.org/basic-syntax/#line-breaks

aduh95:
```suggestion
Update or patch the V8 engine.  
```

docs/git-node.md
```diff
@@ -367,7 +372,7 @@ Options:
 ### `git node v8 minor`
 
 Compare current V8 version with latest upstream of the same major. Applies a
-patch if necessary.  
+patch if necessary.
```
aduh95, line 375:
```suggestion
patch if necessary.  
```

Issue #723, opened 2023-09-06
V8 commands are broken


targos:
https://github.com/nodejs/node-core-utils/pull/722#issuecomment-1706027683

See https://github.com/nodejs/node-v8/actions/runs/6093305552/job/16532764329

I do not have time to fix it until next week.

targos:
Backport command is still broken:

```
$ git node v8 --v8-dir ~/git/chromium/v8/v8 backport 1a3ecc2483b2dba6ab9f7e9f8f4b60dbfef504b7
✔ Update local V8 clone
⠋ V8 commit backport
```

Exit code: 1

PR #722, opened 2023-09-03
chore: remove `execa` dependency

aduh95:
The less dependencies we have, the better.

PR #725, opened 2023-09-11
Revert "chore: remove `execa` dependency (#722)"

targos:
This reverts commit ff79809164bda99c14979efd955617e0c6eb7630.

The way `execa` worked is quite different from our `runAsync` function
and it's not easy to fix quickly. Revert the change to unbreak canary
V8 updates.

Closes: https://github.com/nodejs/node-core-utils/issues/723

### DIFF
--- a/components/git/v8.js
+++ b/components/git/v8.js
@@ -1,11 +1,11 @@
 import path from 'node:path';
 
+import { execa } from 'execa';
 import logSymbols from 'log-symbols';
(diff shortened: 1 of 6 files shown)


MoLow (approved):

aduh95:
I've opened https://github.com/nodejs/node-core-utils/pull/726 as an alternative for this.

PR #732, opened 2023-09-22
fix(backport): checkout from staging branch and ask to update if needed

### LINKED ISSUE #731: update requested staging branch before backporting in `git-node backport`
Currently, if my `staging` branch is behind `upstream` one or not on the `staging` branch at all I will create a bad backport PR

so we should ask do you want to sync `staging` with `upstream` (if it's behind)

and we should checkout from the staging branch

rluvaton:
Fixes #731

this aligns with steps 1-3 in [`doc/contributing/backporting-to-release-lines.md` - How to submit a backport pull request](https://github.com/nodejs/node/blob/main/doc/contributing/backporting-to-release-lines.md#how-to-submit-a-backport-pull-request)

https://github.com/nodejs/node/blob/c2cd74453e7d2794ad81cab63e68371e08bad04f/doc/contributing/backporting-to-release-lines.md?plain=1#L48-L62

### DIFF
--- a/lib/backport_session.js
+++ b/lib/backport_session.js
@@ -61,7 +61,7 @@ export default class BackportSession extends Session {
         cli.stopSpinner(`${file} does not exist in current working tree`,
           cli.SPINNER_STATUS.WARN);
         continue;
-      };
+      }
       if (ancestors.length === 0) {
         cli.stopSpinner(`Cannot find ancestor commits of ${file}`,
           cli.SPINNER_STATUS.INFO);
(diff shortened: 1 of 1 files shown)


lib/backport_session.js
```diff
@@ -155,12 +155,33 @@ export default class BackportSession extends Session {
...
+          defaultAnswer: true
+        });
+
+      if (shouldCreateStagingBranch) {
+        this.syncBranchWithUpstream(this.stagingBranch);
+      }
+    } else if (!this.isBranchUpToDateWithUpstream(this.stagingBranch)) {
+      const shouldSyncBranch = await cli.prompt(
+        `It seems like your ${this.stagingBranch} is behind the upstream remote` +
+        'do you want to sync it?', { defaultAnswer: false });
```
RafaelGSS, line 171:
```suggestion
        'do you want to sync it?', { defaultAnswer: true });
```

RafaelGSS (approved):

lib/backport_session.js
```diff
@@ -61,7 +61,7 @@ export default class BackportSession extends Session {
         cli.stopSpinner(`${file} does not exist in current working tree`,
           cli.SPINNER_STATUS.WARN);
         continue;
-      };
+      }
```
Trott, line 64: Unrelated style change? (I'm surprised we don't lint these files.)

Trott: (Although, yeah, that semicolon shouldn't be there. Maybe leave it and ignore my comment. But uh, future enhancement for someone: Add linting or adjust the current linting to catch that stuff automatically.)

rluvaton:
I totally with you on adding test but there where no test to begin with

I will try to add one...

lib/backport_session.js
```diff
@@ -226,4 +247,51 @@ export default class BackportSession extends Session {
...
+      return false;
+    }
+  }
+
+  syncBranchWithUpstream(branch) {
+    const currentBranch = this.getCurrentBranch();
+
+    runSync('git',
+      [currentBranch !== branch ? 'fetch' : 'pull', 'upstream', `${branch}:${branch}`, '-f']
+    );
```
aduh95, line 295:
```suggestion
    runSync('git', ['fetch', upstream, branch]);
    if (currentBranch === branch) runSync('git', ['reset', 'FETCH_HEAD', '--hard']);
```
I'm not even sure the `if` condition is necessary, but I guess I'm not using git like everyone else 🤷‍♂️

aduh95 (changes requested):
Please do not hard code `'upstream'`, and take the name of the upstream from the config instead.

lib/backport_session.js
rluvaton: the fetch here will not update the local branch...

rluvaton:
> Please do not hard code `'upstream'`, and take the name of the upstream from the config instead.

@aduh95 fixed

lib/backport_session.js
```diff
@@ -226,4 +247,56 @@ export default class BackportSession extends Session {
...
+
+  syncBranchWithUpstream(branch) {
+    const currentBranch = this.getCurrentBranch();
+
+    runSync('git',
+      [
+        currentBranch !== branch ? 'fetch' : 'pull',
+        this.upstream,
+        `${branch}:${branch}`,
+        '-f'
```
aduh95, line 298: Here's what I happen when I try the command locally:

```console
$ git fetch upstream main:main -f
fatal: refusing to fetch into branch 'refs/heads/main' checked out at '…/node'
```

Are you sure it does work?

rluvaton: It's weird, it's in the docs of backporting and it's working for me

Are you currently on main branch

lib/backport_session.js
aduh95: I don't use local branches, so I guess that's why this looks weird to me.

PR #733, opened 2023-09-22
chore: remove EoL Node.js 16 from test matrix

rluvaton:


RafaelGSS (approved):

benjamingr (approved):

VoltrexKeyva (approved):

Trott:
Can you please update the "engines" field in package.json as well?

Trott:
Totally not necessary for this PR but optional here or as a future enhancement: If we use the pkgjs prepare-node-test-matrix GitHub Action instead of hardcoding the Node.js version numbers, it will look at the "engines" property in package.json and only test the LTS versions at or above that level along with the current version. See https://github.com/pkgjs/action/tree/main for more information. (pkgjs is the tooling org managed in a very hands-off way by the TSC for Node.js.)

rluvaton:
@Trott moved to `prepare-node-test-matrix` 😄

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Utilities for Node.js core collaborators",
   "type": "module",
   "engines": {
-    "node": ">=16.8.0"
+    "node": ">=18.0.0"
```
aduh95, line 7:
```suggestion
    "node": "^18.18.0 || >=20.0.0"
```

rluvaton: why the 20 is needed?

aduh95: 20.x is the current release line, it makes sense to support it. My suggestion is to not use a range that includes 19.x (notice the `^` before `18.18`).

rluvaton: oh, I forgot 19 is dead

PR #759, opened 2023-12-04
feat: add support for nvm

UlisesGascon:


UlisesGascon:
> Not sure what's the point? We want the utils to work on all supported versions, so I think it's good if we have a variety of versions used by collaborators.

I think that is important to use minimum compatible version when developing the library so we prevent to introduce bugs. For example using Node@20 features that are not available in 18.18.0.

The consumers can use a variety of version defined by the engine https://github.com/nodejs/node-core-utils/blob/main/package.json#L7 independently 👍

targos:
Ok, I won't block it but I don't use nvm so I'll let someone else review. Please if it lands, don't use `feat:` (or `fix:`) in the commit message, as this change doesn't impact users who install the package AFAIK.

joyeecheung:
> For example using Node@20 features that are not available in 18.18.0.

That should be checked by the CI instead, I believe. The config won't do much for contributors to this repo who don't use nvm, then they won't catch it anyway...

Issue #752, opened 2023-11-13
CI fails on Node.js 21


targos:
See https://github.com/nodejs/node-core-utils/actions/runs/6847526575/job/18615958948?pr=751

targos:
@nodejs/test_runner Is there a way to execute the test runner with arguments that work in all supported Node.js versions?

targos:
@MoLow Do you know of a solution?

MoLow:
not a good one.
I would try changing the command based on the node version,
we can also run `node --test $(find . -name 'test/unit/**.js' | tr '\n' ' ')` which will be compatible for all versions

targos:
This is unfortunate. Isn't it a big issue for adoption of the test runner by projects that want to support all release lines?

MoLow:
It probably is, we should have probably thought about that - although this change was a semver major change, and was necessary since it made it possibe to run typescript tests or tests other then the default pattern

aduh95:
I guess we could backport it behind a `--experimental-glob-in-test-runner` flag or something like that

ljharb:
Major changes for a test runner built into the platform will always cause this problem; the only solution is to never have them, or, not have a test runner that’s built in.

PR #761, opened 2023-12-06
fix(ci): use shell expansion to run test files

aduh95:
Fixes: https://github.com/nodejs/node-core-utils/issues/752

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
   },
   "scripts": {
     "test": "npm run test-unit && npm run lint",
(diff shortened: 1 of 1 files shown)


aduh95:
Blocked on https://github.com/rvagg/ghauth/pull/31

targos (approved):
SGTM assuming we don't want to support Windows.

aduh95:
Yeah not sure why the CI doesn't run on Windows. I think it should work on Windows as long as you're using Node.js 21+ to run the tests.

PR #771, opened 2024-01-22
feat: automate security release issue creation

marco-ippolito:
This PR automates the first step of the security release: `security --start`.
The `Next Security Release` issue will not contain the list of reports but they will be stored in the new repository `security-release` in a json format.
This way it will be much easier to automate and manipulate json rather than markdown strings.
@RafaelGSS

### DIFF
--- a/lib/prepare_security.js
+++ b/lib/prepare_security.js
@@ -7,6 +7,7 @@ import path from 'node:path';
 
 export const PLACEHOLDERS = {
   releaseDate: '%RELEASE_DATE%',
+  vulnerabilitiesPRURL: '%VULNERABILITIES_PR_URL%',
   preReleasePrivate: '%PRE_RELEASE_PRIV%',
   postReleasePrivate: '%POS_RELEASE_PRIV%',
   affectedLines: '%AFFECTED_LINES%'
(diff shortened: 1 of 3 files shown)


lib/prepare_security.js
```diff
@@ -53,31 +94,39 @@ class SecurityReleaseIssue {
...
+  async promptReleaseDate(cli) {
+    return cli.prompt('Enter target release date in YYYY-MM-DD format:', {
       questionType: 'input',
       defaultAnswer: 'TBD'
     });
-    this.fillTargetDate(target);
+  }
 
-    return this.content;
+  async buildIssue(releaseDate) {
```
RafaelGSS, line 104:
```suggestion
  buildIssue(releaseDate) {
```

lib/prepare_security.js
```diff
@@ -16,31 +25,63 @@ export default class SecurityReleaseSteward {
     });
 
     const req = new Request(credentials);
-    const create = await cli.prompt(
+    const release = new PrepareSecurityRelease(req);
+    const releaseDate = await release.promptReleaseDate(cli);
```
RafaelGSS, line 29: We have `promptReleaseDate` and `cli.prompt(
      'Create the vulnerabilities.json?',
      { defaultAnswer: true });`

we should make a pattern here, either you remote the `promptReleaseDate` or you create `release.promptVulnerabilitiesJSON()`

lib/prepare_security.js
```diff
@@ -16,31 +25,63 @@ export default class SecurityReleaseSteward {
...
-        cli.ok('Created: ' + data.html_url);
-      } else {
-        cli.error(data);
-      }
+    if (createIssue) {
+      // prepare the security release issue content and release date
+      const { content } = await release.buildIssue(releaseDate);
+
+      // create the security release issue
+      await release.createIssue(content, { cli });
```
RafaelGSS, line 38: I think we could include a reference to the vulnerabilities.json PR in the main issue, so instead of creating the issue now, you do it after `createVulnerabilitiesJSON` and make it required instead of optional through a prompt.

lib/prepare_security.js
```diff
@@ -16,31 +25,63 @@ export default class SecurityReleaseSteward {
...
+      const reports = await release.chooseReports(cli);
+
+      // create the vulnerabilities.json file in the security-release repo
+      const filePath = await release.createVulnerabilitiesJSON(reports, { cli });
+
+      const review = await cli.prompt(
+        'Please review vulnerabilities.json and press enter to proceed.',
+        { defaultAnswer: true });
+
+      if (!review) return;
```
RafaelGSS, line 56: We should log the next steps since we aren't storing state.

lib/prepare_security.js
```diff
@@ -16,31 +25,63 @@ export default class SecurityReleaseSteward {
...
+
+      runSync('git', ['add', filePath]);
+
+      const commitMessage = 'chore: create vulnerabilities.json for next security release';
+      runSync('git', ['commit', '-m', commitMessage]);
+      runSync('git', ['push', '-u', 'origin', 'next-security-release']);
```
RafaelGSS, line 62: We should check the remote first and I suggest moving it to a new method

lib/prepare_security.js
```diff
@@ -16,31 +25,63 @@ export default class SecurityReleaseSteward {
...
+      cli.ok(`Pushed commit: ${commitMessage}`);
```
RafaelGSS, line 63:
```suggestion
      cli.ok(`Pushed commit: ${commitMessage} to ${branchName}`);
```

lib/prepare_security.js
```diff
@@ -16,31 +25,63 @@ export default class SecurityReleaseSteward {
...
+      await release.createPullRequest(req, { cli });
     }
+    cli.ok('Done!');
   }
 }
 
-class SecurityReleaseIssue {
+class PrepareSecurityRelease {
   constructor(req) {
+    this.repository = {
```
RafaelGSS, line 79: We should receive it as arguments and default to nodejs-private

RafaelGSS (approved):

PR #772, opened 2024-01-30
feat: update security release date

marco-ippolito:
this pr includes an overall code refactor, creating a some utilities and  allows us to update the vulnerabilities json with the target date for the security release


components/git/security.js
```diff
@@ -17,17 +22,31 @@ export function builder(yargs) {
   yargsInstance = yargs;
   return yargs.options(securityOptions).example(
     'git node security --start',
-    'Prepare a security release of Node.js');
+    'Prepare a security release of Node.js')
+    .example(
+      'git node security --finalize',
+      'Finalize the date of the security release of Node.js'
```
RafaelGSS, line 28: The name doesn't sound correct. We should make it more clear, like: `git node security --update=date`

To make it applicable for future updates, we should make it generic enough.

lib/finalize_security_release.js
```diff
@@ -0,0 +1,68 @@
+import {
+  NEXT_SECURITY_RELEASE_FOLDER,
+  NEXT_SECURITY_RELEASE_REPOSITORY,
+  checkoutOnSecurityReleaseBranch,
+  commitAndPushVulnerabilitiesJSON
+} from './security-release/security-release.js';
+import fs from 'node:fs';
+import path from 'node:path';
+
+export default class FinalizeSecurityRelease {
```
RafaelGSS, line 10:
```suggestion
export default class UpdateSecurityRelease {
```

we could rename it to `UpdateSecurityRelease`

then the correct `start` function transform into `updateReleaseDate`

lib/finalize_security_release.js
```diff
@@ -0,0 +1,68 @@
...
+    commitAndPushVulnerabilitiesJSON(updatedVulnerabilitiesFiles,
+      commitMessage, { cli, repository: this.repository });
+    cli.ok('Done!');
+  }
+
+  async updateReleaseDate(releaseDate, { cli }) {
+    const vulnerabilitiesJSONPath = path.join(process.cwd(),
+      NEXT_SECURITY_RELEASE_FOLDER, 'vulnerabilities.json');
+
+    const content = JSON.parse(fs.readFileSync(vulnerabilitiesJSONPath, 'utf8'));
```
RafaelGSS, line 37: Can you `fs.existSync` so we provide a meaningful error when the file for some reason doesn't exist?

lib/finalize_security_release.js
```diff
@@ -0,0 +1,68 @@
...
+    content.releaseDate = releaseDate;
+
+    fs.writeFileSync(vulnerabilitiesJSONPath, JSON.stringify(content, null, 2));
+
+    cli.ok(`Updated the release date in vulnerabilities.json: ${releaseDate}`);
+
+    const newFolderPath = path.join(process.cwd(),
+      'security-release', releaseDate.replaceAll('/', '-'));
```
RafaelGSS, line 45: I think this change should only happen **after** the security release is concluded. So I would put it in a different command.

lib/security-release/security-release.js
```diff
@@ -0,0 +1,55 @@
...
+  postReleasePrivate: '%POS_RELEASE_PRIV%',
+  affectedLines: '%AFFECTED_LINES%'
+};
+
+export function checkRemote(cli, repository) {
+  const remote = runSync('git', ['ls-remote', '--get-url', 'origin']).trim();
+  const { owner, repo } = repository;
+  const securityReleaseOrigin = `https://github.com/${owner}/${repo}.git`;
+
+  if (remote !== securityReleaseOrigin) {
```
RafaelGSS, line 24: See: https://github.com/nodejs/node-core-utils/pull/778

RafaelGSS (commented):
Can you update the PR description elaborating on what this PR is about? Thanks

components/git/security.js
```diff
@@ -8,6 +9,10 @@ const securityOptions = {
   start: {
     describe: 'Start security release process',
     type: 'boolean'
+  },
+  'update-date': {
```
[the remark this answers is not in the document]
marco-ippolito: I dont like it to be too generic, I immagine we would want to update other things in the future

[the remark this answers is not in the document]
RafaelGSS: >  I immagine we would want to update other things in the future

That's the idea of being generic.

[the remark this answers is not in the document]
marco-ippolito: can you give me an example on how you want to use it?
I immagined `git node security --update-date=21/12/2023` or you want to do it with cli prompt?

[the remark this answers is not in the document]
RafaelGSS: I thought using the prompt for that. But I believe it might make things harder to automate. So, let's keep update-date then

components/git/security.js
```diff
@@ -17,17 +22,32 @@ export function builder(yargs) {
...
+      'Updates the target date of the security release'
+    );
 }
 
 export function handler(argv) {
   if (argv.start) {
     return startSecurityRelease(argv);
   }
+  if (argv['update-date']) {
+    return updateReleaseDate(argv);
```
RafaelGSS, line 37: Can we validate if the string provided is correct?

PR #776, opened 2024-03-06
chore: update GH actions used in workflows

targos:


.github/workflows/nodejs.yml
```diff
@@ -29,6 +29,6 @@ jobs:
       test-command: npm run coverage:ci
       post-test-steps: |
         - name: Upload coverage report to Codecov
-          uses: codecov/codecov-action@v3
-    strategy:
-      fail-fast: false
```
targos, line 34: This was wrong. Fail-fast is setup inside the pkgjs action and is `false` by default: https://github.com/pkgjs/action/blob/main/docs/USAGE.md#strategy-fail-fast

.github/workflows/nodejs.yml
```diff
@@ -29,6 +29,6 @@ jobs:
...
+          uses: codecov/codecov-action@v4
+          with:
+            token: ${{ secrets.CODECOV_TOKEN }}
```
targos, line 34: Looks like secrets are not allowed here: https://github.com/nodejs/node-core-utils/actions/runs/8170900413
I'll remove the codecov update from this PR.

richardlau (approved):

PR #788, opened 2024-03-26
feat: prompt dependency updates url in vulnerabilities.json creation

marco-ippolito:
Fixes: https://github.com/nodejs-private/security-release/issues/9

I think I should also create an automation to add/remove them.

cc @RafaelGSS


lib/prepare_security.js
```diff
@@ -261,4 +266,26 @@ class PrepareSecurityRelease {
...
+    const updates = await cli.prompt('Are there dependency updates in this security release?', {
+      defaultAnswer: true,
+      questionType: 'confirm'
+    });
+
+    if (!updates) return deps;
+
+    let asking = true;
+    while (asking) {
+      const dep = await promptDependencies(cli);
```
marco-ippolito, line 282: we might wanna fetch it and get save something like:
```json
{
"title": "deps: update foo to bar",
"link": "http://foobar"
}
```

lib/prepare_security.js
```diff
@@ -48,15 +51,16 @@ export default class SecurityReleaseSteward {
     cli.ok('Done!');
   }
 
-  async createVulnerabilitiesJSON(req, release, { cli }) {
+  async createVulnerabilitiesJSON(req, release, releaseDate, { cli }) {
```
RafaelGSS, line 54: I think this PR does more than prompt dependency updates, right? You are managing the releaseDate

marco-ippolito: Yeah it adds the releaseDate in the vulnerabilities.json, but its a minor thing

RafaelGSS: may you split it into a different commit at least?

marco-ippolito: removed it, will do in a different pr

marco-ippolito:
@RafaelGSS this is ready to land

RafaelGSS (approved):

PR #790, opened 2024-03-27
fix: pre-release crashing on missing vulnerability rating

marco-ippolito:
Fixes: https://github.com/nodejs-private/security-release/issues/13


lib/prepare_security.js
```diff
@@ -120,7 +120,7 @@ class PrepareSecurityRelease {
   async promptReleaseDate(cli) {
     return cli.prompt('Enter target release date in YYYY/MM/DD format:', {
       questionType: 'input',
-      defaultAnswer: 'TBD'
+      defaultAnswer: ''
```
RafaelGSS, line 123: We should default to Date.now() + 7 days ~

RafaelGSS (approved):

PR #796, opened 2024-04-14
feat: add releaseDate to json

marco-ippolito:
Fixes: https://github.com/nodejs-private/security-release/issues/23

@RafaelGSS

### DIFF
--- a/lib/prepare_security.js
+++ b/lib/prepare_security.js
@@ -33,7 +33,10 @@ export default class SecurityReleaseSteward {
     const req = new Request(credentials);
     const release = new PrepareSecurityRelease(req);
     const releaseDate = await release.promptReleaseDate(cli);
-    validateDate(releaseDate);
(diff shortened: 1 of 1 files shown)


RafaelGSS (approved):

RafaelGSS (commented):
actually, we need to check if a valid date is given, if not it should leave it empty. For instance, `TBD`.

marco-ippolito:
> actually, we need to check if a valid date is given, if not it should leave it empty. For instance, `TBD`.
if invalid date is given it will throw right now

RafaelGSS:
> > actually, we need to check if a valid date is given, if not it should leave it empty. For instance, `TBD`.
> 
> if invalid date is given it will throw right now

But, what if we don't have a target date? I would like to pass TBD

marco-ippolito:
ok so I'll validate only if different from 'TBD'

RafaelGSS (approved):

PR #806, opened 2024-05-08
feat(ncu-ci): check for `request-ci` label

aduh95:
When commits were pushed to a PR since the last approving review, check if the last time a `request-ci` label was added, it was done by a Collaborator and after the last push event on the PR.

I don't really like the fact `"request-ci"` is hard-coded, and also it might make sense to enable this check only in the GHA workflow (as the code doesn't search for `'unlabeled'` events, so if someone adds the label and removes it, it would still still count), so if you have suggestions regarding that, I'd be interested.

Fixes: https://github.com/nodejs/node-core-utils/issues/801

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -526,14 +526,14 @@ export default class PRChecker {
   async checkCommitsAfterReviewOrLabel() {
     if (this.checkCommitsAfterReview()) return true;
 
-    await Promise.all([this.data.getLabels(), this.data.getCollaborators()]);
+    await Promise.all([this.data.getLabeledEvents(), this.data.getCollaborators()]);
 
     const {
       cli, data, pr
     } = this;
 
     const { updatedAt } = pr.timelineItems;
-    const requestCiLabels = data.labels.findLast(
+    const requestCiLabels = data.labeledEvents.findLast(
       ({ createdAt, label: { name } }) => name === 'request-ci' && createdAt > updatedAt
     );
     if (requestCiLabels == null) return false;
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -523,6 +523,28 @@ export default class PRChecker {
     return true;
   }
 
+  async checkCommitsAfterReviewOrLabel() {
+    if (this.checkCommitsAfterReview()) return true;
+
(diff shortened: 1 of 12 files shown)


MoLow (approved):
LGTM. can you add tests?

aduh95:
> can you add tests?

Tests would be great, I'd prefer if someone else could contribute that, I'm not sure when I'd have time to work on that

lib/pr_data.js
```diff
@@ -90,6 +91,14 @@ export default class PRData {
     ]);
   }
 
+  async getLabels() {
```
fahrradflucht, line 94: I was confused for a moment why this extra method is needed, given that labels are already requested with the PR, but that's because this isn't actually getting the _labels_, but the _labled events_ from the timeline. I think this should be renamed. (Would apply to the query and the property as well.)

lib/pr_checker.js
```diff
@@ -523,6 +523,32 @@ export default class PRChecker {
     return true;
   }
 
+  async checkCommitsAfterReviewOrLabel() {
+    if (this.checkCommitsAfterReview()) return true;
+
+    await Promise.all([this.data.getLabels(), this.data.getCollaborators()]);
```
fahrradflucht, line 529: Sorry, delayed second thought: In general, why did you decide to fetch the labeled events lazily and not with the rest of the timeline events on the PR? If there is no particular reason, it looks to me like just extending the PR query would be less fragile _and_ easier to test.

aduh95: How would that change the testability? Making extra requests will make it more likely that the bot will reach its quota of requests and be rate limited, making the requests lazilly seems like a low-hanging fruit.

fahrradflucht: With "easier to test" I didn't mean it is "hard to test" this way either, just that the work of extending the existing stubs with some values and having all the data present synchronously vs testing lazy fetching would be slightly less. Anyway, the rate limit is a fine reason. 👍

fahrradflucht:
I created a PR against this branch on your fork (https://github.com/aduh95/node-core-utils/pull/1) 🙂

targos (approved):

PR #807, opened 2024-05-11
fix(v8): remove trace_event starting with 12.6

targos:
Also add a custom error message when a dependency disappears to avoid
the unhelpful `Cannot read properties of undefined (reading 'split')`.

Refs: https://github.com/v8/v8/commit/f8fa22073a0db4a62775efba87ddee5f949a2095


targos:
Should fix: https://github.com/nodejs/node-v8/actions/runs/9041586116/job/24847117683

richardlau (approved):

PR #809, opened 2024-05-12
fix(v8): do not ignore failures when running git

targos:
Also always use `forceRunAsync` because `runAsync` exits the process on error.


targos:
@nodejs/node-core-utils

targos:
I tested these scenarios locally:
- Major update
- Minor update (already up-to-date)
- Minor update (with patch)
- Backport (patch applies cleanly)
- Backpor (patch doesn't apply cleanly)

PR #812, opened 2024-05-22
chore: update dependencies

VoltrexKeyva:
- Update the `sinon` development dependency to its new major version.
- Update the other dependencies and development dependencies to their latest versions.


VoltrexKeyva:
I'm holding off on updating ESLint to v9 and its plugins and configs because `eslint-config-standard` doesn't support ESLint v9 yet.

targos (approved):

aduh95 (approved):

PR #814, opened 2024-06-07
fix(vote): improve UX when posting comment fails

aduh95:
When the token doesn't have the permission to post the comment directly, we can fallback to showing the user the error message and what command they should type.

### DIFF
--- a/lib/voting_session.js
+++ b/lib/voting_session.js
@@ -114,7 +114,7 @@ export default class VotingSession extends Session {
     const body = 'I would like to close this vote, and for this effect, I\'m revealing my ' +
                  `key part:\n\n${'```'}\n${keyPart}\n${'```'}\n`;
     if (this.postComment) {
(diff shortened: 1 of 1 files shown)


lib/voting_session.js
```diff
@@ -124,13 +124,23 @@ export default class VotingSession extends Session {
...
         body: JSON.stringify({ body })
       });
-      this.cli.log('Comment posted at:', html_url);
-    } else if (isGhAvailable()) {
+      if (html_url) {
+        this.cli.log(`Comment posted at: ${html_url}`);
+        return;
+      } else {
+        this.cli.error('Failed to post comment');
+        this.cli.info(message);
```
aduh95, line 132:
```suggestion
        this.cli.warn(message);
        this.cli.error('Failed to post comment');
```

MoLow (approved):

PR #815, opened 2024-06-07
lib: add PR_URL to vuln.json and fetch from H1

RafaelGSS:
Fixes: https://github.com/nodejs-private/security-release/issues/30

Refs: https://github.com/nodejs/security-wg/issues/860

### DIFF
--- a/lib/security-release/security-release.js
+++ b/lib/security-release/security-release.js
@@ -142,7 +142,7 @@ export async function createIssue(title, content, repository, { cli, req }) {
 export async function pickReport(report, { cli, req }) {
   const {
     id, attributes: { title, cve_ids },
-    relationships: { severity, weakness, reporter }
+    relationships: { severity, weakness, reporter, custom_field_values }
   } = report;
   const link = `https://hackerone.com/reports/${id}`;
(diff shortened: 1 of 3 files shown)


lib/security-release/security-release.js
```diff
@@ -165,16 +165,24 @@ export async function pickReport(report, { cli, req }) {
...
-    'Add github username of the authors of the patch (split by comma if multiple)', {
-      questionType: 'input',
-      defaultAnswer: ''
-    });
-
-  if (!patchAuthors) {
-    patchAuthors = [];
+  let prURL = '';
+  let patchAuthors = [];
+  if (custom_field_values.data.length) {
```
marco-ippolito, line 170: If the custom attribute is not set it will throw, we might need to check if  present
```suggestion
  if (custom_field_values?.data.length) {
```

RafaelGSS: The field `custom_field_values` will always exist, the difference is if it's empty (no custom fields). I have tested it with a couple of different reports (with and without the custom field).

marco-ippolito (approved):

PR #829, opened 2024-06-26
chore: update dependencies

VoltrexKeyva:
- Update the `replace-in-file` dependency and the `c8` development dependency to their new major versions.
- Update all the other dependencies and development dependencies to their latest versions.


VoltrexKeyva:
Hmm there seems to be 2 high severity vulnerabilities reported by npm because of the latest version of `inquirer` depending on a vulnerable version of `lodash.set`, should we wait for that to be resolved?

targos (approved):

aduh95 (approved):

PR #831, opened 2024-07-09
lib: sort affected versions ASC

RafaelGSS:
Fixes: https://github.com/nodejs-private/security-release/issues/36

### DIFF
--- a/lib/security_blog.js
+++ b/lib/security_blog.js
@@ -331,7 +331,7 @@ export default class SecurityBlog {
         affectedVersions.add(affectedVersion);
       }
     }
-    return Array.from(affectedVersions).join(', ');
+    return Array.from(affectedVersions).sort().join(', ');
   }
 
   getSecurityPreReleaseTemplate() {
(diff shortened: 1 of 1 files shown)


lib/security_blog.js
```diff
@@ -331,7 +331,7 @@ export default class SecurityBlog {
         affectedVersions.add(affectedVersion);
       }
     }
-    return Array.from(affectedVersions).join(', ');
+    return Array.from(affectedVersions).sort().join(', ');
```
richardlau, line 334: Is this a numeric or string sort?

lib/security_blog.js
```diff
@@ -331,7 +331,12 @@ export default class SecurityBlog {
         affectedVersions.add(affectedVersion);
       }
     }
-    return Array.from(affectedVersions).join(', ');
+    const parseToNumber = str => str.match(/[\d.]+/g)[0];
```
richardlau, line 334: This doesn't actual parse to a number? (i.e. this is parsing to a string comprising of digits.)

RafaelGSS: Yeah, I forgot the `+`.

richardlau (approved):

PR #833, opened 2024-07-16
feat: add git node security --cleanup

RafaelGSS:
This adds a final step to cleanup

### DIFF
--- a/components/git/security.js
+++ b/components/git/security.js
@@ -44,8 +44,8 @@ const securityOptions = {
     describe: 'Create the post-release announcement',
     type: 'boolean'
   },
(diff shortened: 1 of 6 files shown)


components/git/security.js
```diff
@@ -43,6 +43,10 @@ const securityOptions = {
   'post-release': {
     describe: 'Create the post-release announcement',
     type: 'boolean'
+  },
+  finish: {
```
marco-ippolito, line 47:
```suggestion
  finalize
```

Maybe `finalize` or `cleanup`. Finish doesn't sound good 😄

PR #837, opened 2024-07-29
fix(ncu-ci): use correct Jenkins URL to start builds

targos:
I found the solution on https://stackoverflow.com/questions/55533294/getting-400-this-page-expects-a-form-submission-when-making-a-rest-call-to-tri

I don't know how it has worked before nor why it suddenly started to fail.


targos:
Fixes: https://github.com/nodejs/node/pull/54087#issuecomment-2254828903

targos:
It's not working. While the request returned code 201, no build was started.

targos:
The main CI host's disk was full. I ran a cleanup script and it's working again.

PR #838, opened 2024-08-06
fix: ignore codecov check suite

RedYetiDev:
Ref https://openjs-foundation.slack.com/archives/C019MGJQ8RH/p1722871958000979


targos (approved):

lib/pr_checker.js
```diff
@@ -373,8 +374,8 @@ export default class PRChecker {
 
     // GitHub new Check API
     for (const { status, conclusion, app } of checkSuites.nodes) {
-      if (app && app.slug === 'dependabot') {
-        // Ignore Dependabot check suites. They are expected to show up
+      if (app && IGNORED_CHECK_SLUGS.includes(app.slug)) {
+        // Ignore app check suites. They are expected to show up
```
aduh95, line 378:
```suggestion
        // Ignore GitHub Apps check suites. They are expected to show up
```
That makes me think; should we filter only status for checks from `{"slug":"github-actions"}`?

targos: Might be fine as long as we don't require checks from other apps :)

RedYetiDev: I think just ignoring dependabot and codecov is fine for now, but if the project installs more apps, maybe then ignore all apps.

aduh95 (approved):

PR #839, opened 2024-08-12
chore: ignore exuberant-ctags

RafaelGSS:
As titled

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ tmp-*
 .eslintcache
 .ncu
 package-lock.json
+tags


targos:
I don't understand the title.

RafaelGSS:
exuberant-ctags (https://ctags.sourceforge.net/) is an index generator (tags) for projects. It helps identify functions and classes when using this tool.

It's very common to use ctags with vim (my case) and a file `tags` is generated every time you open the project or update it. Adding it to gitignore would help me and others with the same developer workflow.

RafaelGSS:
> Isn't it the kind of paths that should be in a global gitignore (like `.DS_Store` and other developer-specific files)?

That works too. Anyway, I thought it wouldn't hurt to have it on gitignore too. That's fine.

PR #842, opened 2024-08-20
feat(git-node): auto-fetch latest release tag when preparing release

aduh95:
As suggested by https://github.com/nodejs/node/pull/53828#discussion_r1676554998

### DIFF
--- a/lib/prepare_release.js
+++ b/lib/prepare_release.js
@@ -4,7 +4,6 @@ import { promises as fs } from 'node:fs';
 import semver from 'semver';
 import { replaceInFile } from 'replace-in-file';
 
-import { getMergedConfig } from './config.js';
 import { runAsync, runSync } from './run.js';
 import { writeJson, readJson } from './file.js';
 import Request from './request.js';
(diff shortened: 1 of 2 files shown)


lib/prepare_release.js
```diff
@@ -369,24 +374,35 @@ export default class ReleasePreparation {
...
+    return {
+      major,
+      minor,
+      patch,
+      tag: `v${major}.${minor}.${patch}`
     };
+  }
 
-    const changelog = this.getChangelog();
+  calculateNewVersion() {
```
RafaelGSS, line 391: Any reason not to use `pkjs@nv` instead of match parsing the changelog? Also, note we are parsing the CHANGELOG **before** resetting/sync the branch with upstream/vX.-staging, which means it could be outdated and the "new version" already exists

aduh95: > note we are parsing the CHANGELOG **before** resetting/sync the branch with upstream/vX.-staging

It's the same time we were checking for the tag before, so I don't think it can produce different results (i.e. if you try to prepare a release from an outdated branch, you might be preparing a release that's already out, before or after this change).

> Any reason not to use `pkjs@nv` instead of match parsing the changelog?

What's that?

RafaelGSS: Wrong name, sorry. I meant https://github.com/pkgjs/nv.

aduh95: One problem would be that it only exposes an async function – which makes sense because it does a network call to <https://nodejs.org/dist/index.json> – and we can't have an async constructor. It is technically possible to make it sync with e.g. a [synchronous worker](https://www.npmjs.com/package/synchronous-worker), not sure it's worth it though

aduh95: I refactored it to a new async `prepareLocalBranch` method, and kept the changelog parsing, but after suggesting fetching the git branch from the upstream. PTAL

RafaelGSS (approved):

lib/prepare_release.js
RafaelGSS: I still believe the pkgjs@nv would simplify the code, but I can see your point of view, and that's fine for me. Approved :)

PR #846, opened 2024-08-27
feat(git-node): auto-fetch comparison branch when preparing release

aduh95:
A few things in the release preparation script were still relying on the local clone to be up-to-date (e.g. to get the Current line (to prepare a LTS release), it relied on the release tag to be present, or when comparing branches). Given preparing a release requires internet access anyway, it seems easy enough to rely on it rather than the local clone.

Testing it from a codespace like this:

```
git clone https://github.com/nodejs/node.git --single-branch --branch=v20.x-staging --no-tags
cd node
../bin/git-node.js release --prepare
```

### DIFF
--- a/lib/prepare_release.js
+++ b/lib/prepare_release.js
@@ -4,7 +4,7 @@ import { promises as fs } from 'node:fs';
 import semver from 'semver';
 import { replaceInFile } from 'replace-in-file';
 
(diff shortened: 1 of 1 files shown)


targos (approved):
I assume you tried it locally.

PR #847, opened 2024-08-28
fix: landing session on different repo/org

RafaelGSS:
This allows the `git node land` usage on different forks. It might also help on nodejs-private.

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -21,7 +21,7 @@ export default class LandingSession extends Session {
     prid, backport, lint, autorebase, fixupAll,
     checkCI, oneCommitMax, ...argv
   } = {}) {
-    super(cli, dir, prid);
+    super(cli, dir, prid, argv);
     this.req = req;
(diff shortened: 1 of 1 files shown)


aduh95:
It's not obvious (IMO) how this change is related to landing session on different repo/org, could you explain?

RafaelGSS:
> It's not obvious (IMO) how this change is related to landing session on different repo/org, could you explain?

When you use `git node land https://github.com/ANOTHER-ORG/ANOTHER-REPO` it just ignores the argv because we are not passing argv to `Session` as it's specified in https://github.com/nodejs/node-core-utils/blob/main/lib/session.js#L18

So, when we attempt to read PRData we use the default argv instead of the expected argv (owner/repo) https://github.com/nodejs/node-core-utils/blob/a8529edaab765e293965054b7027774a3f4938f2/components/git/land.js#L204.

aduh95 (approved):

PR #848, opened 2024-08-30
fix(actions): add appropriate permissions for release

RedYetiDev:
As it currently stands, this workflow is failing due to lack of permissions. This PR remedies that by given the workflow the permissions required to update releases.


aduh95:
Superseded by https://github.com/nodejs/node-core-utils/pull/852

PR #851, opened 2024-09-01
fix(v8): handle error in git apply

targos:
Capture stdout so it can be used in the catch block when there is a conflict.

Fixes: https://github.com/nodejs/node-core-utils/issues/850


targos:
I was able to debug and open https://github.com/nodejs/node/pull/54682 with this change.

aduh95 (approved):

richardlau (approved):

PR #852, opened 2024-09-01
chore: update release-please config

targos:
- Use new repository for the action
- Use nodejs-github-bot's PAT
- Set minimal permissions for the default token

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   release-please:
     runs-on: ubuntu-latest
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -11,10 +11,11 @@ jobs:
     outputs:
       release_created:  ${{ steps.release.outputs.release_created }}
     steps:
-      - uses: google-github-actions/release-please-action@v4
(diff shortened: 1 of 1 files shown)


aduh95 (approved):
We could also set explicitly `permissions: contents: read` since that's the only needed permission required from now on IIUC

targos:
According to the docs, it needs write permission: https://github.com/googleapis/release-please-action#workflow-permissions

aduh95:
> According to the docs, it needs write permission: https://github.com/googleapis/release-please-action#workflow-permissions

Setting `permissions: contents: read` only affects permission of the `GITHUB_TOKEN`, which is only used by `actions/checkout` after this PR. The required permissions for `googleapis/release-please-action` are irrelevant, because it uses a different token IIUC.

targos:
I see. Updated, thanks!

aduh95:
`RequestError [HttpError]: Error creating Pull Request: Resource not accessible by personal access token` 🤔

targos:
I might have given the wrong permission. Will check when I'm on my computer unless someone else does it first

targos:
https://github.com/settings/personal-access-tokens/3911092

CleanShot 2024-09-01 at 19 33 40@2x

targos:
I forgot to give write access to the bot in https://github.com/nodejs/node-core-utils/settings/access. Now it's done.

PR #855, opened 2024-09-18
fix(git-node): followup fix for supported wpt jsons

RedYetiDev:
CC @panva 

Followup https://github.com/nodejs/node/pull/54593/files#diff-044c4e3128d6bcf139f31a36cf054f2e59954367d4493e1c1b1a42d7b9d88067

### DIFF
--- a/components/git/wpt.js
+++ b/components/git/wpt.js
@@ -52,7 +52,7 @@ async function main(argv) {
   if (fs.existsSync(statusFolder)) {
     const jsons = fs.readdirSync(statusFolder);
     supported = supported.concat(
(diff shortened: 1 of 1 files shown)


panva (approved):

RedYetiDev:
@nodejs/node-core-utils this hasn't had any objections, and without it, updating the WPT is harder (for interfaces)

aduh95:
It's not immediately obvious why this change is needed now, did the WPT suite start using non-JSON files recently?

RedYetiDev:
Yes. In a recent PR, the WebCryptoAPI status file was changed from a JSON file to a JS file.

PR #856, opened 2024-09-19
chore(wpt): warn for non-existent wpt path

jazelly:
In case of `git node wpt anyUnknownPath`, currently we get a `TypeError: cannot read from undefined`, which is not intuitive.

### DIFF
--- a/lib/github/tree.js
+++ b/lib/github/tree.js
@@ -35,7 +35,16 @@ export default class GitHubTree {
       branch,
       path
     });
-    return data.repository.ref.target.history.nodes[0].oid;
+
+    const targetHistoryNodes = data.repository.ref.target.history.nodes;
+    if (!targetHistoryNodes.length) {
+      this.cli.stopSpinner(
+        `Cannot find commit for "${path}". Please check the path name.`,
+        this.cli.SPINNER_STATUS.FAILED
+      );
+      throw new Error(`Cannot find commit for "${path}"`);
+    }
+    return targetHistoryNodes[0].oid;
   }
 
   /**
--- a/test/unit/wpt_updater.test.js
+++ b/test/unit/wpt_updater.test.js
@@ -1,0 +1,64 @@
+import { describe, it, before, after } from 'node:test';
+import assert from 'node:assert';
+import TestCLI from '../fixtures/test_cli.js';
+import sinon from 'sinon';
+
+import { WPTUpdater } from '../../lib/wpt/index.js';
+
+describe('WPTUpdater', function() {
+  const UNKNOWN_PATH = 'unknown path';
+  let request;
+  let wptUpdater;
+  let nodedir;
+  let cli;
+  let path;
+  const emptyData = {
+    repository:
+    {
+      ref: { target: { history: { nodes: [] } } }
+    }
+  };
+
+  before(() => {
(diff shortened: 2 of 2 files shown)


lib/github/tree.js
```diff
@@ -35,7 +35,15 @@ export default class GitHubTree {
...
       path
     });
-    return data.repository.ref.target.history.nodes[0].oid;
+
+    const targetHistoryNodes = data.repository.ref.target.history.nodes;
+    if (!targetHistoryNodes.length) {
+      // Sometimes this function is in a spinner, and we need to print the error
+      // after a new line.
+      console.log();
+      throw new Error(`Cannot find WPT for "${path}". Please check the WPT name.`);
```
joyeecheung, line 44: Does `cli.stopSpinner(message, cli.SPINNER_STATUS.FAILED)` work here?

jazelly: Ah I see. It works much better.

lib/github/tree.js
```diff
@@ -29,13 +29,29 @@ export default class GitHubTree {
 
   async _getLastCommit() {
     const { request, owner, repo, branch, path } = this;
+    console.log('asdasd', COMMIT_QUERY, {
+      owner,
+      repo,
+      branch,
+      path
+    });
```
aduh95, line 37:
```suggestion
```

lib/github/tree.js
```diff
@@ -29,13 +29,29 @@ export default class GitHubTree {
...
     const data = await request.gql(COMMIT_QUERY, {
       owner,
       repo,
       branch,
       path
     });
-    return data.repository.ref.target.history.nodes[0].oid;
+
+    console.log('data', data);
```
aduh95, line 45:
```suggestion
```

lib/github/tree.js
```diff
@@ -29,13 +29,29 @@ export default class GitHubTree {
...
+    const targetHistoryNodes = data.repository.ref.target.history.nodes;
+    if (!targetHistoryNodes.length) {
+      this.cli.stopSpinner(
+        `Cannot find WPT for "${path}". Please check the WPT name.`,
+        this.cli.SPINNER_STATUS.FAILED
+      );
+      process.exit();
```
aduh95, line 52: We need to set a non-zero exit code so the terminal (or whatever has spawn the process) can report the error
```suggestion
      process.exit(1);
```
But IMO it would be more user-friendly to throw an actual error, with a stack trace, so anyone trying to debug the code can follow what's going wrong.
```suggestion
      throw new Error('Cannot find last commit');
```

lib/github/tree.js
jazelly: oops

lib/github/tree.js
jazelly: SGTM

jakecastelli: Not sure throw an error is better than `process.exit(1)` 😄 As `this.cli.stopSpinner` already has a clear `message` to indicate the cause and easy to locate in the code base and `this.cli.SPINNER_STATUS.FAILED` to indicate something went wrong + error exit code.

aduh95 (approved):

joyeecheung (approved):

PR #857, opened 2024-09-24
fix(git-node): ignore all non-gha nodes when checking for GitHub CI

aduh95:
Implements https://github.com/nodejs/node-core-utils/pull/838#discussion_r1706574660. The problem has come back now that we have a new `jenkins-node-js-ci` (see https://github.com/nodejs/admin/issues/916), and IMO it only makes sense to check for GHA-related jobs only when checking the state of the GitHub CI.

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -374,7 +374,7 @@ export default class PRChecker {
     // GitHub new Check API
     for (const { status, conclusion, app } of checkSuites.nodes) {
       if (app.slug !== 'github-actions') {
-        // Ignore Dependabot and Codecov check suites.
+        // Ignore all non-github check suites, such as Dependabot and Codecov.
         // They are expected to show up on PRs whose head branch is not on a
         // fork and never complete.
         continue;
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -29,7 +29,6 @@ const GITHUB_SUCCESS_CONCLUSIONS = ['SUCCESS', 'NEUTRAL', 'SKIPPED'];
 const FAST_TRACK_RE = /^Fast-track has been requested by @(.+?)\. Please 👍 to approve\.$/;
 const FAST_TRACK_MIN_APPROVALS = 2;
 const GIT_CONFIG_GUIDE_URL = 'https://github.com/nodejs/node/blob/99b1ada/doc/guides/contributing/pull-requests.md#step-1-fork';
-const IGNORED_CHECK_SLUGS = ['dependabot', 'codecov'];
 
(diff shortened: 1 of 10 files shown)


lib/pr_checker.js
```diff
@@ -374,9 +373,10 @@ export default class PRChecker {
 
     // GitHub new Check API
     for (const { status, conclusion, app } of checkSuites.nodes) {
-      if (app && IGNORED_CHECK_SLUGS.includes(app.slug)) {
+      if (app.slug !== 'github-actions') {
```
VoltrexKeyva, line 376:
```suggestion
      if (app && app.slug !== 'github-actions') {
```

RedYetiDev:
```suggestion
      if (app?.slug !== 'github-actions') {
```

a suggestion upon your suggestion

lib/pr_checker.js
```diff
@@ -374,9 +373,10 @@ export default class PRChecker {
...
         // Ignore Dependabot and Codecov check suites.
```
RedYetiDev, line 377:
```suggestion
        // Ignore all non-github checks, such as Dependabot and Codecov.
```

lib/pr_checker.js
aduh95: I'm not sure this is needed.

richardlau (approved):

targos:
>  The problem has come back now that we have a new `jenkins-node-js-ci`

Now I don't understand:
- How this can affect "production" jobs. I have only tried to use the plugin on a [copy of the linter job](https://ci.nodejs.org/view/All/job/node-test-linter-targos/)
- How can I see these checks? I have still been unable to see any effect of the option on the job copy (in GitHub UI or Jenkins output).

aduh95:
> * How can I see these checks?

One way is to apply the following diff and run `git node land` on a PR which head branch in on nodejs/node.

```diff
diff --git a/lib/pr_checker.js b/lib/pr_checker.js
index 064c647..129b9a5 100644
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -381,6 +381,7 @@ export default class PRChecker {
       }
 
       if (status !== 'COMPLETED') {
+        console.log(checkSuites.nodes);
         cli.error('GitHub CI is still running');
         return false;
       }
@@ -395,6 +396,7 @@ export default class PRChecker {
     if (commit.status) {
       const { state } = commit.status;
       if (state === 'PENDING') {
+        console.log(checkSuites.nodes);
         cli.error('GitHub CI is still running');
         return false;
       }
```

targos (approved):

PR #860, opened 2024-09-27
chore: update dependencies

VoltrexKeyva:
- Replace the `inquirer` dependency with `@inquirer/prompts` as the former is now legacy.
- Update the `log-symbols` dependency and the `sinon` development dependency to their new major
versions.
- Update all other dependencies & development dependencies to their latest versions.


VoltrexKeyva:
Please note that I have **not** tested the inquirer changes because to see the prompts I would need to use commands that may have real effects on the main repository, it'd be appreciated if someone can test them.

Side note: I still have not updated the ESLint and related development dependencies because `eslint-plugin-import` still does not support ESLint v9.

targos (approved):
I tested the inquirer changes with a couple commands.

VoltrexKeyva:
> I tested the inquirer changes with a couple commands.

Thanks a lot, any idea about the error in the tests I mentioned?

targos:
I haven't looked at it

aduh95:
> any idea about the error in the tests I mentioned?

CI is green though 🤔 Is this an error you see locally, or was it fixed since your comment?

VoltrexKeyva:
> > any idea about the error in the tests I mentioned?
> 
> CI is green though 🤔 Is this an error you see locally, or was it fixed since your comment?

I don't know how the CI is green despite that error, if you check the logs of the "Run tests" step of every check in the CI run, the same error is present, [here](https://github.com/nodejs/node-core-utils/actions/runs/11074428387/job/30773193587#step:14:1) is an example.

VoltrexKeyva:
@aduh95 ping.

PR #862, opened 2024-11-01
feat: add --yes option to git node release

RafaelGSS:
This will enable us to automate release proposal
creation: https://github.com/nodejs/security-wg/issues/860

---

I've fixed a few lint nits in the git-node.md too.

### DIFF
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -32,6 +32,12 @@ const releaseOptions = {
   startLTS: {
     describe: 'Mark the release as the transition from Current to LTS',
     type: 'boolean'
+  },
+  yes: {
+    type: 'boolean',
+    default: false,
+    describe: 'Assume "yes" as answer to all prompts and run ' +
+    'non-interactively.'
   }
 };
 
@@ -65,6 +71,10 @@ function release(state, argv) {
   const logStream = process.stdout.isTTY ? process.stdout : process.stderr;
   const cli = new CLI(logStream);
   const dir = process.cwd();
+
+  if (argv.yes) {
+    cli.setAssumeYes();
+  }
 
   return runPromise(main(state, argv, cli, dir)).catch((err) => {
     if (cli.spinner.enabled) {
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -40,8 +40,7 @@ const releaseOptions = {
   yes: {
     type: 'boolean',
     default: false,
-    describe: 'Assume "yes" as answer to all prompts and run ' +
-    'non-interactively.',
+    describe: 'Skip all prompts and run non-interactively'
   }
 };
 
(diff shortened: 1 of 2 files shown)


docs/git-node.md
```diff
@@ -224,6 +224,8 @@ Options:
   --security  Demarcate the new security release as a security release [boolean]
   --startLTS  Mark the release as the transition from Current to LTS   [boolean]
   --filterLabel Filter PR by label when preparing a security release   [string]
+  --yes       Assume "yes" as answer to all prompts and run
+              non-interactively. [boolean] [default: false]
```
aduh95, line 228: I'm probably stating the obvious but it will assume "No" when the default answer is a "No". Not a big deal but the wording is somewhat confusing as is

RafaelGSS: hmm, aI thought it would assume `Yes` for all cases. I might need to change all defaults to `yes`.

aduh95: It might still be worth removing the potentially wrong part
```suggestion
  --yes       Skip all prompts and run
              non-interactively. [boolean] [default: false]
```

docs/git-node.md
```diff
@@ -336,7 +338,7 @@ ncu-config set waitTimeMultiApproval 48
 
 ## `git node v8`
 
-Update or patch the V8 engine.  
+Update or patch the V8 engine.
```
aduh95, line 341: I'm not sure we want to include those changes (the two spaces are there to force a line break), in any case it's unrelated to the rest of the PR

docs/git-node.md
```diff
@@ -373,7 +375,7 @@ Options:
 ### `git node v8 minor`
 
 Compare current V8 version with latest upstream of the same major. Applies a
-patch if necessary.  
+patch if necessary.
```
aduh95, line 378: ditto
```suggestion
patch if necessary.  
```

docs/git-node.md
```diff
@@ -223,6 +223,8 @@ Options:
   --prepare   Prepare a new release of Node.js                         [boolean]
   --security  Demarcate the new security release as a security release [boolean]
   --startLTS  Mark the release as the transition from Current to LTS   [boolean]
+  --yes       Skip all prompts and run
+              non-interactively. [boolean] [default: false]
```
aduh95, line 227:
```suggestion
  --startLTS  Mark the release as the transition from Current to LTS   [boolean]
  --yes       Skip all prompts and run non-interactively               [boolean]
```

components/git/release.js
```diff
@@ -33,9 +37,11 @@ const releaseOptions = {
...
     type: 'boolean'
   },
-  releaseDate: {
-    describe: 'Default relase date when --prepare is used. It must be YYYY-MM-DD',
-    type: 'string'
+  yes: {
+    type: 'boolean',
+    default: false,
+    describe: 'Assume "yes" as answer to all prompts and run ' +
+    'non-interactively.',
```
aduh95, line 44:
```suggestion
    describe: 'Skip all prompts and run non-interactively'
```

RafaelGSS:
PTAL @aduh95

aduh95 (approved):

PR #865, opened 2024-11-02
chore: fix `ci_start` tests

aduh95:
Refs: https://github.com/nodejs/node-core-utils/pull/860#issuecomment-2406024971

It looks like without this change, async assertions errors are not always picked up by the test runner, maybe it's being swallowed by sinon.


VoltrexKeyva:
The error still seems to be swallowed since it's in the logs, I'm not sure why.

PR #869, opened 2024-11-13
feat(git-node): prompt before attempting branch-diff

aduh95:
I hit this when preparing v18.20.5 during the collab summit, it seems like a good idea to use an interaction to help the releaser.


targos (approved):
I always run the prepare command with `--skipBranchDiff`. I personally prefer to prepare and test the staging branch separately from the release proposal.

PR #870, opened 2024-11-13
feat(git-node): use a single `git push` command

aduh95:
I cannot see why we would prefer having several `git push` commands. However, I see a number of reasons to have only one, at the very end of the `--promote` sequence:

- it makes it less likely a release would go half through, e.g. if the releaser has to abandon the release before the `git push` for whatever reason.
- for folks watching the repo, the delay between "the proposal branch gets merged" and "the release tag is out" can be frustrating.
- it removes the need for a `--continue` (cc @RafaelGSS); with this PR, you either have pushed and the only remaining command is `tools/release.sh`, or you haven't pushed and can start over.

### DIFF
--- a/lib/promote_release.js
+++ b/lib/promote_release.js
@@ -134,11 +134,8 @@ export default class ReleasePromotion extends Session {
 
     // Set up for next release.
     cli.startSpinner('Setting up for next release');
-    await this.setupForNextRelease();
+    const workingOnNewReleaseCommit = await this.setupForNextRelease();
     cli.stopSpinner('Successfully set up for next release');
-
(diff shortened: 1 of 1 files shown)


RafaelGSS (approved):
LGTM

aduh95:
It's probably wiser to wait for a release proposal PR to be open so this can be tested, I reckon I might be missing a `.trim()` call somewhere

RafaelGSS:
Right, so I'm going to land #868 without this change

aduh95:
Just tested it with https://github.com/nodejs/node/pull/55879, a `.trim()` call was indeed needed

PR #871, opened 2024-11-16
fix(git-node): do not assume release commit will conflict

aduh95:
As I was testing `git node release --promote` with https://github.com/nodejs/node/pull/55879, I realized the assumption that a release commit would necessarily conflict when cherry-picked to the default branch is wrong for the case of semver-patch releases.

### DIFF
--- a/lib/promote_release.js
+++ b/lib/promote_release.js
@@ -147,23 +147,31 @@ export default class ReleasePromotion extends Session {
       cli.warn(`Aborting release promotion for version ${version}`);
       throw new Error('Aborted');
     }
-    await this.cherryPickToDefaultBranch();
+    const appliedCleanly = await this.cherryPickToDefaultBranch();
 
-    // Update `node_version.h`
-    await forceRunAsync('git', ['checkout', 'HEAD', '--', 'src/node_version.h'],
+    // Ensure `node_version.h`'s `NODE_VERSION_IS_RELEASE` bit is not updated
+    await forceRunAsync('git',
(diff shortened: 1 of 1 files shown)


lib/promote_release.js
```diff
@@ -147,23 +147,31 @@ export default class ReleasePromotion extends Session {
       cli.warn(`Aborting release promotion for version ${version}`);
       throw new Error('Aborted');
     }
-    await this.cherryPickToDefaultBranch();
+    const appliedCleanly = await this.cherryPickToDefaultBranch();
```
RafaelGSS, line 150: I think it's better to check if the release is `semver-patch` instead of `appliedCleanly`. It will be confusing to people when reading the code.

aduh95: I don't think it's a good idea to assume that a semver-patch release never conflicts, and that a non-semver-patch would always conflict, that would be setting ourselves to fail. I don't understand why it would be confusing

RafaelGSS: I feel confusing the `git checkout HEAD^` | `git checkout HEAD` part considering a `appliedCleanly` variable. Maybe a comment? I'm fine leaving it as `appliedCleanly` with a comment on that part

targos: We could use `git rev-parse HEAD` before trying to cherry-pick, and then `git checkout` (or the less ambiguous `git restore`) with the exact commit sha.

aduh95: I decided to add comments

RafaelGSS (approved):

PR #872, opened 2024-11-17
feat(git-v8): preserve original author when backporting

aduh95:
It feels wrong to me to take credit when backporting a V8 commit. I see a reason for not cherry-picking the actual commit instead of creating a new one, so this PR is doing that (with a flag to roll back to the previous behavior).

I'm also adding the `-S` flag for those who want to selectively sign commits.

### DIFF
--- a/components/git/v8.js
+++ b/components/git/v8.js
@@ -44,10 +44,22 @@ export function builder(yargs) {
             describe: 'Bump V8 embedder version number or patch version',
             default: true
           })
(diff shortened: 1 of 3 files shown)


targos:
SGTM for clean cherry-picks (I was reluctant to do it when we had the AUTHORS file). What about backports with conflicts? I feel like we should add a `Co-authored-by` trailer.

PR #881, opened 2024-12-23
feat(git-node): add support for auto-fill Notable Changes section

aduh95:
Allows PR author (or repo maintainers) to define themselves the Notable Change using a `<details>` section in the PR description.

We should also update the workflow that comments when the label is added to add the instructions.

Fixes: https://github.com/nodejs/Release/issues/1068

### DIFF
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -116,8 +116,10 @@ async function main(state, argv, cli, dir) {
   if (prID) {
     argv.prid = Number(prID[1]);
   }
+  const credentials = await auth({ github: true });
+  const request = new Request(credentials);
   if (state === PREPARE) {
-    const release = new ReleasePreparation(argv, cli, dir);
+    const release = new ReleasePreparation(argv, request, cli, dir);
 
     await release.prepareLocalBranch();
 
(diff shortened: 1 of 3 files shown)


marco-ippolito (approved):

aduh95:
What I imagine as a next step is to update the `NOTABLE_CHANGE_MESSAGE` in the nodejs/node repo to add instructions and an example/sample to guide contributors, and see how far it takes us. My expectation is it should work "well enough" to begin with, and that would tell us if we need to do more.

joyeecheung:
That would still meant a hassle for PR authors once it's labeled notable-change, no? If we make the section generic enough to be part of the PR template for all PRs, then the author generally don't have to do anything more after it's labeled notable-change. They can just put the summary in the templated section as soon as the PR is opened and forget about it. That might eliminate the need of the bot all together, too, though it doesn't hurt to keep it to emphasize it again.

joyeecheung:
That would still mean a hassle for PR authors once it's labeled notable-change, no? If we make the section generic enough to be part of the PR template for all PRs, then the author generally don't have to do anything more after it's labeled notable-change. They can just put the summary in the templated section as soon as the PR is opened and forget about it. That might eliminate the need of the bot all together, too, though it doesn't hurt to keep it to emphasize it again.

joyeecheung:
That would still mean a hassle for PR authors once it's labeled notable-change, no? If we make the section generic enough to be part of the PR template for all PRs, then the author generally don't have to do anything more after it's labeled notable-change. They can just put the summary in the templated section as soon as the PR is opened and forget about it. That might eliminate the need of the bot all together, too, though it doesn't hurt to keep it to emphasize it again.

PR #875, opened 2024-11-27
Add `git node staging` command

ruyadorno:
Add a new `git node staging` command that automates cherry-picking commits into staging branches.

It works by cherry-picking all commits that have no conflicts and stopping when finding commits that have conflicts so that the releaser can either manually fix that conflict to continue or skip.

Usage:

Fetches a commit list using `branch-diff` and automatically cherry-picks / skips commits based on whether or not they land cleanly:

```
  git node staging
```

By default the script will add `backport-request` labels to any skip commit, in order to run in a "dry-run" mode, avoiding the GH cli interactions, you can use the `skipGH` flag, e.g:

```
  git node staging --skipGH
```

After fixing up manual conflicts using `git cherry-pick --continue` the releaser can resume the cherry-pick queue:

```
  git node staging --continue
```

Similarly, after manually skipping a commit using `git cherry-pick --skip` the releaser can resume the cherry-pick queue using:

```
  git node staging --skip
```

If you want to run the script and automatically skip any commit that is unable to land cleanly you can use the `--autoSkip` flag:

```
  git node staging --autoSkip
```

Sets a custom reporter at the end of the automated operation:

```
  git node staging --reporter=json
```

Limits to 10 the number of commits to be cherry-picked:

```
  git node staging --pagination=10
```

Automates the backport request message, this won't run any of the `branch-diff` or cherry-pick routines. Useful for when you removed a faulty commit from the branch and want to signal to PR author and collaborators that commit now needs backporting, just use its PR#:

```
  git node staging --backport=12345
```

### More:

The automate cherry-pick logic also includes local persistency of the ongoing commit list, in case a fatal error happens during the command execution, it's possible to resume after cleaning up the git repo state by running `git node staging` again.

---

## Report

A report is generated at the end of operations listing information on the commits that were successfully cherry picked and commits that failed to be cherry-picked due to conflicts.

> [!IMPORTANT]
> Below is a short example of how these reports look like. The following report was generated for an operation that was limited using `--pagination=10`

# Cherry-pick Report
## 6 successfully cherry-picked commits:
* [`a21be0294d0`](https://github.com/nodejs/node/commit/a21be0294d0027773d0f60f81ffa0a34ba38e680) [build,win: enable pch for clang-cl](https://github.com/nodejs/node/pull/55249) `windows`, `build`, `v8 engine`, `tools`, `needs-ci`, `gyp`
[...]

PR #882, opened 2025-01-06
Fix `git node release --prepare` update proposal workflow

ruyadorno:
This allows for updating a proposal once the proposal branch has already been created, PR opened, e.g:

```sh
git checkout v1.x-staging # go back to the staging branch
# make changes, rebase, add more commits, (likely force) push v1.x-staging to upstream
git node release -S --prepare 1.y.z # run release prepare script again
# proposal branch is now ready to be force-pushed to upstream, release commit and changelog are updated
```

### DIFF
--- a/lib/prepare_release.js
+++ b/lib/prepare_release.js
@@ -93,9 +93,9 @@ export default class ReleasePreparation extends Session {
     } = this;
 
     // Create new proposal branch.
-    cli.startSpinner(`Creating new proposal branch for ${newVersion}`);
+    cli.startSpinner(`Switching to proposal branch for ${newVersion}`);
     const proposalBranch = await this.createProposalBranch(releaseBranch);
-    cli.stopSpinner(`Created new proposal branch for ${newVersion}`);
+    cli.stopSpinner(`Switched to proposal branch for ${newVersion}`);
 
     const success = await this.cherryPickSecurityPRs(filterLabels);
     if (!success) {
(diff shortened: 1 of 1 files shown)


ruyadorno:
Note: there are two foundational commits included that are borrowed from https://github.com/nodejs/node-core-utils/pull/875 adding the ability to capture `stderr` output and increasing unit test coverage for `lib/run.js`, it'd be better to rebase and merge this PR to keep them as separate commits but anyways if this lands before https://github.com/nodejs/node-core-utils/pull/875 these commits can then be removed from that PR.

lib/prepare_release.js
```diff
@@ -533,13 +535,37 @@ export default class ReleasePreparation extends Session {
...
+          base
+        ]);
+        return {
+          proposalBranch,
+          resultMsg: `Updated proposal branch for ${newVersion}`
+        };
+      } else {
+        throw err;
+      }
+    }
```
aduh95, line 564: Looking at https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt-emgitcheckoutem-b-Bltnew-branchgtltstart-pointgt, it looks like we can simplify all this
```suggestion
    await forceRunAsync('git', [
        'checkout',
        '-B',
        proposalBranch,
        base
      ], { ignoreFailure: false });
    }
```

ruyadorno: Given this part I was not sure that it would work:

> the branch is not reset/created unless "git checkout" is successful (e.g., when the branch is in use in another worktree, not just the current branch stays the same, but the branch is not reset to the start-point, either).

but I just tried replacing with your suggestion here and it works as a charm! so scrap the initial implementation here, I'll go with this 😊

RafaelGSS (approved):

aduh95 (approved):

PR #883, opened 2025-01-08
chore(git-node): Add make lint-md reminder

ruyadorno:
After two consecutive releases in which I mistakenly pushed a broken release commit to main, I'm convinced that the process needs improvement.

This PR proposes to add a reminder to `git node release promote` to lint markdown files prior to pushing branches to upstream remote.

Refs: https://github.com/nodejs/node/commit/a344300bfa146fce9d1f20a71b073263a4720eab
Refs: https://github.com/nodejs/node/pull/56508

### DIFF
--- a/lib/promote_release.js
+++ b/lib/promote_release.js
@@ -195,6 +195,12 @@ export default class ReleasePromotion extends Session {
       }
     }
 
+    cli.separator();
+    cli.warn('Reminder: Lint markdown files');
+    cli.info('You might need to amend the commit if there are issues:');
+    cli.info('make lint-md');
+    cli.separator();
+
     // Push to the remote the release tag, and default, release, and staging branch.
     await this.pushToRemote(workingOnNewReleaseCommit);
 


richardlau (approved):

RafaelGSS (approved):

lib/promote_release.js
```diff
@@ -195,6 +195,12 @@ export default class ReleasePromotion extends Session {
       }
     }
 
+    cli.separator();
+    cli.warn('Reminder: Lint markdown files');
+    cli.info('You might need to amend the commit if there are issues:');
+    cli.info('make lint-md');
+    cli.separator();
```
aduh95, line 202: Maybe it makes more sense to move that to the branch where we know there were conflicts (line 168), for the cases where there are no conflicts, it seems safe to skip it

ruyadorno: That was my original assessment but in the specific occurrence linked in the PR body the bad formatting was introduced by a git automatic resolution, meaning that in theory a successfully cherry-picked commit could potentially still introduce markdown issues.

aduh95: I don't believe it, if I try locally:

```sh
git reset a344300bfa146fce9d1f20a71b073263a4720eab^ --hard # rewind just before the faulty commit
git cherry-pick v22.13.0^0 # cherry-pick the release commit
```

All 7 of the YAML sections fixed by https://github.com/nodejs/node/commit/5119049ca606fa2d0f6c64187ceab3150dcbe5c1 show up as conflict that must be manually resolved. There might be something else that "does it for you" (like an IDE extension or whatnot), but I'm quite confident git would never

anonrig (approved):

Issue #893, opened 2025-01-16
Change ESLint base config?


targos:
Standard doesn't support flat config / ESLint 9 and it seems abandoned: https://github.com/standard/eslint-config-standard

VoltrexKeyva:
https://github.com/standard/standard/issues/1948#issuecomment-2138078249, we could probably switch to [neostandard](https://github.com/neostandard/neostandard) if we want to continue using *standard* style.

targos:
Suggestion LGTM. I personally don't care about which style we use.

PR #902, opened 2025-01-26
chore: migrate to ESLint v9 and update deps

VoltrexKeyva:
- Migrate to ESLint flat config
- Switch from `eslint-config-standard` to `neostandard` as the former does not support ESLint v9 and seems to be unmaintained. See https://github.com/standard/standard/issues/1948#issuecomment-2138078249.
- Conform to the rules of the ESLint plugins/configs.
- Update all dependencies to their latest versions.

Closes https://github.com/nodejs/node-core-utils/issues/893.


VoltrexKeyva:
Not sure what to do about this coverage miss 🤔.

bin/git-node.js
```diff
@@ -27,4 +27,6 @@ Promise.all(commandFiles.map(importCommand)).then((commands) => {
     .epilogue(epilogue)
     .help('help')
     .parse();
+}).catch((err) => {
+  throw err;
```
targos, line 31: What's the rule for this change?

components/git/metadata.js
```diff
@@ -90,5 +90,6 @@ export function handler(argv) {
       if (status === false) {
         throw new Error(IGNORE);
       }
+      return undefined;
```
targos, line 93: Ditto

eslint.config.js
```diff
@@ -0,0 +1,44 @@
...
+import importPlugin from 'eslint-plugin-import';
+
+export default [
+  pluginJs.configs.recommended,
+  ...neostandard(),
+  nodePlugin.configs['flat/recommended'],
+  pluginPromise.configs['flat/recommended'],
+  importPlugin.flatConfigs.recommended,
+  {
+    ignores: [
```
targos, line 15: I've learned that global ignores have to be alone in the first config element

bin/git-node.js
VoltrexKeyva: https://github.com/eslint-community/eslint-plugin-promise/blob/HEAD/docs/rules/catch-or-return.md

components/git/metadata.js
VoltrexKeyva: https://eslint.org/docs/latest/rules/getter-return

VoltrexKeyva: https://github.com/eslint-community/eslint-plugin-promise/blob/HEAD/docs/rules/always-return.md

eslint.config.js
VoltrexKeyva: I'm not sure I understand, mind to elaborate?

targos: https://eslint.org/docs/latest/use/configure/ignore#ignoring-files

Doesn't have to be the first element in the config array but must have no other keys.

targos:
We don't enforce coverage

RafaelGSS (approved):

targos (approved):

PR #907, opened 2025-02-20
fix(git-node): check the tag signature and not the commit one

aduh95:
During release promotion, we want to verify the tag is signed with one of the releasers' key. The commit can be signed with a different signature (e.g. the bot's, or a soon-to-be releaser).

### DIFF
--- a/lib/promote_release.js
+++ b/lib/promote_release.js
@@ -226,20 +226,28 @@ export default class ReleasePromotion extends Session {
 
   async verifyTagSignature() {
     const { cli, version } = this;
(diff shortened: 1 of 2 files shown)


targos:
The commit message makes me suspicious. The commit signature should have been good in my case as I had to rebase the PR locally.

aduh95:
Yeah that's because we're also checking the commit author, instead of the committer. Anyway, the current way is wrong on many levels, this PR is fixing that, happy to edit the commit message to something else

targos (approved):

PR #917, opened 2025-04-05
chore: remove duplicated code

aduh95:
`lib/cherry_pick.js` contains lots of lines of code copied from other part of the code base, making it much harder to maintain. Instead, let's take advantage of the features of JS that let us inherit from the specific methods we need.

### DIFF
--- a/lib/cherry_pick.js
+++ b/lib/cherry_pick.js
@@ -1,23 +1,13 @@
-import os from 'node:os';
 import path from 'node:path';
 import { getMetadata } from '../components/metadata.js';
 
 import {
-  runAsync, runSync, forceRunAsync
+  runAsync, runSync
 } from './run.js';
-import { writeFile } from './file.js';
-import {
-  shortSha, getEditor
-} from './utils.js';
 import { getNcuDir } from './config.js';
+import LandingSession, { LINT_RESULTS } from './landing_session.js';
 
-const LINT_RESULTS = {
-  SKIPPED: 'skipped',
-  FAILED: 'failed',
-  SUCCESS: 'success'
-};
-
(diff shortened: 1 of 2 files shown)


lib/cherry_pick.js
```diff
@@ -120,68 +100,6 @@ export default class CheckPick {
     }
   }
 
-  async downloadAndPatch(expectedCommitShas) {
-    const { cli, repo, owner, prid } = this;
-
-    cli.startSpinner(`Downloading patch for ${prid}`);
-    // fetch via ssh to handle private repo
-    await runAsync('git', [
-      'fetch', `git@github.com:${owner}/${repo}.git`,
```
targos, line 129: In `LandingSession`, there is `https://...` instead. The comment here seems to indicate the SSH fetch was intentional.

lib/cherry_pick.js
```diff
@@ -120,68 +100,6 @@ export default class CheckPick {
...
-    }
-
-    const commitInfo = { base, head, shas: commitShas };
-
-    try {
-      await forceRunAsync('git', ['cherry-pick', `${base}..${head}`], {
-        ignoreFailure: false
-      });
-    } catch (ex) {
-      await forceRunAsync('git', ['cherry-pick', '--abort']);
```
targos, line 159: `LandingSession` doesn't abort the cherry-pick on failure.

aduh95: Yeah, that's one of the inconsistency I'm trying to get rid of, the `--abort` call was removed in 1e6f5d3cbc7e837f1590122458b15f24ff9b378d, but not in this file.

RafaelGSS (approved):

PR #920, opened 2025-04-05
feat(ncu-ci): parse comments to find a safe commit

aduh95:
Currently, the `request-ci` only works for approved PRs. This is annoying because:

- collaborators cannot approve their own PRs, and have to wait for someone else to approve their PR for them, or start the CI manually.
- it gives an incentive to approve a PR just to run the CI rather than because the collaborator actually supports it.

There have been suggestions to use PR author, or PR head repo owner, or commit signature, or a combination of the three – but I don't really like that idea, because:

- it doesn't help for PRs from non-collaborators.
- we should care about who is requesting the CI, not who's proposing the change.

IIUC comments are not editable by triagers, so should be a safe way to determine a safe SHA for unapproved PRs. 

EDIT 2026: it is annoying, but also seems to be mostly working, folks have gotten used to wait for approval or start CI manually. Not sure this is actually needed.

### DIFF
--- a/lib/ci/run_ci.js
+++ b/lib/ci/run_ci.js
@@ -16,6 +16,26 @@ export const CI_PR_URL = `https://${CI_DOMAIN}/job/${CI_PR_NAME}/build`;
 const CI_V8_NAME = CI_TYPES.get(CI_TYPES_KEYS.V8).jobName;
 export const CI_V8_URL = `https://${CI_DOMAIN}/job/${CI_V8_NAME}/build`;
 
+const REQUEST_CI_COMMENT = /^@nodejs-github-bot .+([0-9a-f]{40})$/;
+
+async function findSafeFullCommitSHA(cli, prData, request) {
+  // First, let's check if the head was approved.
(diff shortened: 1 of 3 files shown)


bjohansebas:
> IIUC comments are not editable by triagers, so should be a safe way to determine a safe SHA for unapproved PRs.

yes, only people with write access have the ability to edit comments

targos:
Would you like a review before updating the tests?

targos (approved):
AFAICT, this allows to request CI for a commit SHA that is not part of the pull request. Is that a bug or a feature?

aduh95:
> AFAICT, this allows to request CI for a commit SHA that is not part of the pull request. Is that a bug or a feature?

The commit SHA provided by the collaborator with be passed as `COMMIT_SHA_CHECK `, Jenkins checks that `GIT_REMOTE_REF` (i.e. `refs/pull/${prID}/head`) resolves to that expected commit SHA, otherwise the CI fails. So if someone pass the SHA of a commit not in the PR branch, it won't match the tip of the PR head, and it won't work

PR #926, opened 2025-05-04
normalize idl name when matching asset

panva:
the convention where a name matches exactly no longer holds since https://github.com/web-platform-tests/wpt/pull/52264


panva:
Actually, even the name lowercased name doesn't hold anymore... gotta think of something else.

PR #927, opened 2025-05-04
fix: prepare interfaces updater for updated webcrypto

panva:
adds webcrypto to the list of known interfaces to pull since https://github.com/web-platform-tests/wpt/pull/52264 makes it so the file no longer matches the directory name in webcrypto's case.


targos (approved):

panva:
Can we merge and release this? Future WPT updates to interfaces and webcrypto are blocked at the moment. https://github.com/nodejs/node/pull/58176

PR #928, opened 2025-05-09
fix(ncu-ci): pass `COMMIT_SHA_CHECK` to V8 CI

richardlau:
Similar to `node-test-pull-request`, `node-test-commit-v8-linux` also now requires `COMMIT_SHA_CHECK` to be passed to it.

Refs: https://github.com/nodejs/node-core-utils/pull/911

### DIFF
--- a/lib/ci/run_ci.js
+++ b/lib/ci/run_ci.js
@@ -62,7 +62,8 @@ export class RunPRJob {
       parameter: [
         { name: 'GITHUB_ORG', value: this.owner },
         { name: 'REPO_NAME', value: this.repo },
-        { name: 'GIT_REMOTE_REF', value: `refs/pull/${this.prid}/head` }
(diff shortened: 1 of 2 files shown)


richardlau:
This needs tests, but I've spent an hour and half failure to modify [`test/unit/ci_start.test.js`](https://github.com/nodejs/node-core-utils/blob/main/test/unit/ci_start.test.js) so I'll need help.

Specifically how to deal with https://github.com/nodejs/node-core-utils/blob/404c0ac9c225a89513cb794505751764209dc284/test/unit/ci_start.test.js#L24-L45 -- sinon has `onFirstCall`/`onSecondCall` for stubs but I couldn't get it to work.

targos (approved):
I have no experience with writing tests in this repo.

joyeecheung (approved):

PR #931, opened 2025-05-12
fix: auto-push git node --security

RafaelGSS:
**fix: auto-push git node --security**
```
This change makes git node --security to
auto push the changes to remote. This makes
this feature aligned with other git node security
features that also push to the remote automatically
```


RafaelGSS:
PTAL @marco-ippolito

PR #935, opened 2025-05-20
feat(git-node): add support for promoting several releases at once

aduh95:
And from a different remote. Currently, the `git node release --promote` workflow doesn't work well for security releases as it assumes the release proposal exists on the same repo as where we want to make the release, and releases are to be released one-by-one.
This PR aims at addressing all this shortcomings.

The process is now roughly:

1. for each proposal PR:
   1. Verify if it's ready (correct date, changelog entry, etc.).
   2. Create a signed tag for it.
   3. create a `Working on …` commit.
   4. rebase the public staging branch.
2. checkout the default branch, and cherry-pick all releases commit on top of it.
3. Print out a single `git push` command to push everything at once to the public repo.
4. Rest of the workflow is not affected.

I've tested this with 24.1.0 and 22.16.0, seems to work fine!

### DIFF
--- a/components/git/release.js
+++ b/components/git/release.js
@@ -6,7 +6,7 @@ import TeamInfo from '../../lib/team_info.js';
 import Request from '../../lib/request.js';
 import { runPromise } from '../../lib/run.js';
 
-export const command = 'release [prid|options]';
+export const command = 'release [prid..]';
 export const describe = 'Manage an in-progress release or start a new one.';
 
 const PREPARE = 'prepare';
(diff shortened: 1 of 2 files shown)


aduh95:
> or we should add an opt-in flag to promote several releases at once.

It is opt-in, you have to provide the URLs of the release proposal PR(s) you want to promote

RafaelGSS:
what we do normally (when more than one releaser) is to push all the proposals to private branch first, so we merge the proposal to vN.x and vN.x-staging, and once all releasers complete that, we push it to public. It reduces even more the "risk-time".

Issue #311, opened 2018-11-22
Why do we need package-lock.json?


joyeecheung:
package-lock.json is not published, and we are a CLI, so whoever installs this package via npm are not going to use it, the only use case seems to be for people who `git clone & npm link` but I don't think the file is necessary for that use case?

joyeecheung:
cc @targos

PR #939, opened 2025-06-15
chore: add back `package-lock.json` and enable dependabot

aduh95:
It was removed based of https://github.com/nodejs/node-core-utils/issues/311, however having a reproducible dev env is good actually.


targos:
In this project, I don't know. It has already happened in other projects I maintain.

targos (approved):
RSLGTM on the dependabot config

anonrig (approved):

PR #960, opened 2025-07-29
fix: support list detail of failed GitHub CI

himself65:
Get confused on the PR, I get no infomation of somehow cancelled CI

https://github.com/nodejs/node/pull/59221

### DIFF
--- a/test/unit/pr_checker.test.js
+++ b/test/unit/pr_checker.test.js
@@ -1941,6 +1941,297 @@ describe('PRChecker', () => {
       assert(status);
       cli.assertCalledWith(expectedLogs);
     });
+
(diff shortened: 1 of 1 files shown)


lib/queries/PRCommits.gql
```diff
@@ -31,7 +31,15 @@ query Commits($prid: Int!, $owner: String!, $repo: String!, $after: String) {
                   slug
                 }
                 conclusion,
-                status
+                status,
+                checkRuns(first: 40) {
```
himself65, line 35: github doesn't support 100,000 nodes query so I have to reduce to 40

aduh95 (approved):

PR #964, opened 2025-08-08
fix(ncu-config): support encrypted config in ncu-config

joyeecheung:
This otherwise always re-write the config into ~/.ncurc instead of ~/.ncurc.gpg

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -2,7 +2,7 @@ import path from 'node:path';
 import os from 'node:os';
 
 import { readJson, writeJson } from './file.js';
-import { existsSync } from 'node:fs';
+import { existsSync, mkdtempSync, rmSync } from 'node:fs';
 import { spawnSync } from 'node:child_process';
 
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -61,13 +61,31 @@ export function getConfigPath(configType, dir) {
 };
 
 export function writeConfig(configType, obj, dir) {
-  writeJson(getConfigPath(configType, dir), obj);
+  const configPath = getConfigPath(configType, dir);
+  const encryptedConfigPath = configPath + '.gpg';
+  if (existsSync(encryptedConfigPath)) {
+    const tmpDir = mkdtempSync(path.join(os.tmpdir(), 'ncurc-'));
```
aduh95, line 67: Could we use stdin instead of a temp dir?

richardlau (approved):

RafaelGSS (approved):

PR #969, opened 2025-08-30
fix(git-node): improve release tag signature verification

aduh95:
Add support for non-RSA keys, and for keys with several identities.

### DIFF
--- a/lib/promote_release.js
+++ b/lib/promote_release.js
@@ -196,7 +196,7 @@ export default class ReleasePromotion extends Session {
 
   async verifyTagSignature(version) {
     const { cli } = this;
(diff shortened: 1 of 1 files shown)


targos (approved):
Changes SGTM but I think it will still not match in my case.

The output of the `verify-tag` command looks like that:

```
gpg: Signature made Wed Aug 27 23:12:59 2025 CEST
gpg:                using RSA key 8FCCA13FEF1D0C2E91008E09770F7A9A5AE15600
gpg: Good signature from "Michaël Zasso (Targos) <targos@protonmail.com>" [ultimate]
```

Issue #974, opened 2025-09-17
Pin the dependencies in the published CLI via shrinkwrap


joyeecheung:
Since ncu is generally installed via `npm install -g` it seems useful to pin the transitive dependencies down via [npm-shrinkwrap.json](https://docs.npmjs.com/cli/v11/configuring-npm/npm-shrinkwrap-json)

cc @nodejs/tsc

PR #976, opened 2025-09-17
feat(deps): use shrinkwrap instead of package lock

joyeecheung:
As node-core-utils are CLI tools, it makes more sense to use shrinkwrap instead of package lock to allow users to install the CLI globally while pining transitive dependencies.

Fixes: https://github.com/nodejs/node-core-utils/issues/974


targos (approved):

joyeecheung:
Based on https://github.com/dependabot/dependabot-core/blob/ce5651df0d2302a9ed46639fd7a86be43b909d62/npm_and_yarn/lib/dependabot/npm_and_yarn/update_checker/dependency_files_builder.rb#L111 I am guessing it does.

targos:
Proven by https://github.com/nodejs/node-core-utils/pull/971

PR #982, opened 2025-09-23
fix(package): include shrinkwrap file in published package

targos:


targos:
@nodejs/node-core-utils

PR #989, opened 2025-10-12
feat(git-node): verify tag with official keyring

aduh95:
Instead of trying to parse `git tag` output (which is not stable), we could simply download the keyring from nodejs/release-keys and rely on the exit code. It should no longer provide false negatives.


targos (approved):
still LGTM

PR #1005, opened 2025-11-16
docs: nudge users into using partially encrypted config files

aduh95:
With https://github.com/nodejs/node-core-utils/issues/1004 landed, it probably makes little sense for users to encrypt their entire config file, and instead only encrypt tokens. This PR updates the doc to remove mention of full config file encryption, and present token encryption as the default / recommended way for storing them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,9 +119,22 @@ To obtain the Jenkins API token
    `~/.ncurc.gpg` or `$XDG_CONFIG_HOME/ncurc.gpg`) with `jenkins_token` as key,
    like this:
 
+   <details open name="set-jenkins-token"><summary>With encryption (recommended)</summary>
+   
    ```sh
    ncu-config set -x jenkins_token
    ```
+
+   Note: Encryption is available only if you have `gpg` setup on your machine.
+
+   </details>
+   <details name="set-jenkins-token"><summary>Without encryption</summary>
+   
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,18 +76,17 @@ Optionally, if you want to grant write access so `git-node` can write comments:
...
+```sh
+ncu-config set username your_github_username
+# It is strongly recommended to encrypt your token. Do not provide it in the CLI,
+# `ncu-config` will prompt you for it.
+ncu-config set -x token
 ```
 
-Note: you could use `ncu-config` to configure these variables, but it's not
-recommended to leave your tokens in your command line history.
+Note: Encryption is available only if you have `gpg` setup on your machine. If
```
joyeecheung: I think we should just split it into two different instructions with the encryption one prioritized as "Recommended". It's going to catch people off guard to recommend `-x` first only to come back with a note saying they should setup gpg first - that sounds like a pre-requisite that should be mentioned _before_ we suggest the `-x` flag (and also, what if you don't have gpg set up? The doc provides no suggestions, and it is not great to leave new contributors guessing what they should do in that case when they are new to all this and are just trying to set it up).

aduh95:
Ping @nodejs/node-core-utils for reviews

PR #1014, opened 2025-12-11
fix: improve suggestion next patch or next minor version

RafaelGSS:


lib/update_security_release.js
```diff
@@ -268,17 +269,26 @@ Summary: ${summary}\n`,
...
 
-      const nextPatchVersion = parseInt(version.split('.')[2]) + 1;
+      const nextPatchVersion = semver.inc(version, 'patch');
+      const nextMinorVersion = semver.inc(version, 'minor');
       const patchedVersion = await this.cli.prompt(
         `What is the patched version (>=) for release line ${affectedVersion}?`,
-        { questionType: 'input', defaultAnswer: nextPatchVersion });
+        {
+          questionType: 'input',
+          defaultAnswer: isPatchRelease ? nextPatchVersion : nextMinorVersion
```
aduh95: Nit: Let’s not compute both if we’re using only one
```suggestion
          defaultAnswer: semver.inc(version, isPatchRelease ? 'patch' : 'minor')
```

PR #1050, opened 2026-03-27
feat(git-node): add `git node security --apply-patches`

aduh95:
I've used that to prepare the 22.22.2 proposal; being able to fetch and apply the patches locally without running the release preparation script is much better UX IMO, and it's nice to be able to do that without needing to change the local config or manually download `vulnerabilities.json`.

### DIFF
--- a/components/git/security.js
+++ b/components/git/security.js
@@ -79,7 +79,7 @@ export function builder(yargs) {
       'Prepare a security release of Node.js'
     )
     .example(
-      'git node security --prepare-local-branch',
+      'git node security --apply-patches',
       'Fetch all the patches for an upcoming security release'
     )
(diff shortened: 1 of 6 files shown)


RafaelGSS (commented):
This is missing documentation on git-node

components/git/security.js
```diff
@@ -59,6 +76,10 @@ export function builder(yargs) {
       'git node security --start',
       'Prepare a security release of Node.js'
     )
+    .example(
+      'git node security --prepare-local-branch',
```
aduh95:
```suggestion
      'git node security --apply-patches',
```

RafaelGSS (approved):

PR #1062, opened 2026-04-15
fix(git-node): handle multi-line trailers

aduh95:
The current code assumes trailers will always be single-line, having a multi-line trailer currently results in non-sense output. This PR switch to relying on calling `git interpret-trailers` multiple times to filter out the trailers consistently.

Refs: https://github.com/nodejs/node/issues/62577#issuecomment-4245165710

### DIFF
--- a/lib/landing_session.js
+++ b/lib/landing_session.js
@@ -326,24 +326,25 @@ export default class LandingSession extends Session {
     const originalTrailers = interpretTrailers(original);
     const containCVETrailer = CVE_RE.test(originalTrailers);
 
-    const filterTrailer = (line) => ([key]) =>
-      line.startsWith(key) &&
-          new RegExp(`^${RegExp.escape?.(key) ?? key}\\s*:`).test(line);
     const amended = original.trim().split('\n');
-    const stillInTrailers = () => {
-      const result = interpretTrailers(amended.join('\n'));
-      return result.length && originalTrailers.startsWith(result.trim());
-    };
-    for (let i = amended.length - 1; amended[i] === '' || stillInTrailers(); i--) {
-      // Remove last line until git no longer detects any trailers
-      amended.pop();
-    }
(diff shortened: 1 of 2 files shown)


lib/landing_session.js
```diff
@@ -316,35 +316,43 @@ export default class LandingSession extends Session {
...
-    for (let i = amended.length - 1; amended[i] === ''; i--) {
+    const amended = original.trim().split('\n');
+    const stillInTrailers = () => {
+      const result = interpretTrailers(amended.join('\n'));
+      return result.length && originalTrailers.startsWith(result.trim());
+    };
+    for (let i = amended.length - 1; amended[i] === '' || stillInTrailers(); i--) {
+      // Remove last line until git no longer detects any trailers
       amended.pop();
     }
```
Renegade334: Presumably we could skip this step if `originalTrailers` is empty?

Renegade334: I wonder if there's a better way than spawning a git process per line until the trailers are completely identified. Doesn't `git interpret-trailers` require a double-newline before a trailer block – could we look backwards for this instead?

Renegade334 (commented):
Just as a courtesy, I changed the "Fixes" in the PR description as GH was threatening to cross-close Matteo's announcement issue on the core repo.

Renegade334 (approved):

lib/landing_session.js
aduh95: That's fair, I've added a shortcut so it skips all the lines until it finds an empty string, so most (all?) commit messages only need 2 spawn of `git interpret-trailers`

MikeMcC399:
Is this PR going ahead? I'm leaving my PR https://github.com/nodejs/node/pull/62738 open as a test with a multi-line 

```text
Signed-off-by: Mike McCready
  <66998419+MikeMcC399@users.noreply.github.com>
```

PR #1072, opened 2026-05-01
chore!: remove support for Node.js 20 and 25

targos:


Renegade334 (approved):

targos:
force-pushed to test with the released node 26

Renegade334:
> force-pushed to test with the released node 26

All published versions of c8 depend on yargs 17, whose extensionless cjs module is broken since nodejs/node#62176. Probably easiest if we just disable the coverage job on v26.x for now.

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Utilities for Node.js core collaborators",
   "type": "module",
   "engines": {
-    "node": "^20.19.0 || ^22.20.0 || >=24.10.0"
+    "node": "^22.20.0 || >=24.10.0"
```
aduh95: If we want to be able to bump `which`, we should align with https://github.com/npm/node-which/releases/tag/v7.0.0
```suggestion
    "node": "^22.22.2 || ^24.15.0 || >=26.0.0"
```

aduh95 (approved):

PR #1073, opened 2026-05-09
feat(v8): add llvm-libc to V8 deps

targos:
Selectively keep only the folders that are needed at the moment

Refs: https://chromium-review.googlesource.com/c/v8/v8/+/7828379

I tested a canary update locally and this fixes the build.


joyeecheung:
From the POV of keeping canary green, LGTM, but this implies we are going to statically compile some part of llvm-libc in? Is that going to work on e.g. macOS?

joyeecheung (approved):

targos:
I tested it on my macOS machine, but I don't really understand the implications.

richardlau (approved):

PR #1076, opened 2026-05-12
 chore: test upgrade-policy: lts/strict in matrix

joyeecheung:
The current release (26) doesn't work currently with the yargs version locked in by c8 until https://github.com/yargs/yargs/issues/2509 or https://github.com/bcoe/c8/pull/578 is fixed.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           node-version: lts/*
       - name: Install dependencies
-        run: npm install
+        run: npm ci
       - name: Lint
         run: npm run lint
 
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs.yml
```diff
@@ -15,9 +15,9 @@ jobs:
       - name: Use latest Node.js LTS
         uses: actions/setup-node@v6
         with:
-          node-version: lts/*
+          node-version: 24
       - name: Install dependencies
-        run: npm install
+        run: npm ci
```
joyeecheung, line 20: drive-by: this should use npm ci

joyeecheung:
hmm, oh wait actually the problem comes from pkgjs/action/.github/workflows/node-test.yaml

PR #1083, opened 2026-05-27
chore(deps): bump `which` from version 6 to 7

aduh95:
Refs: https://github.com/npm/node-which/releases/tag/v7.0.0


Renegade334 (approved):
The branch name did not go unnoticed...

(Technically "sub-breaking" under c564d818fb55f7cdd3159708a65584db64fd43b6 I guess, I don't know whether it's worth marking it as such for the releaser.)

PR #1089, opened 2026-06-01
chore: replace `process.exit` calls with actual error thrown

aduh95:
I can't think of a reason why we'd want to call `process.exit` instead of throwing an error, on the contrary throwing an error makes debugging and testing much more straightforward


targos:
I didn't write this code but from my experience, I tend to avoid throwing errors for cases that are not bugs in the code. As a CLI user, a stack trace is not the best UX for operational errors.

aduh95:
> As a CLI user, a stack trace is not the best UX for operational errors.

Those are disabled by default since #990.

targos (approved):

PR #1092, opened 2026-06-10
feat: add confirmation steps to any write operation

RafaelGSS:
One concern that @aduh95 shared with me is that `git node security *` performs a lot of things in the background without asking for permissions. This PR aims to add a security confirmation before any write operation.

### DIFF
--- a/lib/security-release/security-release.js
+++ b/lib/security-release/security-release.js
@@ -30,7 +30,36 @@ export const PLACEHOLDERS = {
   downloads: '%DOWNLOADS%'
 };
 
-export function checkRemote(cli, repository) {
+function formatCommand(command, args) {
+  return [command, ...args].join(' ');
+}
+
+export async function confirmSecurityStep(cli, action, detail) {
+  const lines = [`Allow action: ${action}?`];
+  if (detail) {
+    lines.push('', detail);
+  }
+  const message = lines.join('\n');
+
+  const allowed = await cli.prompt(message, { defaultAnswer: false });
+  if (!allowed) {
+    cli.info(`Aborted: ${action}.`);
+    process.exit(0);
+  }
+}
+
+export async function runSecurityGitCommand(cli, args, detail) {
+  const command = formatCommand('git', args);
+  await confirmSecurityStep(cli, `run \`${command}\``, detail);
+  return runSync('git', args);
+}
+
+export async function writeSecurityFile(cli, filePath, content, detail) {
+  await confirmSecurityStep(cli, `write \`${filePath}\``, detail);
+  return fs.writeFileSync(filePath, content);
+}
+
+export async function checkRemote(cli, repository) {
   const remote = runSync('git', ['ls-remote', '--get-url', 'origin']).trim();
   const { owner, repo } = repository;
   const securityReleaseOrigin = [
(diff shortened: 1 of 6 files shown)


lib/security-release/security-release.js
```diff
@@ -30,7 +30,36 @@ export const PLACEHOLDERS = {
...
+  const lines = [`Allow action: ${action}?`];
+  if (detail) {
+    lines.push('', detail);
+  }
+  const message = lines.join('\n');
+
+  const allowed = await cli.prompt(message, { defaultAnswer: false });
+  if (!allowed) {
+    cli.info(`Aborted: ${action}.`);
+    process.exit(0);
```
aduh95: Let's not call `process.exit`, that's an anti-pattern (refs https://github.com/nodejs/node-core-utils/pull/1089)

lib/security-release/security-release.js
```diff
@@ -30,7 +30,36 @@ export const PLACEHOLDERS = {
...
+function formatCommand(command, args) {
+  return [command, ...args].join(' ');
+}
+
+export async function confirmSecurityStep(cli, action, detail) {
+  const lines = [`Allow action: ${action}?`];
+  if (detail) {
+    lines.push('', detail);
+  }
+  const message = lines.join('\n');
```
aduh95: nit
```suggestion
  let message = `Allow action: ${action}?`;
  if (detail) {
    message += `\n\n${detail}`;
  }
```

lib/security-release/security-release.js
```diff
@@ -44,26 +73,42 @@ export function checkRemote(cli, repository) {
...
+        `This stages ${currentPath} for the security release commit.`
+      );
     }
   } else {
-    runSync('git', ['add', filePath]);
+    await runSecurityGitCommand(
+      cli,
+      ['add', filePath],
+      `This stages ${filePath} for the security release commit.`
+    );
```
aduh95: I don't think it makes sense to prompt for this, as this is a local-only operation, and non destructive.

lib/security-release/security-release.js
```diff
@@ -72,15 +117,31 @@ export function commitAndPushVulnerabilitiesJSON(filePath, commitMessage, { cli,
     return;
   }
 
-  runSync('git', ['commit', '-m', commitMessage]);
+  await runSecurityGitCommand(
+    cli,
+    ['commit', '-m', commitMessage],
+    `This creates a local commit with message: ${commitMessage}`
+  );
```
aduh95, line 114: ~Same here, prompting is unnecessary~ actually it's the right place to ask to review the diff. I would add we should probably get inspiration from `git node land` and offer to edit the commit message, but that's unrelated to this PR

lib/security-release/security-release.js
```diff
@@ -72,15 +107,31 @@ export function commitAndPushVulnerabilitiesJSON(filePath, commitMessage, { cli,
...
 
   try {
-    runSync('git', ['push', '-u', 'origin', NEXT_SECURITY_RELEASE_BRANCH]);
+    await runSecurityGitCommand(
+      cli,
+      ['push', '-u', 'origin', NEXT_SECURITY_RELEASE_BRANCH],
+      `This pushes the security release branch to origin/${NEXT_SECURITY_RELEASE_BRANCH}.`
+    );
```
aduh95, line 121: For the previous security release preparation, I commented this out and that was a much better UX, especially when GH is down. Maybe we could have a way to mark it as non-critical – or have an option to decide if you want the automatic pushes

aduh95 (approved):

PR #1097, opened 2026-06-18
fix(git-node): fix SSH passphrase input when promoting a release

aduh95:


Renegade334 (approved):
I'm guessing this isn't the only place where ssh could potentially go interactive?

aduh95:
> I'm guessing this isn't the only place where ssh could potentially go interactive?

It's the only place I've had problems with it recently

PR #1098, opened 2026-06-20
feat: output email when syncing team

avivkeller:
If possible, we should include the emails in these lists


avivkeller:
cc @nodejs/node-core-utils

Renegade334 (approved):

PR #1111, opened 2026-07-07
feat(security): add include-all report selection mode

RafaelGSS:
Reviewing each report was ok when we were handling 6~ reports per security release, now it's better to review it through the Webviewer

image

PTAL: @aduh95 @ulisesgascon @marco-ippolito


RafaelGSS:
PTAL @marco-ippolito

marco-ippolito (approved):

PR #1113, opened 2026-07-08
feat: add metadata JSON readiness output

panva:
This adds structured readiness output to `git node metadata --json` so automation can consume PR landability checks without matching human-readable CLI output.

The JSON output includes:

- `ready`
- `readiness`
- `exitCode`
- `metadata`
- `reasonCodes`
- `reasons`

It also adds stable reason codes from `PRChecker` for the existing readiness checks.

## Exit Codes

`git node metadata --json` uses category-level exit codes:

- `0`: PR is ready
- `20`: PR is not ready for deferrable metadata reasons
- `40`: PR is not ready for a hard or mixed reason
- `1`: unexpected tool/runtime failure

`20`-`29` are reserved for future deferrable metadata states, and `40`-`49` are reserved for future hard metadata failure states.

The current deferrable set is scoped to the lightweight commit-queue selector use case from nodejs/node#64343:

- ~`missing-approval`~
- ~`missing-tsc-approval`~
- `wait-time`


aduh95:
> The current deferrable set is scoped to […]:
>
> * `missing-approval`
> * `missing-tsc-approval`
> * `wait-time`

Can we change that to only capture wait time? E.g. missing TSC approval should be a hard error, no amount of time is going to help with that, and not erroring would be confusing IMO. Missing approval would mean that someone has added the CQ label on a 0-approval PR, again that seems like an error

panva:
@aduh95 makes sense, i'll update the utils and then the node PR

aduh95 (approved):
LGTM, we might also want to add PENDING_CI as deferable, so we can clean up https://github.com/nodejs/node/blob/db7cc5b40018e685f9f409b382884f84dec71c65/tools/actions/commit-queue.sh#L42-L46

panva:
I'm not sold on pending ci. With how flaky ci can get i think noone should be adding cq to a PR with an in progress ci. As such the cleanup you speak of would be imho better suited to go in the other direction.

aduh95:
That's fine by me. I think the idea is to get the CQ to notify you if the CI failed

PR #1114, opened 2026-07-09
feat(git-node): add `git node benchmark`

aduh95:
Add a tool to help spawning GHA benchmark from the CLI. Here's the output from my test run:

```console
$ git node benchmark https://github.com/aduh95/node/pull/43/commits/b511522444cd573cbc17ab50019edb29410f357a
✔  Fetched data for https://github.com/aduh95/node/pull/43
✔  Found 53 benchmark categories
✔  Found 0 benchmark file(s) in the PR changes
   ℹ  Based on the PR labels, this/these benchmark(s) seem relevant:
   ℹ  zlib
   ℹ  If not relevant, do not forget to unselect it/them.
✔ Select the benchmark categories to run: domain
✔ Substring to filter which benchmarks run (leave empty to run all): 
✔ Should the workflow post the results as a comment on the PR? Yes
---------------------------------- Benchmark -----------------------------------
PR:        https://github.com/aduh95/node/pull/43
Commit:    b511522444cd573cbc17ab50019edb29410f357a
Categories:domain
Comment on PR:yes
Workflow:  aduh95/node (benchmark.yml@main)
✔ Trigger the benchmark workflow with these settings? Yes
✔  Benchmark workflow dispatched
```

N.B.: the tests have been vibe-coded, part of the implementation has also been LLM-generated

### DIFF
--- a/lib/benchmark.js
+++ b/lib/benchmark.js
@@ -1,0 +1,239 @@
+import { basename } from 'node:path';
+
+import { getPrURL } from './links.js';
+import { IGNORE } from './run.js';
+
+const BENCHMARK_DIR = 'benchmark';
+// Entries in the benchmark folder that are not benchmark categories.
+const NON_CATEGORY_DIRS = new Set(['fixtures']);
+
+export default class BenchmarkSession {
+  constructor(cli, request, argv) {
+    this.cli = cli;
+    this.request = request;
+    this.argv = argv;
+    this.workflow = argv.workflow || 'benchmark.yml';
+  }
+
+  get prUrl() {
+    const { owner, repo, prid } = this.argv;
+    return getPrURL({ owner, repo, prid });
+  }
+
+  async getPullRequest() {
+    const { cli } = this;
+    cli.startSpinner(`Fetching data for ${this.prUrl}`);
+    const pr = await this.request.getPullRequest(this.prUrl);
+    if (!pr || !pr.head) {
+      cli.stopSpinner(
+        `Could not find pull request ${this.prUrl}`,
+        cli.SPINNER_STATUS.FAILED);
+      throw new Error(IGNORE);
+    }
+    cli.stopSpinner(`Fetched data for ${this.prUrl}`);
+    this.pr = pr;
+    return pr;
+  }
+
+  async getCategories() {
+    const { cli, pr } = this;
+    const head = pr.head;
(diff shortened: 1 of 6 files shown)


lib/benchmark.js
```diff
@@ -0,0 +1,241 @@
...
+
+    const workflowOwner = argv.workflowOwner || argv.owner;
+    const workflowRepo = argv.workflowRepo || argv.repo;
+    const prRepo = `${argv.owner}/${argv.repo}`;
+    // The workflow token can only comment when it runs in the PR repository;
+    // default to not commenting when the workflow lives elsewhere.
+    const sameRepo = `${workflowOwner}/${workflowRepo}` === prRepo;
+    const postComment = await cli.prompt(
+      'Should the workflow post the results as a comment on the PR?',
+      { defaultAnswer: sameRepo });
```
aduh95, line 188: This assumes that https://github.com/nodejs/node/pull/64395 will land, I would need to remove it if it doesn't

richardlau:
In the last few days we have added a GH workflow for [benchmark](https://github.com/nodejs/node/blob/main/.github/workflows/benchmark.yml) (which this PR is using) but also one for [stress runs](https://github.com/nodejs/node/blob/main/.github/workflows/stress-test.yml). 

Since these are not Jenkins maybe these don't belong to [ncu-ci](https://github.com/nodejs/node-core-utils/blob/main/docs/ncu-ci.md), but do we want to group all potential GH workflows into a group of commands? Or do we not care about the Jenkins/GH distinction and put these all under `ncu-ci`?

I'm not saying we need to add the other workflows now, just asking/thinking about whether we need a pattern of where this (and potential other) commands go.

aduh95:
I agree it would feel weird on `ncu-ci`. It's unclear to me whether you're suggesting a new binary, or a new `git-node` component with a more "neutral" name so it can be used for both (and potentially more in the future) workflows. It seems to me we would be better off having them both on `git-node` on different components (e.g. `git node benchmark` and `git node stress-test`), and they can both use the same `Request.prototype.dispatchWorkflow`, wdyt?

richardlau:
>  or a new `git-node` component with a more "neutral" name so it can be used for both (and potentially more in the future) workflows. 

This is what I was thinking of, but it's more of a suggestion and not a blocking issue for me.

mcollina (approved):
lgtm

PR #1116, opened 2026-07-13
refactor(v8): cache new V8 version during major update

Renegade334:
After updating the V8 tree, we currently re-fetch the new V8 version from the filesystem several times during various tasks. Instead, we can cache it in the same way that we do the existing version.

Updates `getCurrentV8Version()` to a generic task that accepts a label for the version, either 'current' or 'new', from which the context property name is derived.


targos (approved):
SGTM

PR #1117, opened 2026-07-14
feat(v8)!: synchronize Rust crates with major updates

Renegade334:
An implementation of https://github.com/nodejs/node/pull/64191#issuecomment-4890398627. Depends on #1116 and nodejs/node#64482.

Broadly speaking, the process is as follows:

1. Checkout the `third_party/rust` repository alongside the other DEPS. Note that this never enters the Node.js git tree, as it's deliberately not exempt from .gitignore, but fetching it here can make use of the existing DEPS machinery.
2. Use `cargo tree` to walk the temporal_capi dependency tree in `third_party/rust/chromium_crates_io`, and populate a list of transitive dependencies, so that we only vendor what we need for Temporal.
3. Build a new Cargo.toml manifest, copying any custom configs for crates that appear in the temporal_capi dependency tree.
4. Copy the vendored dependencies from `third_party/rust/chromium_crates_io` to `deps/crates`.
   - We don't use `cargo vendor` here, because Chromium uses its own gnrt toolchain to vendor its crate dependencies, which also applies any floating patches in `third_party/rust/chromium_crates_io/patches`. Using `cargo vendor` would fetch fresh copies from crates.io, and miss these patches.
   - The slight caveat is that, unlike `cargo vendor`, gnrt always adds some variation of the version string when naming the vendored crate directories. It doesn't matter for the rust build, since cargo doesn't care about the directory names, but it does mean that we need to update the temporal_capi include path in crates.gyp whenver we update the crates.
5. Regenerate the lockfile.
6. Update a standalone variable in crates.gyp containing the temporal_capi crate directory name, so that the V8 C++ build has the correct include path for the temporal_rs headers.
7. 🏁 

Example result: nodejs/node@f3bf26048a2bdaa38d9a56a84ca7d9a21317da57

Like Chromium, this approach doesn't add a `rust-version` to the crates manifest; Cargo will read this from the dependencies' manifests anyway. However, while it's not necessary for the build, I appreciate that it may be useful to see if our Rust requirement has changed, for updating BUILDING.md etc. If desired, we can either add this field or output the MSRV to terminal somehow; it'd just require interrogating each dependency's Cargo.toml while walking the tree.


targos (approved):

legendecas (commented):
Is this ready for review?

Renegade334:
If you have any comments on the overall approach then fire away, but I'll mark as ready for review once the associated core PR is merged, so that people can do proper test runs.

Renegade334:
> Like Chromium, this approach doesn't add a `rust-version` to the crates manifest; Cargo will read this from the dependencies' manifests anyway. However, while it's not necessary for the build, I appreciate that it may be useful to see if our Rust requirement has changed, for updating BUILDING.md etc. If desired, we can either add this field or output the MSRV to terminal somehow; it'd just require interrogating each dependency's Cargo.toml while walking the tree.

This wasn't actually much effort to implement, so I've added this in.

New example run: nodejs/node@2c67a5108e7a663d5662f4f6f3063ff60219efba

aduh95:
> If you have any comments on the overall approach then fire away, but I'll mark as ready for review once the associated core PR is merged, so that people can do proper test runs.

FWIW having the other PR merged is not necessary for doing test runs, one can apply the patch on nodejs/node already. I'd like this PR to land soon so we can release 7.0.0 which is long overdue at this point (but since this is marked semver major, it makes sense to wait for it)

Renegade334:
The upstream gyp PR has landed, so should be easy to test out on main.

I've drafted an update PR at nodejs/node#64543 so that the output can be checked with a CI run on different platforms.

legendecas (approved):

legendecas:
Thank you!

lib/update-v8/updateCrates.js
```diff
@@ -0,0 +1,189 @@
...
+const chromiumCratesDir = 'deps/v8/third_party/rust/chromium_crates_io';
+const nodeCratesDir = 'deps/crates';
+
+function enumerateTemporalDependencies() {
+  return {
+    title: 'Enumerate Temporal dependency tree',
+    task: async(ctx) => {
+      const tree = await forceRunAsync(
+        ctx.cargo,
+        ['tree', '-Z', 'bindeps', '--package', 'temporal_capi', '--prefix', 'none'],
```
targos, line 36:
```suggestion
        ['+nightly', 'tree', '-Z', 'bindeps', '--package', 'temporal_capi', '--prefix', 'none'],
```

Should we add this change? Without it, this command fails on my computer with:

```
➤ cargo tree -Z bindeps --package temporal_capi --prefix none
error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel
See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
```

Renegade334: Looks like Chromium sets [`RUSTC_BOOTSTRAP=1`](https://doc.rust-lang.org/beta/unstable-book/compiler-environment-variables/RUSTC_BOOTSTRAP.html) in its toolchain, does this work with your setup?

The issue with `+xyz` is that it only works in a rustup environment.

targos: I tested with this patch and it works too:

```patch
Index: lib/update-v8/updateCrates.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lib/update-v8/updateCrates.js b/lib/update-v8/updateCrates.js
--- a/lib/update-v8/updateCrates.js	(revision c263c2ae7d9be82feb79766a89a7bdaf5527b86b)
+++ b/lib/update-v8/updateCrates.js	(date 1785233230153)
@@ -37,7 +37,12 @@
         {
           ignoreFailure: false,
           captureStdout: true,
-          spawnArgs: { cwd: path.join(ctx.nodeDir, chromiumCratesDir) }
+          spawnArgs: {
+            cwd: path.join(ctx.nodeDir, chromiumCratesDir), env: {
+              ...process.env,
+              RUSTC_BOOTSTRAP: '1'
+            }
+          }
         }
       );
       const crates = new Set();
```

PR #1135, opened 2026-08-03
chore!: remove npm shrinkwrap, bundle package dependencies

### LINKED ISSUE #1078: npm shrink wrap is deprecated

Renegade334:
Alternative to #1121.

"Locks" the full dependency tree for consumers by bundling the dependencies (as constrained by our own package-lock.json) with the published package.

The size of the @node-core/utils package goes from

```
npm notice Tarball Details
npm notice package size: 206.1 kB
npm notice unpacked size: 839.8 kB
npm notice total files: 108
```

to

```
npm notice Tarball Details
npm notice package size: 8.0 MB
npm notice unpacked size: 39.7 MB
npm notice bundled deps: 355
npm notice bundled files: 0
npm notice own files: 7076
npm notice total files: 7076
```

although the dependencies that are now bundled would previously have to be resolved and installed by npm anyway.

The main differences from shrinkwrap:
- npm extracts the bundled node\_modules tree verbatim into node-core-utils/node\_modules, as opposed to performing npm dependency resolution with fixed versions from npm\_shrinkwrap. This means that it's no longer possible for npm to perform any deduplication at parent levels, so in the theoretical worst-case scenario where node-core-utils is installed globally alongside a load of other packages that have near-identical dependencies, this could add a maximum of ~40MB in no-longer-deduplicated node\_modules overhead (although in reality this should never be a significant issue).
- In what appears to be an npm bug, `npm pack` goes into an infinite resolution loop if a bundled dependency is overridden by an `overrides` entry, which we were relying upon until recently (984125a91eb8809b8210dacd9973662ac8ba0246). I'll report upstream, but for the time being, we wouldn't be able to call upon this feature if needed.

Fixes: https://github.com/nodejs/node-core-utils/issues/1078

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,31 @@
     "": {
       "name": "@node-core/utils",
       "version": "6.4.0",
+      "bundleDependencies": [
+        "@inquirer/prompts",
+        "@listr2/prompt-adapter-enquirer",
+        "@node-core/caritat",
+        "@pkgjs/nv",
+        "branch-diff",
+        "chalk",
+        "changelog-maker",
+        "cheerio",
+        "core-validate-commit",
+        "figures",
+        "ghauth",
+        "git-secure-tag",
+        "js-yaml",
+        "listr2",
+        "lodash",
+        "log-symbols",
+        "ora",
+        "replace-in-file",
+        "semver",
+        "smol-toml",
+        "undici",
+        "which",
+        "yargs"
+      ],
       "license": "MIT",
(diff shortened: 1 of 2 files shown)


aduh95 (approved):

Renegade334:
Rebased.

aduh95 (approved):
Based

PR #1140, opened 2026-08-07
chore(ncu-ci)!: remove `--copy` flag

aduh95:
Removes the dependency on `clipboardy`. Instead, pass the `--markdown` flag and pipe the result to e.g. `pbcopy`

Closes: https://github.com/nodejs/node-core-utils/pull/1131


Renegade334 (approved):
Definite +1 here, depending on unauditable binaries is not an enviable position for this package to be in.

PR #1153, opened 2026-08-09
perf: query CI metadata only for pull request head

panva:
To avoid exceeding GH API rate limits (e.g. [nodejs/node/actions/runs/31326707444/job/93278146885](https://github.com/nodejs/node/actions/runs/31326707444/job/93278146885)).

Landing checks only consume CI data from the final commit. Avoid querying check suites and runs for every commit, which can exhaust the GraphQL rate limit, and merge the head data after verifying its OID.

### DIFF
--- a/lib/pr_data.js
+++ b/lib/pr_data.js
@@ -60,9 +60,28 @@ export default class PRData {
       this.getThreadData(),
       this.getCommits()
     ]).then(() => {
+      this.mergeHeadCommitChecks();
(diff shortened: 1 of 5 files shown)


aduh95 (approved):
~This increases the number of queries~ nevermind, I read it wrong

panva:
> This increases the number of queries

The added `commits(last: 1)` is part of the existing PR GraphQL request, so it adds another connection selection but no additional API request.

PR #1158, opened 2026-08-10
More commit queue fixes

panva:
Filter check suites by the GitHub Actions App ID rather than requesting App metadata for every suite and checking a slug.

Preserve all GraphQL errors and render their structured data as JSON.

Assisted-by: codex:gpt-5.6-sol


lib/queries/PR.gql
```diff
@@ -29,11 +29,12 @@ query PR($prid: Int!, $owner: String!, $repo: String!) {
         nodes {
           commit {
             oid
-            checkSuites(first: 100) {
+            # 15368 is the GitHub Actions App ID on GitHub.com. Filtering avoids
+            # querying App metadata that GITHUB_TOKEN cannot read for suites
+            # created by third-party Apps such as Jenkins and Codecov.
+            # https://api.github.com/apps/github-actions
+            checkSuites(first: 100, filterBy: { appId: 15368 }) {
```
aduh95, line 36: This should solve the issue of phantom Jenkins jobs, right?

panva: I don't think so, but it's possible?. We still process the same suites as before downstream because we used to filter by app.slug amongst the results, now we just filter by the appId behind that slug server side.

aduh95 (approved):

Issue #1160, opened 2026-08-10
checkGitHubCI blocks indefinitely on orphaned check suites stuck in QUEUED


trivenay:
### Problem

`checkGitHubCI()` in `pr_checker.js` blocks landing with `✘ GitHub CI is still running` when any GitHub Actions check suite has `status !== 'COMPLETED'`. However, GitHub occasionally creates check suites that remain stuck in `QUEUED` status with 0 check runs indefinitely — they are never dispatched and will never complete.

This has blocked at least two recent PRs on `nodejs/node`:
- [#64991](https://github.com/nodejs/node/pull/64991) — `Test macOS` suite stuck 3+ days, landed manually
- [#64830](https://github.com/nodejs/node/pull/64830) — same `Test macOS` suite stuck, landed manually

In both cases, the orphaned suite:
- Has `status: QUEUED`, `conclusion: null`
- Has 0 check runs (never started executing)
- Has a valid `workflowRun` (`Test macOS`, workflow ID 858952)
- Was created at the same time as sibling suites that completed fine
- Never gets re-triggered by `request-ci` because GitHub sees it as already "queued"
- The only workaround is a collaborator manually running `git node land --yes`

The `commit-queue.sh` pre-check (`gh pr checks | grep pending`) passes fine because it evaluates individual check *runs* (all completed), not suite containers.

### Data

PR #64991, SHA `94a47d6f15b2`:
- 25 GitHub Actions suites total
- 24 completed (success/skipped)
- 1 orphaned: suite `84102428609`, `Test macOS`, `QUEUED`, 0 check runs, created `2026-08-05T13:19:48Z`

No newer run of `Test macOS` exists for this SHA — `request-ci` does not re-trigger a workflow GitHub considers already "queued."

### Proposed fix

Consider a suite as orphaned and skip it if:
- `status !== 'COMPLETED'`
- `checkRuns.nodes.length === 0` (nothing has ever executed inside it)
- Suite age exceeds a configurable threshold (suggesting 3 hours since `createdAt` — legitimate suites dispatch runs within seconds)

```js
if (status !== 'COMPLETED') {
  const runCount = checkRuns?.nodes?.length ?? 0;
  const ageMs = Date.now() - new Date(createdAt).getTime();
  if (runCount === 0 && ageMs > ORPHANED_SUITE_TIMEOUT) {
[...]

PR #1163, opened 2026-08-13
fix: skip orphaned check suites when checking GitHub CI

trivenay:
### Problem

`checkGitHubCI()` blocks landing with `✘ GitHub CI is still running` whenever any GitHub Actions check suite has a status other than `COMPLETED`. GitHub occasionally creates check suites that remain stuck in `QUEUED` with zero check runs — they are never dispatched and will never complete. These orphaned suites block the commit queue indefinitely, forcing collaborators to land PRs manually with `git node land`.

This was discussed in #1160 and observed on at least two recent nodejs/node PRs ([#64991](https://github.com/nodejs/node/pull/64991), [#64830](https://github.com/nodejs/node/pull/64830)), both stuck on an orphaned `Test macOS` suite and landed manually.

### Fix

Skip a non-completed suite only when **both**:
- it has zero check runs (`checkRuns.nodes.length === 0`), and
- it was created more than 3 hours ago.

A legitimately queued suite gets its check runs attached within seconds, so it will not match. Suites with active runs still block as before, and a freshly created empty suite (within the window) is not skipped. This is strictly narrower than the current manual workaround (`git node land --yes`), which bypasses *all* checks including real failures.

Adds `createdAt` to the `checkSuites` GraphQL query to compute suite age.

### Trade-off

If someone adds `commit-queue` before `request-ci` has finished triggering all workflows, a suite that should have run but hasn't started yet could — after aging past the 3h threshold — be skipped, landing the PR without that workflow's results. The 3h window makes this unlikely in practice, and the timeout can be made configurable in a follow-up if desired.

### Test

Added a unit test with an orphaned suite (0 runs, old `createdAt`) alongside a completed suite; the checker now passes instead of blocking. `test/unit/pr_checker.test.js`: 74/74 pass.

Refs: #1160

### Validated against the affected PRs

Ran the skip condition against the live check-suite data of both PRs that were stuck on this:

| PR | Orphaned suite | runs | age | result with this fix |
|----|----------------|------|-----|----------------------|
| [#64991](https://github.com/nodejs/node/pull/64991) | `queued` | 0 | ~199h | skipped → unblocked |
| [#64830](https://github.com/nodejs/node/pull/64830) | `queued` | 0 | ~358h | skipped → unblocked |

In both cases the single orphaned suite is exactly what blocked the commit queue, and this change lets the check pass instead of hanging indefinitely.

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -25,6 +25,10 @@ const WAIT_TIME_SINGLE_APPROVAL = 24 * 7;
 
 const GITHUB_SUCCESS_CONCLUSIONS = ['SUCCESS', 'NEUTRAL', 'SKIPPED'];
 const GITHUB_ACTIONS_APP = 'github-actions';
+// A GitHub Actions check suite with no runs that has not started after this
+// long is considered orphaned (created but never dispatched by GitHub) and is
+// ignored so it does not block landing. See nodejs/node-core-utils#1160.
+const ORPHANED_CHECK_SUITE_TIMEOUT = MINUTE * 60 * 3;
 
 const FAST_TRACK_RE = /^Fast-track has been requested by @(.+?)\. Please 👍 to approve\.$/;
 const FAST_TRACK_MIN_APPROVALS = 2;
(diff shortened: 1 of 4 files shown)


trivenay:
@aduh95 @panva — tagging you as you have worked on `pr_checker.js` and `PR.gql`. When you have a moment, I would appreciate a review of this small fix for orphaned check suites blocking the commit queue (context in https://github.com/nodejs/node-core-utils/issues/1160). No rush — happy to adjust the approach or the timeout if you would prefer something different.

PR #1170, opened 2026-08-27
fix: ignore fully skipped startup failures

panva:
GitHub Actions can report a check suite as `STARTUP_FAILURE` even when every check run in that suite was skipped. This caused `checkGitHubCI()` to mark the PR as unlandable without reporting a specific failed job or reason code.

Ignore this suite-level result when every check run was fetched and every run completed with `SKIPPED`. Other startup failures continue to block landing and now fall back to a suite-level diagnostic when no failed check run is available.

Query the check-run `totalCount` to ensure truncated results are not mistaken for a fully skipped suite.

Observed in https://github.com/nodejs/node/pull/65538#issuecomment-5445313746

https://github.com/nodejs/node/actions/runs/32985142178/attempts/1
https://github.com/nodejs/node/actions/runs/32984879569/attempts/1

Screenshot 2026-08-27 at 23 36 02

### DIFF
--- a/lib/pr_checker.js
+++ b/lib/pr_checker.js
@@ -467,43 +467,54 @@ export default class PRChecker {
       }
 
       if (!GITHUB_SUCCESS_CONCLUSIONS.includes(conclusion)) {
-        hasFailures = true;
+        const runs = checkRuns?.nodes ?? [];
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

PR #1171, opened 2026-08-28
fix: exclude PR-modified tests from the flaky count

### LINKED ISSUE #1164: ncu-ci walk: exclude PR-modified tests from the flaky count
### Current Behavior

Failures currently contribute to the flaky count even when the PR modified the failing test. Such failures are likely caused by the PR and should not count as independent flaky occurrences.

For example, CI build [75799](https://ci.nodejs.org/job/node-test-pull-request/75799/) failed `test-ffi-fast-buffer`, but the PR modified `test/ffi/test-ffi-fast-buffer.js`. Excluding this occurrence would leave only one qualifying PR, so the test would not appear in the reliability report:

https://github.com/nodejs/reliability/blob/main/reports/2026-08-15.md

### Expected Behavior

Exclude an occurrence from the flaky count when its test file was modified by the tested PR revision.

zeexzeex:
**What**

A failure in a test that the pull request itself modified is counted as an independent flaky occurrence, which overstates the number of affected pull requests.

Taking the example from the issue, `ffi/test-ffi-fast-buffer` is listed with two failed PRs in the [2026-08-15 report](https://github.com/nodejs/reliability/blob/main/reports/2026-08-15.md):

| PR | Files changed |
| --- | --- |
| [#65233](https://github.com/nodejs/node/pull/65233) | `src/ffi/fast.cc`, `test/ffi/test-ffi-fast-buffer.js` |
| [#65113](https://github.com/nodejs/node/pull/65113) | `lib/fs.js`, unrelated fs tests |

Only #65113 is a genuine occurrence.

**How**

Compare the failing test against the files the pull request touched, and drop the occurrence when they match. `formatAsMarkdown` already skips entries with fewer than two PRs, so such an entry leaves the report on its own.

When the file list cannot be fetched the occurrence is kept, following `lib/benchmark.js`: incomplete information should never remove a genuine failure.

This makes `aggregate()` async. Both call sites in `bin/ncu-ci.js` were already async methods. The lazy fallbacks inside `formatAsMarkdown()` and `display()` now throw instead, since a pending promise there would fail silently.

**Tests**

Four tests cover the exclusion, the untouched case, the fetch failure, and non-test failures. Removing the exclusion makes only the first fail, while the other three keep passing.

Fixes: https://github.com/nodejs/node-core-utils/issues/1164

### DIFF
--- a/lib/ci/failure_aggregator.js
+++ b/lib/ci/failure_aggregator.js
@@ -20,14 +20,47 @@ function uniqBy(array, key) {
 }
 
 export class FailureAggregator {
-  constructor(cli, data) {
+  constructor(cli, data, request) {
     this.cli = cli;
+    this.request = request;
     this.health = data[0];
     this.failures = data.slice(1);
     this.aggregates = null;
   }
 
-  aggregate() {
+  /**
+   * Tells whether the pull request that triggered the run also modified the
+   * test that failed. Such a failure is likely caused by the change itself,
+   * so it should not count as an independent flaky occurrence.
+   */
+  async isSelfInflicted(failure) {
+    const { file, source } = failure;
+    if (!file || !this.request) {
+      return false;
+    }
+
+    const pr = parsePRFromURL(source);
+    if (!pr) {
+      return false;
+    }
+
+    const path = `test/${file}.js`;
+    try {
+      for await (const changed of this.request.getPullRequestFiles(pr)) {
+        if (changed.filename === path) {
+          return true;
+        }
+      }
+    } catch {
+      // Not being able to fetch the changed files is not fatal: keep the
+      // occurrence rather than dropping it on incomplete information.
+      this.cli.warn(`Could not determine the files changed by ${source}`);
+    }
+
+    return false;
+  }
+
+  async aggregate() {
     const groupedByReason = Object.groupBy(this.failures, getHighlight);
     const data = [];
     for (const reason of Object.keys(groupedByReason).sort()) {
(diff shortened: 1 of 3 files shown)


lib/ci/failure_aggregator.js
```diff
@@ -37,7 +70,11 @@ export class FailureAggregator {
 
       // If multiple sub builds of one PR are failed by the same reason,
       // we'll only take one of those builds, as that might be a genuine failure
-      const prs = uniqBy(failures, 'source')
+      const candidates = uniqBy(failures, 'source');
+      const selfInflicted = await Promise.all(
+        candidates.map(failure => this.isSelfInflicted(failure)));
+      const prs = candidates
+        .filter((_, index) => !selfInflicted[index])
```
panva: This can leave `prs` empty, but `display()` still reads `prs[prs.length - 1].upstream`.

lib/ci/failure_aggregator.js
```diff
@@ -20,14 +20,47 @@ function uniqBy(array, key) {
...
+    if (!file || !this.request) {
+      return false;
+    }
+
+    const pr = parsePRFromURL(source);
+    if (!pr) {
+      return false;
+    }
+
+    const path = `test/${file}.js`;
```
panva: Node's test runner strips both `.js` and `.mjs` extensions from the reported name. For example, `test/parallel/test-cli-print-promise.mjs` is reported as `parallel/test-cli-print-promise`, so this always looks for the wrong file and keeps the occurrence in the flaky count.

lib/ci/failure_aggregator.js
```diff
@@ -20,14 +20,47 @@ function uniqBy(array, key) {
...
+    try {
+      for await (const changed of this.request.getPullRequestFiles(pr)) {
```
panva: This fetches the current PR's changed files, which may differ from the revision CI tested. If a contributor pushes a fix to the failing test after the run, the earlier failure gets excluded even though the tested revision never touched that file. That can reduce the count from two PRs to one and remove the reliability entry. Compare against the tested revision, or keep the occurrence when we cannot verify that the current PR head matches it.

PR #1172, opened 2026-08-28
chore(v8): replace Enquirer prompt adapter with Inquirer

trivikr:
Replace `@listr2/prompt-adapter-enquirer` with the [recommended](https://listr2.kilic.dev/task/prompts.html#adapters) Inquirer adapter.

The project already depends on `@inquirer/prompts`, so this avoids installing and maintaining a second prompt stack. The merge-conflict prompt in `update-v8` now uses the Inquirer `input` prompt while preserving its existing validation.

The [enquirer](https://listr2.kilic.dev/task/prompts.html#enquirer) is effectively unmaintained, and is kept on lifeline support only.

### DIFF
--- a/lib/update-v8/backport.js
+++ b/lib/update-v8/backport.js
@@ -3,8 +3,8 @@ import {
   promises as fs
 } from 'node:fs';
 
-import { confirm } from '@inquirer/prompts';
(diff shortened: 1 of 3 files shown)


trivikr:
cc @Renegade334 for review, as they'd made changes in `update-v8` folder in the recent past.

Renegade334 (approved):
Looks fine to me.

Image

PR #1173, opened 2026-08-31
feat: check large pull request approvals in git node land

### LINKED ISSUE #1063: Add large pull request checks in `git node land`
Refs: https://github.com/nodejs/node/pull/62829

There should be two points in `git node land` based on the above PR:
- Criteria to trigger this new check.
- Approval requirements (additional two TSC members approval)

zeexzeex:
Node.js requires two TSC approvals for large pull requests, but nothing enforces it at land time. This adds that check to `git node land`.

Whether a pull request is large is not something that can be derived from the diff. The policy counts a new subsystem as large regardless of size, and exempts routine dependency updates but not other dependency changes; neither is visible in the pull request data. An earlier revision of this PR tried a 5000 line threshold, and checking it against the 6 pull requests currently carrying the `large-pr` label, it missed 2 of them: https://github.com/nodejs/node/pull/64429 is well under the threshold, and https://github.com/nodejs/node/pull/62217 has a net negative diff.

The project already records this judgement with the `large-pr` label, so this checks that instead, the way `semver-major` is checked. The approval requirement is the same for both, so the existing branch is widened rather than duplicated, with a distinct message and reason code so a failure tells the author which rule applies. A pull request carrying both labels is reported as semver-major.

Refs: https://github.com/nodejs/node/blob/main/doc/contributing/large-pull-requests.md
Fixes: https://github.com/nodejs/node-core-utils/issues/1063


Renegade334:
I don't think that trying to automatically detect whether a PR falls into the policy definition of a large PR is ever going to be feasible. This can't detect whether a PR adds a new subsystem, or detect whether a non-automated dependency change PR meets the large PR threshold (it's only automated commits that are exempt), and indeed a PR only needs to touch a single gypfile in `/deps` for this logic to exempt it.

I think we would be much better off making this a label-based check, à la `fast track` or `semver-major`. Large PRs are not getting landed quickly, it's highly likely that a clued-in reviewer will check in to add the label long before it's in a position to land, and we can always couple it with an action to conditionally comment a reminder when the PR is opened.

zeexzeex:
You're right. I checked against the `large-pr` label: of the 6 PRs carrying it, my logic misses 2. #64429 is well under the threshold, and #62217 has a net negative diff plus the `dependencies` label, so it gets exempted twice over despite not being a routine update.

I'll rewrite this as a `large-pr` label check alongside the existing `semver-major` branch, and drop the line counting, the exemption list, and the `additions`/`deletions` query fields.

lib/pr_checker.js
```diff
@@ -206,16 +215,26 @@ export default class PRChecker {
       return false;
     }
 
-    if (isSemverMajor) {
+    if (isSemverMajor || isLargePR) {
```
Renegade334: Leave the existing semver-major processing alone and add the large PR handling separately. It'll make the code more easily readable and allows the behaviour of the two to diverge in the future if we need.

lib/pr_checker.js
```diff
@@ -221,6 +226,26 @@ export default class PRChecker {
...
+    if (isLargePR) {
+      const tscApproved = approved
+        .filter((p) => p.reviewer.isTSC())
+        .map((p) => p.reviewer.login);
+      if (tscApproved.length < 2) {
+        const message =
+          'large pull requests require at least 2 TSC approvals';
+        cli.error(message);
+        this.addReason(
+          PR_CHECK_REASON_CODES.MISSING_LARGE_PR_TSC_APPROVAL,
```
Renegade334, line 236: We can use the existing code here, rather than create a new one.

lib/pr_checker.js
```diff
@@ -43,6 +43,7 @@ export const PR_CHECK_REASON_CODES = Object.freeze({
   MISSING_FULL_JENKINS_CI: 'missing-full-jenkins-ci',
   MISSING_GITHUB_CI: 'missing-github-ci',
   MISSING_JENKINS_CI: 'missing-jenkins-ci',
+  MISSING_LARGE_PR_TSC_APPROVAL: 'missing-large-pr-tsc-approval',
```
Renegade334, line 46: I would get rid of this, the existing `MISSING_TSC_APPROVAL` code is appropriate

PR #1174, opened 2026-08-31
chore: use listr2 figure utilities

trivikr:
Replace the standalone [figures](https://www.npmjs.com/package/figures) dependency with the [figures utilities](https://github.com/listr2/listr2/blob/master/packages/listr2/src/utils/format/figures.ts) already exported by Listr2.

~Since Listr2 does not provide an `info` figure, its `pointer` figure is used for informational messages.~

|              | figures | listr2 |
|--------------|---------|--------|
| warning      | ⚠       | ⚠      |
| cross        | ✘       | ✖      |
| info/pointer | ℹ       | ℹ       |
| tick         | ✔       | ✔      |

### DIFF
--- a/lib/figures.js
+++ b/lib/figures.js
@@ -1,7 +1,7 @@
 import chalk from 'chalk';
-import figures from 'figures';
+import { figures } from 'listr2';
 
 export const warning = chalk.yellow(figures.warning);
 export const error = chalk.red(figures.cross);
-export const info = chalk.blue(figures.info);
+export const info = chalk.blue(figures.pointer);
 export const success = chalk.green(figures.tick);
--- a/lib/figures.js
+++ b/lib/figures.js
@@ -3,5 +3,5 @@ import { figures } from 'listr2';
 
 export const warning = chalk.yellow(figures.warning);
 export const error = chalk.red(figures.cross);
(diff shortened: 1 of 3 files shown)


lib/figures.js
```diff
@@ -1,7 +1,7 @@
 import chalk from 'chalk';
-import figures from 'figures';
+import { figures } from 'listr2';
 
 export const warning = chalk.yellow(figures.warning);
 export const error = chalk.red(figures.cross);
-export const info = chalk.blue(figures.info);
+export const info = chalk.blue(figures.pointer);
```
Renegade334: Honestly, I think that changing ℹ to ❯ loses readability-at-a-glance.

trivikr: I created a feature request in `listr2` to add `figures.info` https://github.com/listr2/listr2/issues/789
This PR can be revisited when it's added.

trivikr:
Moved to draft since we the `figures.info` is missing in `listr2`.
It can be revisited if the info symbol is added in https://github.com/listr2/listr2/issues/789

lib/figures.js
trivikr: PR now uses `figures.info` from `listr2@11.1.0`

Renegade334 (approved):

PR #1176, opened 2026-08-31
chore: replace cheerio with htmlparser2

trivikr:
Replace [cheerio](https://www.npmjs.com/package/cheerio) with the lighter, purpose-built [htmlparser2](https://www.npmjs.com/package/htmlparser2), avoiding the additional DOM and query abstractions we do not use.

The project only needs document text and `<a>` text/`href` values, so Cheerio’s selector and manipulation APIs are unnecessary. Using `htmlparser2` [callbacks](https://github.com/fb55/htmlparser2#usage) preserves the existing behavior without constructing a DOM, while reducing the total dependency count.

Comparison between the two packages: https://npmx.dev/compare?packages=cheerio,htmlparser2

---

Assisted-by: codex:gpt-5.6-sol


trivikr:
cc @aduh95 for review since he'd worked with `cheerio` in the past in this repo.

trivikr:
cc @panva Can you review this PR?

panva (approved):

PR #1183, opened 2026-09-03
refactor: replace chalk with util.styleText

trivikr:
Replace [chalk](https://www.npmjs.com/package/chalk) with Node.js’s built-in [util.styleText()](https://nodejs.org/api/util.html#utilstyletextformat-text-options).

### DIFF
--- a/lib/verbosity.js
+++ b/lib/verbosity.js
@@ -1,6 +1,4 @@
-import util from 'node:util';
-
-import chalk from 'chalk';
+import { format, styleText } from 'node:util';
 
 const VERBOSITY = {
   NONE: 0,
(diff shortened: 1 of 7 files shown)


lib/verbosity.js
```diff
@@ -25,5 +23,5 @@ export function setVerbosityFromEnv() {
 
 export function debuglog(...args) {
   // Prepend a line break in case it's logged while the spinner is running
-  console.error(chalk.green(util.format('\n[DEBUG]', ...args)));
+  console.error(styleText('green', format('\n[DEBUG]', ...args)));
```
Renegade334, line 26: Unrelated to this PR, but as a drive-by, wouldn't we want the line break _before_ the styleText sequence?

Renegade334 (approved):

lib/verbosity.js
trivikr: Ideally it should be. I'll pick it up in a follow-up PR.

panva (approved):

lib/verbosity.js
trivikr: Followed-up at https://github.com/nodejs/node-core-utils/pull/1185

PR #642, opened 2022-07-19
Add `ncu-cu resume <prid>` command

MoLow:
fixes: https://github.com/nodejs/node/issues/40817
refs: https://github.com/nodejs/node/issues/42125

PR #1187, opened 2026-09-09
feat: add ncu-ci resume command

panva:
Reimplements #642 by @MoLow on current main, adding `ncu-ci resume <prid>` with these additions:

- Accepts PR URLs as well as numeric IDs.
- Handles missing CI links without crashing and rejects invalid PR IDs.
- Rejects running jobs and results other than `FAILURE` or `ABORTED`. Checks Jenkins’ resume action, allowing resumable aborted jobs.
- Refuses to resume when the CI-approved commit differs from the current PR HEAD, or cannot be determined.
- Refuses when a failed test or failure diagnostic references a file changed by the PR. Handles renamed files, abbreviated test names, and literal or escaped Windows paths.
- Streams failed-job logs one at a time with bounded memory, including for large individual lines and TAP blocks. Requests compressed responses and decompresses them while streaming.
- Cancels the HTTP download when a match is confirmed and skips remaining logs.
- Uses available failure details without blocking on unknown or unavailable diagnostics.

Includes command documentation, CLI exit-code tests, and regression fixtures captured with `ncu-ci walk`.


jasnell (approved):

bin/ncu-ci.js
```diff
@@ -278,11 +299,13 @@ class RunPRJobCommand {
...
   }
 
-  async start() {
-    const {
-      cli, request, prid, repo, owner
-    } = this;
+  validate() {
+    const { cli, prid, repo, owner } = this;
     let validArgs = true;
+    if (!Number.isSafeInteger(prid) || prid <= 0) {
```
Renegade334, line 305: Bold to assume that we won't have received 2^53 agentic PRs by the end of the year...

lib/ci/failure_file_scanner.js
```diff
@@ -0,0 +1,127 @@
+import { StringDecoder } from 'node:string_decoder';
+
+// Keep failure context as filenames and flags, never as accumulated log text.
+// These markers cover the diagnostic families recognized by CIFailureParser.
+const diagnostic = /error:|error C\d+:|Error:|ERROR:|fatal:|FATAL:|java\.io\.IOException|Build timed out|hudson\.plugins\.git\.GitException|Cannot rebase:|sh: line |dtrace: failed to compile script|Makefile:.+failed|make.*:.*(?:Error \d|write error)|warning: failed/;
+const infrastructure = /Read-only file system|Device or resource busy|There is not enough space in the file system\./;
```
Renegade334: Is it worth modifying ci_failure_parser such that these can be extracted from there rather than duplicated?

docs/ncu-ci.md
```diff
@@ -172,6 +173,40 @@ ncu-ci run https://github.com/nodejs/node/pull/34127/commits/35ea6ded7315cf9d058
...
+`ncu-ci resume <prid>` resumes the latest `node-test-pull-request` CI run linked
+in the PR description, comments, or reviews. The job must have finished with
+`FAILURE` or `ABORTED` and expose Jenkins' resume action. Running jobs and jobs with
+other results are not resumed. If no PR CI run is found, the command reports that
+and exits unsuccessfully.
+
+The CI-approved commit (`COMMIT_SHA_CHECK`) must match the PR's current HEAD.
+The command refuses to resume if they differ or the approved commit cannot be
+determined.
+
```
aduh95, line 187: FWIW Jenkins already checks for that (for the HEAD, not for approvals). Just saying in case that adds too much complexity

docs/ncu-ci.md
```diff
@@ -172,6 +173,40 @@ ncu-ci run https://github.com/nodejs/node/pull/34127/commits/35ea6ded7315cf9d058
...
+Before resuming, the command streams failed-job console output and compares
+failure diagnostics with the PR's changed files, including previous filenames
+for renames. It refuses to resume if a failed test or a file referenced in a
```
aduh95: > including previous filenames for renames

What's the logic of that? If it's been renamed it won't show up in CI

docs/ncu-ci.md
panva: Yes, but that check runs inside the build script, so we'd get a resume link followed by a failed build, rather than `[resume-ci-failed]` on the PR, which is what I was aiming for.

lib/ci/resume_ci.js
```diff
@@ -0,0 +1,150 @@
...
+    this.cli = cli;
+    this.request = request;
+    this.owner = owner;
+    this.repo = repo;
+    this.prid = prid;
+  }
+
+  async checkFailures(jobid) {
+    const { cli, request, owner, repo, prid } = this;
+    cli.startSpinner('Checking failures against changed PR files');
```
aduh95: I would move this out of this function

docs/ncu-ci.md
panva: The old path can still appear in failure diagnostics, e.g. when renaming a fixture leaves an import pointing to its previous filename. That would catch a failure caused by the rename. I'll adjust the regression test to demonstrate the broken-reference case instead of what is there atm. That being said GHA would likely also be red for that case.

lib/ci/resume_ci.js
```diff
@@ -114,6 +113,7 @@ export class ResumePRJob {
         return false;
       }
 
+      cli.startSpinner('Checking failures against changed PR files');
       if (!(await this.checkFailures(job.jobid))) {
         return false;
       }
```
aduh95, line 118: The `stopSpinner` calls should be moved here as well

mcollina (approved):
lgtm


nice stream processing with fetch

PR #1191, opened 2026-09-11
fix: exclude stale matrix runs from CI reports

panva:
Jenkins can include older configuration runs in a later matrix build. Store the upstream cause where the filter reads it and match the full build number so those runs do not appear in subsequent CI reports.

For example, `ncu-ci pr 77202` reports `client-proxy/test-https-proxy-request-invalid-char-in-url` as failing for [nodejs/node#65852](https://github.com/nodejs/node/pull/65852). That [osx13-x64 #73203 failure](https://ci.nodejs.org/job/node-test-commit-osx/nodes=osx13-x64/73203/console) actually belongs to the preceding [CI #77201](https://ci.nodejs.org/job/node-test-pull-request/77201/), testing nodejs/node#64722 on Node.js 22.

Jenkins includes that older configuration in the `runs` array for [macOS matrix #73204](https://ci.nodejs.org/job/node-test-commit-osx/73204/). This fix filters it out using its upstream build number, so #77202 no longer reports the unrelated proxy failure while retaining its actual failures.

Refs: https://github.com/nodejs/reliability/blob/main/reports/2026-09-11.md

Assisted-by: Codex

### DIFF
--- a/lib/ci/build-types/normal_build.js
+++ b/lib/ci/build-types/normal_build.js
@@ -76,7 +76,7 @@ export class NormalBuild extends Job {
       if (!upstream) {
         return true;
       } else {
-        return this.jobUrl.includes(upstream + '');
+        return this.jobUrl.endsWith(`/${upstream}/`);
       }
     });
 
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

PR #1193, opened 2026-09-15
fix: handle unavailable Jenkins commit build data

panva:
Jenkins can retain a successful PR build result while returning `null` for its nested commit build. This caused `git node metadata` to throw `TypeError: Cannot read properties of null (reading 'subBuilds')`, aborting [commit queue filtering](https://github.com/nodejs/node/actions/runs/34948903953/job/104314954048).

Preserve the recorded result when child build details are unavailable. For non-successful builds, report the unavailable details without throwing.


mcollina (approved):
lgtm

